PR #107, opened 2016-09-12
Fix flickering of translation strings before save is complete

willyau:
This will show the translation in the state while
save is occuring.  When save is done, the prop
is updated, and the component will be rerendered.


aurambaj:
👍

PR #118, opened 2016-09-30
changes to support xliff exported from xcode

jeeyi:


webapp/src/main/java/com/box/l10n/mojito/okapi/AssetExtractionStep.java
```diff
@@ -72,13 +73,13 @@ protected Event handleTextUnit(Event event) {
         ITextUnit textUnit = event.getTextUnit();
 
         if (textUnit.isTranslatable()) {
-            String name = textUnit.getName();
+            String name = StringUtils.isEmpty(textUnit.getName()) ? textUnit.getId() : textUnit.getName();
```
aurambaj: we're trying to Guava instead of Apache commons. StringUtils#isEmpty() can be replaced by Strings#isNullOrEmpty()

webapp/src/main/java/com/box/l10n/mojito/okapi/TranslateStep.java
```diff
@@ -110,7 +111,7 @@ protected Event handleTextUnit(Event event) {
 
             String translation = null;
 
-            String name = textUnit.getName();
+            String name = StringUtils.isEmpty(textUnit.getName()) ? textUnit.getId() : textUnit.getName();
```
aurambaj: same

jeeyi:
@aurambaj Mojito can handle whitespaces in the xliff if there's xml:space="preserve".  But xcode does not set this attribute when it exports xliff for localization.  Also when xcode imports the xliff with xml:space="preserve", it does not escape the newline character properly in generating .strings.

PR #122, opened 2016-10-04
enable FileFinderTest#findXliffNoBaseName

jeeyi:
@aurambaj The issue was having test method name findXliffNoBasename not exactly matching with actual directory name findXliffNoBaseName (Basename vs. BaseName)


aurambaj:
👍

PR #128, opened 2016-10-06
Xcode xliff

jeeyi:


jeeyi:
@aurambaj Do you have any better idea how to preserve space in xcode generated xliff?

cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
```diff
@@ -203,10 +204,18 @@ LocalizedAssetBody getLocalizedAsset(Repository repository, FileMatch sourceFile
             logger.debug("Getting the asset for path: {} and locale: {}", sourceFileMatch.getSourcePath(), repositoryLocale.getLocale().getBcp47Tag());
             Asset assetByPathAndRepositoryId = assetClient.getAssetByPathAndRepositoryId(sourceFileMatch.getSourcePath(), repository.getId());
 
+            String assetContent = commandHelper.getFileContent(sourceFileMatch.getPath());
+
+            // TODO(P1) This is to inject xml:space="preserve" in the trans-unit element
+            // in the xcode-generated xliff until xcode fixes the bug of not adding this attribute
+            if (fileType != null && fileType.getClass() == XliffNoBasenameFileType.class) {
+                assetContent = commandHelper.setPreserveSpaceInXliff(assetContent);
```
aurambaj: Not sure we want to do that in the CLI. We have all the processing logic in server right now. I'd move it there. Also it is really IOS specific so should we have a type for that Xliff?

jeeyi: I will keep the logic in client side for now because we are not passing filetype to server side.  I can rename XliffNoBasenameFileType to XcodeXliffFileType to be more clear.

PR #132, opened 2016-10-21
Fix issue with node-fetch and compression for application/json.

aurambaj:
Before compression was not activated for json. First, it is bad for performance. Then if proxying requests (say a default Nginx setup that as compression on for json) it would break the app.

This is related to a strange behavior with node-fetch, which gives the following error: "invalid response body at:  http://localhost:8080/ap…mit=10 reason: data error: incorrect header check", type: "system", errno: "Z_DATA_ERROR", code: "Z_DATA_ERROR"})."

Considered removing 'node-fetch' but the native fetch API is only supported by Safari 10 and we'd like to keep support for now. Chrome, Firefox are fine.

 - enable compression for application/json
 - change node-fetch configuration to workaround the issue. Kind of counter intuitive but using "compress: false" prevents the issue and compression is still performed.
 - added "crendentials: 'include'" which is a setting needed if using native fetch (can be useful later if/when we remove node-fetch)

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -14,7 +14,7 @@
     "keycode": "^2.1.0",
     "keymirror": "~0.1.0",
     "lodash": "^3.10.1",
-    "node-fetch": "^1.3.2",
+    "node-fetch": "^1.6.3",
     "node-sass": "^3.3.3",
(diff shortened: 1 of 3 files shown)


jeeyi:
👍

PR #138, opened 2016-12-13
Toggle repository locale stats detail

jeeyi:
togglelocalestats mov

### DIFF
--- a/webapp/src/main/resources/public/js/components/drops/DropDetail.js
+++ b/webapp/src/main/resources/public/js/components/drops/DropDetail.js
@@ -22,15 +22,6 @@ let DropDetail = React.createClass({
     },
 
     /**
-     * Handle onClick close button
-     */
-    onClickClose() {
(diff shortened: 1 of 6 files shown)


jeeyi:
@aurambaj Similar change is applied to project request page.  Please review.

webapp/src/main/resources/public/js/components/drops/DropDetail.js
```diff
@@ -62,9 +53,6 @@ let DropDetail = React.createClass({
 
         return (
                 <div>
-                    <div className="title">
-                        <Button className="close" onClick={this.onClickClose}>×</Button>
```
aurambaj: is css "close" class still in use after those 2 changes?

jeeyi: Good catch.  We can remove the it from css.

PR #139, opened 2016-12-15
Support MacString keys not enclosed in double quotes

jeeyi:
@aurambaj  This change is to support keys that are not enclosed in double quotes.  InfoPlist.strings can have it.  For example,
```
/* Keys without double quotes */
Example 1 = "first example";

/* Keys with double quotes */
"Example 2" = "second example";
```


jeeyi:
@aurambaj I made additional changes to support line comments:

```
// line comments
Example = "This is example";
```

PR #148, opened 2017-03-08
Support for PO files

aurambaj:


aurambaj:
@jeeyi Good catch. No the files are utf8 but looks like the header I copied/generated didn't have a proper encoding defined so okapi replaced it with utf16, regardless of the actual encoding. 

Fixed the test files.

jeeyi:
@aurambaj Good job 👍

PR #151, opened 2017-03-13
Refactor AuthenticatedRestTemplate to distinguish methods that build …

### LINKED ISSUE #149: rest client double encodes uri parameters
Hi

I think I came across a problem whereby path uri params are being double encoded in the rest client.  My call via the cli was similar to the following (via cygwin):

```
java -jar mojito-cli-0.67.jar pull -r MyRepository -s T:/workspace/myproject/src/main/resources/i18n
```

Which  resulted in an error message:

```
Asset with path [abc\def\Example.properties] was not found in repo [MyRepository]
```

I tracked down the problem and came up with a potential fix here: https://github.com/garyhodgson/mojito/commit/f477396a81eaf421c016f5ee6bd31236f57f29d3

In short the getForObject method uses getURIForResource which encodes the parameters as a URI String, but then the restTemplate encodes it again later on.  E.g. "/api/assets?path=abc%5Cdef" is actually sent as "/api/assets?path=abc%255Cdef" which mojito can then not find.  I attempted a crude unit test which shows the problem: https://github.com/garyhodgson/mojito/commit/f477396a81eaf421c016f5ee6bd31236f57f29d3#diff-c8652fa74e390a8c272cab005f7ba9f9R79

My initial attempt at a fix was to use the alternative getForObject method which accepts uriVariables - but this didn't work for some reason.  Perhaps someone with a better understanding of Spring Rest could get it to work as it's a probably the better solution.  My fix is simply to pass an instance of a URI as the parameter, which the rest template recognises and does not try and encode again.

Hope this helps.  I can create a Pull request if you wish.

aurambaj:
…URI with query params from resttemplate method that uses uriVariables (substitued in the URI template) and fix encoding issue.

Fixes #149

### DIFF
--- a/restclient/src/main/java/com/box/l10n/mojito/rest/client/AssetClient.java
+++ b/restclient/src/main/java/com/box/l10n/mojito/rest/client/AssetClient.java
@@ -154,7 +154,7 @@ public class AssetClient extends BaseClient {
             filterParams.put("deleted", deleted.toString());
         }
         
-        return authenticatedRestTemplate.getForObjectAsList(getBasePathForEntity(), Asset[].class, filterParams);
+        return authenticatedRestTemplate.getForObjectAsListWithQueryStringParams(getBasePathForEntity(), Asset[].class, filterParams);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


jeeyi:
👍

PR #158, opened 2017-03-30
added support for xtb file format

jeeyi:


cli/src/main/java/com/box/l10n/mojito/cli/command/param/Param.java
```diff
@@ -50,7 +50,7 @@
 
     public static final String FILE_TYPE_LONG = "--file-type";
     public static final String FILE_TYPE_SHORT = "-ft";
-    public static final String FILE_TYPE_DESCRIPTION = "File type (if none provided it will scan for default formats: XLIFF, XCODE_XLIFF, MAC_STRING, ANDROID_STRINGS, PROPERTIES, PROPERTIES_NOBASENAME, RESW, RESX)";
+    public static final String FILE_TYPE_DESCRIPTION = "File type (if none provided it will scan for default formats: XLIFF, XCODE_XLIFF, MAC_STRING, ANDROID_STRINGS, PROPERTIES, PROPERTIES_NOBASENAME, RESW, RESX, PO, XTB)";
```
aurambaj: 😛

jeeyi: We need to also update Mojito documentation for the new file formats.

aurambaj: yes, wanted to do it but didn't have time yet. Also the import command needs to be documented.

jeeyi: Yes, I can work on it altogether sometime next week.  Also I am going to add a test case to import localized xtb files in ImportLocalizedAssetCommandTest in the separate pull request.

PR #162, opened 2017-04-03
Implement PseudoLocalization in Mojito

srizviBox:


webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	/**
+	 * This is the equivalent to PHP substr method.
+	 * @param string
+	 * @param from
+	 * @param to
+	 * @return
+	 */
+	private String substr(String string, int from, int to) {
```
srizviBox: I need to keep this function because what the native .substring does is not what i want done.

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
```diff
@@ -0,0 +1,203 @@
...
+			int matchesLength = mappingCharsForChar.length();
+			int min = 0 + (int) (Math.random() * (matchesLength - 1));
+			mappedChar = substr(mappingCharsForChar, min, 1);
+		}
+
+		return mappedChar;
+	}
+
+	/**
+	 * This is the equivalent to PHP substr method.
```
willyau: are you reimplementing http://php.net/manual/en/function.substr.php here?

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
willyau: have you looked into google guava or other java libraries that will do the same for you?

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
srizviBox: yeah

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
willyau: Please also add test to this method if you're keeping it.

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
willyau: what does the java substring do?

srizviBox: handle negative indicies :)

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
aurambaj: can you explain it what way it differs? What you're missing? we really don't want to add those kind of function.

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
aurambaj: hm, let's reimplement the code that depends on that the Java way then. It doesn't really make sense to introduce this kind of function here.

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
```diff
@@ -0,0 +1,203 @@
...
+	 * Converts ASCII letter into equivalent characters with accent/diacritics
+	 *
+	 * @param string String to be converted
+	 * @return
+	 */
+	public String convertAsciiToDiacritics(String string) {
+		int stringLength = string.length();
+		StringBuilder str = new StringBuilder();
+		for (int i = 0; i < stringLength; i++) {
+			String character = substr(string, i, 1);
```
jeeyi: char character = string.charAt(i);

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * @author srizvi
+ */
+public class PseudoLocalization {
+
+	private static final int EXPANSION_FACTOR = 30;
+	private static final String BEGINNING_CHAR = "萬";
+	private static final String ENDING_CHAR = "國";
+
+	private static Map<String, String> pseudoLocMap = new HashMap<String, String>();
```
jeeyi: Map<Character, String>

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
```diff
@@ -0,0 +1,203 @@
...
+	private String getMappingCharFromMap(String character) {
+		String mappedChar = character;
+
+		String mappingCharsForChar = pseudoLocMap.get(mappedChar);
+
+		if (mappingCharsForChar != null) {
+
+			int matchesLength = mappingCharsForChar.length();
+			int min = 0 + (int) (Math.random() * (matchesLength - 1));
+			mappedChar = substr(mappingCharsForChar, min, 1);
```
jeeyi: It can also be replaced with char mappedChar = mappingCharsForChar.charAt(min)

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
jeeyi: Or even Map<Character, List<Character>> pseudoLocMap;
Then you would do pseudoLocMap.put('a', Lists.newArrayList('à', 'á', 'â', 'ã', 'ä', 'å', 'ā', 'ă', 'ą', 'ǻ', 'ά', 'α', 'а')); 
And to get the character for 'a' at specific index, pseudoLocMap.get('a').get(index);
Then you don't need substr().

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
```diff
@@ -0,0 +1,165 @@
...
+	 */
+	private char getMappingCharFromMap(char character) {
+		char mappedChar = character;
+
+		String mappingCharsForChar = pseudoLocMap.get(mappedChar);
+
+		if (mappingCharsForChar != null) {
+
+			int matchesLength = mappingCharsForChar.length();
+			int min = 0 + (int) (Math.random() * (matchesLength - 1));
```
jeeyi: For readability,
int maxIndex = mappingCharsForChar.length() - 1;
int randomIndex = (int)(Math.random() * maxIndex);

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
```diff
@@ -0,0 +1,165 @@
...
+	 * Expands the given string by the expansion factor with some unicode characters.
+	 *
+	 * @param string
+	 * @return
+	 */
+	public String expand(String string) {
+
+		int expandedStringLength = (int) Math.ceil(string.length() * EXPANSION_FACTOR / 100.0);
+
+		StringBuilder str = new StringBuilder();
```
aurambaj: nitpicking but can you replace `str` by `sb`

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
```diff
@@ -0,0 +1,165 @@
...
+	/**
+	 * Pseudo localize a string depending on instance settings to choose the type, expansion factor and bracket option.
+	 * If pseudo-localization is disabled, returns the passed in string
+	 *
+	 * @param string string to be pseudo localized
+	 * @return pseudo localized string
+	 */
+	public String convertStringToPseudoLoc(String string)
+	{
+		String expanded = "";
```
aurambaj: can be removed, just use the StringBuilder

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
```diff
@@ -0,0 +1,165 @@
...
+		StringBuilder sb = new StringBuilder();
+
+		if (Strings.isNullOrEmpty(string)) {
+			expanded = convertAsciiToDiacritics(string);
+			expanded = expand(expanded);
+			sb.insert(0,'⟦');
+			sb.append(expanded);
+			sb.append('⟧');
+
+			expanded = sb.toString();
```
aurambaj: remove

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
```diff
@@ -0,0 +1,165 @@
...
+		}
+
+		return expanded;
```
aurambaj: `return sb.toString()` here

webapp/src/main/java/com/box/l10n/mojito/pseudoloc/PseudoLocalization.java
```diff
@@ -0,0 +1,165 @@
...
+
+	/**
+	 * Converts ASCII letter into equivalent characters with accent/diacritics
+	 *
+	 * @param string String to be converted
+	 * @return
+	 */
+	public String convertAsciiToDiacritics(String string) {
+		int stringLength = string.length();
+		StringBuilder str = new StringBuilder();
```
aurambaj: same `sb` instead of `str`

webapp/src/test/java/com/box/l10n/mojito/pseudoloc/PseudoLocalizationTest.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ *
+ */
+public class PseudoLocalizationTest {
+
+	@Test
+	public void testExpandExpandsString() {
+		PseudoLocalization ps = new PseudoLocalization();
+		String expanded = ps.expand("English Sentence");
+		System.out.println(expanded);
```
aurambaj: use logger instead of `System.out.println`

PR #164, opened 2017-04-05
Implement backend changes need enable pseduoloc for mojito

srizviBox:


srizviBox:
Sorry to keep switching the pull requests, this should be the last time i do that. I believe i addressed all the comments now :)

PR #165, opened 2017-04-06
Call PseudoLocalization Class from PseudoLocalizeStep

srizviBox:


webapp/src/main/java/com/box/l10n/mojito/okapi/PseudoLocalizeStep.java
```diff
@@ -13,10 +13,11 @@
  *
  * @author srizvi
  */
-public abstract class PseudoLocalizeStep extends BasePipelineStep {
+public class PseudoLocalizeStep extends BasePipelineStep {
 
     protected String source;
```
aurambaj: remove `source` and `textunit` attributes. Just use local variable in `handleTextUnit`

webapp/src/main/java/com/box/l10n/mojito/okapi/PseudoLocalizeStep.java
```diff
@@ -13,10 +13,11 @@
  *
  * @author srizvi
  */
-public abstract class PseudoLocalizeStep extends BasePipelineStep {
```
aurambaj: you need @Configurable on the class

webapp/src/main/java/com/box/l10n/mojito/okapi/PseudoLocalizeStep.java
```diff
@@ -13,10 +13,11 @@
...
+public class PseudoLocalizeStep extends BasePipelineStep {
 
     protected String source;
     protected ITextUnit textUnit;
+    protected PseudoLocalization pseudoloc;
```
aurambaj: please rename to  `pseudoLocalization` and you need `@Autowired` to inject the dependency into the class

PR #167, opened 2017-04-09
Upgrade okapi

aurambaj:


jeeyi:
I tested importing previously generated xliff with upgraded okapi and it works as expected.  👍

willyau:
👏👏👏 yes!

PR #168, opened 2017-04-12
Adding function in AssetWS.java to getPseudoLocalizedAssetForContent

srizviBox:
Im adding tests soon

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/rest/asset/AssetWS.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/rest/asset/AssetWS.java
@@ -172,7 +172,7 @@ public class AssetWS {
         String bcp47tag = "en-x-psaccent";
         String normalizedContent = NormalizationUtils.normalize(localizedAssetBody.getContent());
 
-        String generateLocalized = tmService.generateLocalized(
+        String generateLocalized = tmService.generatePseudoLocalized(
                 asset,
                 normalizedContent,
                 null,
--- a/webapp/src/main/java/com/box/l10n/mojito/rest/asset/AssetWS.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/rest/asset/AssetWS.java
@@ -169,14 +169,11 @@ public class AssetWS {
         logger.debug("Pseudo localizing content payload with asset id = {}", assetId);
 
         Asset asset = assetRepository.getOne(assetId);
-        String bcp47tag = "en-x-psaccent";
         String normalizedContent = NormalizationUtils.normalize(localizedAssetBody.getContent());
 
         String generateLocalized = tmService.generatePseudoLocalized(
                 asset,
                 normalizedContent,
-                null,
-                bcp47tag,
                 localizedAssetBody.getFilterConfigIdOverride());
 
         localizedAssetBody.setContent(generateLocalized);
--- a/webapp/src/main/java/com/box/l10n/mojito/rest/asset/AssetWS.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/rest/asset/AssetWS.java
@@ -153,7 +153,37 @@ public class AssetWS {
         return localizedAssetBody;
     }
     
-    
+    /**
+     * Pseudo localizes the payload content with translations of a given {@link Asset}.
+     *
+     * @param assetId {@link Asset#id}
+     * @param localizedAssetBody the payload to be localized with optional
+     * parameters
+     * @return the pseudo localized payload as a {@link LocalizedAssetBody}
+     */
+    @RequestMapping(value = "/api/assets/{assetId}/pseudo", method = RequestMethod.POST)
+    public LocalizedAssetBody getPseudoLocalizedAssetForContent(
+            @PathVariable("assetId") long assetId,
+            @RequestBody LocalizedAssetBody localizedAssetBody) {
+
+        logger.debug("Pseudo localizing content payload with asset id = {}", assetId);
+
+        Asset asset = assetRepository.getOne(assetId);
+        String bcp47tag = "en-x-psaccent";
+        String normalizedContent = NormalizationUtils.normalize(localizedAssetBody.getContent());
+
+        String generateLocalized = tmService.generateLocalized(
+                asset,
+                normalizedContent,
+                null,
+                bcp47tag,
+                localizedAssetBody.getFilterConfigIdOverride());
+
+        localizedAssetBody.setContent(generateLocalized);
+
+        return localizedAssetBody;
+    }
+
     //TODO(P1) It would be nice to put this to be POST on .../localized/{localeId}
     // but it won't backward compatible so... The URL is taken by another POST 
     // that is used as a GET because we needed to send the paylaod. Here would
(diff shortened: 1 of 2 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/tm/TMService.java
```diff
@@ -709,7 +710,11 @@ public String generateLocalized(
 
         driver.addStep(new RawDocumentToFilterEventsStep());
         driver.addStep(new CheckForDoNotTranslateStep());
-        driver.addStep(new TranslateStep(asset, repositoryLocale, InheritanceMode.USE_PARENT));
+        if (bcp47Tag.equals("en-x-psaccent")) {
```
aurambaj: invert the test `"en-x-psaccent".equals(bcp47Tag)`, it is better code to avoid NPE
introduce a constant for `"en-x-psaccent"`

webapp/src/main/java/com/box/l10n/mojito/service/tm/TMService.java
```diff
@@ -709,7 +710,11 @@ public String generateLocalized(
...
+            driver.addStep(new PseudoLocalizeStep());
+        } else {
+            driver.addStep(new TranslateStep(asset, repositoryLocale, InheritanceMode.USE_PARENT));
```
aurambaj: so if we're doing that here... there is no point of having a different WS, it would work by calling the normal "pull" command and `getLocalizedAssetForContent` WS. 

I think @jeeyi  was more leaning toward an new webservice + command... then we should probably have 2 function in the `TMService`

srizviBox: @aurambaj I think the idea was to have this code and then create a new "pull command." We would then create a function in AssetClient (getPseudoLocalizedAssetForContent) to call this endpoint, /api/assets{assetId}/pseudo.

jeeyi: I want to have separate CLI command for pseudoloc.  But I don't want to create a separate function in TMService which is very similar to generateLocalized function.  Instead of doing checking `en-x-psaccent".equals(bcp47Tag)`, we can pass in a flag to do pseudoloc.

aurambaj: If we use 2 different commands I would refactor slightly `TMService`. I don't think it is super clean to start adding too much logic from 2 different features in that function, specially since Pseudoloc doesn't need a big chuck of it.

I'd introduce `generateLocalizedBase(BasePipelineStep step, String content, String bcp47tag, FilterConfigIdOverride filterConfigIdOverride)` that contains the end of `generatedLocalized` starting at the pipeline configuration. `generateLocalized` will pass down `new TranslateStep(asset, repositoryLocale, InheritanceMode.USE_PARENT)`  as step and have the logic for output tag. For pseudo, have another function `generatePseudolocalized` that pass down the `PseudoLocalizeStep` into `generateLocalizedBase`

webapp/src/main/java/com/box/l10n/mojito/service/tm/TMService.java
```diff
@@ -706,15 +705,61 @@ public String generateLocalized(
...
+            Asset asset,
+            String content,
+            RepositoryLocale repositoryLocale,
+            String outputBcp47tag,
+            FilterConfigIdOverride filterConfigIdOverride) {
+
+        String bcp47Tag;
+
+        if (outputBcp47tag == null) {
+            bcp47Tag = repositoryLocale.getLocale().getBcp47Tag();
```
aurambaj: this whole block to identify the bcp47 tag should be in `generateLocalized` and not here

webapp/src/main/java/com/box/l10n/mojito/rest/asset/AssetWS.java
```diff
@@ -153,7 +153,37 @@ public LocalizedAssetBody getLocalizedAssetForContent(
...
+     */
+    @RequestMapping(value = "/api/assets/{assetId}/pseudo", method = RequestMethod.POST)
+    public LocalizedAssetBody getPseudoLocalizedAssetForContent(
+            @PathVariable("assetId") long assetId,
+            @RequestBody LocalizedAssetBody localizedAssetBody) {
+
+        logger.debug("Pseudo localizing content payload with asset id = {}", assetId);
+
+        Asset asset = assetRepository.getOne(assetId);
+        String bcp47tag = "en-x-psaccent";
```
aurambaj: bcp47tag is not needed as a parameter

webapp/src/main/java/com/box/l10n/mojito/rest/asset/AssetWS.java
```diff
@@ -153,7 +153,37 @@ public LocalizedAssetBody getLocalizedAssetForContent(
...
+        String normalizedContent = NormalizationUtils.normalize(localizedAssetBody.getContent());
+
+        String generateLocalized = tmService.generatePseudoLocalized(
+                asset,
+                normalizedContent,
+                null,
```
aurambaj: not needed remove from the function signature

webapp/src/main/java/com/box/l10n/mojito/service/tm/TMService.java
```diff
@@ -705,11 +705,52 @@ public String generateLocalized(
...
+     * for a bcp47 tag that is different from the repository locale (which is
+     * still used to fetch the translations). This can be used to generate a
+     * file with tag "fr" even if the translations are stored with fr-FR
+     * repository locale.
+     * @return the localized asset
+     */
+    public String generatePseudoLocalized(
+            Asset asset,
+            String content,
+            RepositoryLocale repositoryLocale,
```
aurambaj: in relation to the other comments, `repositoryLocale` parameter is not needed

webapp/src/main/java/com/box/l10n/mojito/service/tm/TMService.java
```diff
@@ -705,11 +705,52 @@ public String generateLocalized(
...
+            String outputBcp47tag,
```
aurambaj: nor `outputBcp47tag`

jeeyi:
💯

webapp/src/main/java/com/box/l10n/mojito/service/tm/TMService.java
```diff
@@ -705,11 +705,52 @@ public String generateLocalized(
...
+
+    /**
+     * Parses the given content and adds the pseudo localization for every text unit.
+     * Returns the pseudolocalized content.
+     *
+     * @param asset The {@link Asset} used to get translations
+     * @param content The content to be localized
+     * @param repositoryLocale the repository locale used to fetch the
+     * translation. Also used for the output tag if outputBcp47tag is null.
+     * @param outputBcp47tag Optional, can be null. Allows to generate the file
```
aurambaj: remove old java doc for parameters that were removed

webapp/src/main/java/com/box/l10n/mojito/service/tm/TMService.java
```diff
@@ -705,11 +705,52 @@ public String generateLocalized(
...
+        BasePipelineStep pseudoLocalizedStep = (BasePipelineStep) new PseudoLocalizeStep();
+        return generateLocalizedBase(asset, content, filterConfigIdOverride, bcp47tag, pseudoLocalizedStep);
+    }
+
+    /**
+     *
+     * @param asset
+     * @param content
+     * @param filterConfigIdOverride
+     * @param bcp47Tag
```
aurambaj: Keep and update the doc from the old function

PR #169, opened 2017-04-13
For Android, extract note from XML comments in addition of getting th…

aurambaj:
…em from the description attribute. If a description attribute already contains a note the extraction from the XML comments is skipped


webapp/src/main/java/com/box/l10n/mojito/okapi/filters/XMLFilter.java
```diff
@@ -191,11 +197,61 @@ public Event next() {
...
 
         return event;
     }
 
+    /**
+     * Extract the note from XML comments only if there is no note on the 
+     * text unit. In other words if a note was specify via attribute like 
+     * description for android it won't be overridden by an comments present in 
+     * the XML file.
+     *
```
jeeyi: This is cool - won't break existing android strings with description 👍

PR #175, opened 2017-05-18
Fix workbench error that appears with newer dependencies by removing node-fetch

### LINKED ISSUE #172: Uncaught TypeError when trying to access the workbench
Hi
got a really strange error while trying to access the translations in the workbench.

workbench

if I'm downloading the jar's from your latest release directly from your github repository everything is working as expected, whereas when I build the fork locally I get the above error. Obviously setting my head to the 0.73 version tag before building.
Any help is highly  appreciated.

aurambaj:
According to a comment it was added to support Safari 9 (and had some hack for compression). Current Safari version is 10.1, we probably don't need support for 9 anymore. 

Fix #172


jeeyi:
Thanks Jean! 👍

PR #176, opened 2017-05-23
adding more languages in Mojito - de, es, it, ja, pt, zh-Hans and zh-…

jeeyi:
…Hant

### DIFF
--- a/webapp/src/main/resources/properties/zh-Hans.properties
+++ b/webapp/src/main/resources/properties/zh-Hans.properties
@@ -1,0 +1,423 @@
+# Label displayed in the header menu to open the workbench
+header.workbench=Workbench
+
+# Label displayed in the header menu to open the repositories page
+header.repositories=Repositories
+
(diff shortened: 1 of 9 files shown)


webapp/src/main/resources/properties/zh-Hans.properties
```diff
@@ -0,0 +1,423 @@
...
+header.workbench=Workbench
+
+# Label displayed in the header menu to open the repositories page
+header.repositories=Repositories
+
+# Label displayed in the header menu to open the project requests page
+header.projectRequests=Project Requests
+
+# Label displayed in the header menu to open the settings page
+header.settings=Settings
```
aurambaj: @jeeyi we never got the translation for this? just wondering :)

jeeyi: Hanna was working with some students to translate them. I need to look at the old email threads on it.

PR #177, opened 2017-05-24
removed region from getNativeDisplayName

jeeyi:
@hkanabiajeuskaja has requested to remove region from language selector.

screen shot 2017-05-23 at 10 31 11 pm


jeeyi:
Yeah, we need a way to distinguish between Simplified Chinese and Traditional Chinese :)

jeeyi:
Updated screenshot:
screen shot 2017-05-24 at 11 18 36 am

PR #190, opened 2017-08-01
Update Translations

boxmoji:


jeeyi:
@srizviBox Pseudo-localized file is now being generated!

PR #196, opened 2017-08-13
Move to webpack for CSS

aurambaj:


webapp/src/main/resources/templates/index.html
```diff
@@ -4,10 +4,6 @@
         <title>Mojito</title>
         <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
 
-        <link rel="stylesheet" type="text/css" href="{{contextPath}}/css/mojito.css" />
-        <link rel="stylesheet" type="text/css" href="{{contextPath}}/css/bootstrap-multiselect.css" />
-        <link rel="stylesheet" type="text/css" href="{{contextPath}}/css/animate.min.css">
```
aurambaj: do you remember if we use animate somewhere? I couldn't find it.

PR #210, opened 2017-10-27
Pmaster

aurambaj:


jeeyi:
@aurambaj While testing I noticed that stats (string and word counts) are incorrect with Android repos. Have you seen problems with stats?

aurambaj:
I changed the stats computation to take in account the plural forms depending on the language... it should be working. Did you notice issue with some languages or overall?

jeeyi:
I had existing Android repo with plurals that are fully translated. It only had `one` and `other` in the source resource file. When I generate localized files with the latest server, it adds more plural forms in the generated localized files depending on the plural rules of each language. However, those newly generated ones (`one`, `two`, etc) do not show up on the workbench. The existing translations show up without plural tags in the workbench. When the plural string gets updated, then they appear in workbench as  expected with plural tags. However, the existing approved translations become `Needs-Translation`.  Is this all expected?

jeeyi:
Attached screen shot of stats not working:

screen shot 2017-11-17 at 9 36 12 am

jeeyi:
Let's merge!

PR #213, opened 2017-11-16
Chromeext

aurambaj:


jeeyi:
@aurambaj What does chrome extension do?

aurambaj:
The chrome extension is the client side of the in context system. It can be used on Mojito itself for demo purposes.

- Turn on/off in context
screen shot 2017-11-16 at 9 31 31 am
- Allows to set some headers on all request going out to the apps. The headers can be recognized by the app that in turn instruments string (inject special metadata)
- When an app returns instrumented string the chrome extension will process them so that the user can mouse hover (it gets the green border, mouse pointer is gone on the screenshot) on the string and with shift + click you can open the in context modal.
screen shot 2017-11-16 at 9 33 46 am

- the in context modal allows to see information about the string and open it in mojito  
screen shot 2017-11-16 at 9 34 39 am

It is the first iteration, there is some polish to be done and also finish the build/packaging for the extension but it is usable

PR #214, opened 2017-11-16
option to create review drop with inherited translations

jeeyi:
We need an option to create REVIEW drop of child locale with inherited translations.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/translationkit/TranslationKitService.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/translationkit/TranslationKitService.java
@@ -3,6 +3,7 @@ package com.box.l10n.mojito.service.translationkit;
 import com.box.l10n.mojito.common.StreamUtil;
 import com.box.l10n.mojito.entity.Drop;
 import com.box.l10n.mojito.entity.Locale;
+import com.box.l10n.mojito.entity.RepositoryLocale;
 import com.box.l10n.mojito.entity.TM;
 import com.box.l10n.mojito.entity.TMTextUnit;
 import com.box.l10n.mojito.entity.TMTextUnitVariant;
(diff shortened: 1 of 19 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/translationkit/TranslationKitService.java
```diff
@@ -322,4 +304,95 @@ public void updateStatistics(Long translationKitId, Set<String> notFoundTextUnit
...
+    }
+
+    private List<TextUnitDTO> getTextUnitDTOsForTranslationKit(Long repositoryId, Long localeId, StatusFilter statusFilter) {
+
+        int retry = 1;
+        int maxRetry = 3;
+        List<TextUnitDTO> textUnitDTOs = null;
+
+        while (retry <= maxRetry) {
+            try {
```
aurambaj: wondering why we had this retry logic, seems useless. try to remove it?

webapp/src/main/java/com/box/l10n/mojito/service/translationkit/TranslationKitService.java
```diff
@@ -322,4 +304,95 @@ public void updateStatistics(Long translationKitId, Set<String> notFoundTextUnit
...
+        Stack<Long> localeIds = getLocaleInheritanceStack(repositoryId, localeId);
+
+        Map<Long, TextUnitDTO> mergedTextUnitDTOs = new TreeMap<>();
+        List<TextUnitDTO> textUnitDTOs = null;
+        while (!localeIds.empty()) {
+            Long currentLocaleId = localeIds.pop();
+            if (currentLocaleId == localeId) {
+                // child locale
+                textUnitDTOs = getTextUnitDTOsForTranslationKit(repositoryId, currentLocaleId, StatusFilter.REVIEW_NEEDED);
+                logger.info("child locale {} has {} text units for review", currentLocaleId, textUnitDTOs.size());
```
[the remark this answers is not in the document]
jeeyi: You are right. I forgot to change it.

webapp/src/main/java/com/box/l10n/mojito/service/translationkit/TranslationKitService.java
jeeyi: Yeah, let's give a try.

PR #217, opened 2018-01-23
jquery upgraded to 3.0.0

jeeyi:


webapp/package.json
```diff
@@ -11,7 +11,7 @@
     "bootstrap-sass": "^3.3.7",
     "cldr-data": "^28.0.2",
     "cldrjs": "^0.4.3",
-    "jquery": "^2.1.4",
+    "jquery": "~3.0.0",
```
aurambaj: ^3.0.0?

jeeyi: I will change that.

PR #220, opened 2018-01-24
Update the package-lock.json file too

aurambaj:


aurambaj:
@jeeyi maybe that file didn't get generated because you have an older version of npm. probably to update it.

PR #234, opened 2018-03-12
Inheritance mode support for pull command

dragosv:
This is needed so that only translated strings are brought back, otherwise, the integration between Mojito and serge.io (Open-Source Solution for Continuous Localization by Evernote: https://serge.io) will not work properly.

See the plugin at  https://github.com/dragosv/serge/tree/mojito, will need to add this extra parameter once the pull request is accepted. The cli adds --inheritance-mode as a parameter for the pull command and the rest api will default to USE_PARENT if none specified. Also, the serge.io plugin should use REMOVE_UNTRANSLATED.

The plugin will be pushed into its own repository that I will support, as serge.io will not support any other translation services outside Pootle.

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/LocalizedAssetBodyInheritanceMode.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/LocalizedAssetBodyInheritanceMode.java
@@ -6,24 +6,24 @@ import com.box.l10n.mojito.rest.entity.LocalizedAssetBody;
 
 /**
  *
- * @author jaurambault
+ * @author dragosv
  */
 public class LocalizedAssetBodyInheritanceMode implements IStringConverter<LocalizedAssetBody.InheritanceMode> {
 
     @Override
     public LocalizedAssetBody.InheritanceMode convert(String string) {
 
-        LocalizedAssetBody.InheritanceMode statusEqualTarget = null;
+        LocalizedAssetBody.InheritanceMode inheritanceMode = null;
 
         if (string != null) {
             try {
-                statusEqualTarget = LocalizedAssetBody.InheritanceMode.valueOf(string.toUpperCase());
+                inheritanceMode = LocalizedAssetBody.InheritanceMode.valueOf(string.toUpperCase());
             } catch (IllegalArgumentException iae) {
-                throw new ParameterException("Invalid \"inheritance type\" processing type [" + string + "]");
+                throw new ParameterException("Invalid \"inheritance mode\" processing type [" + string + "]");
             }
         }
 
-        return statusEqualTarget;
+        return inheritanceMode;
     }
 
 }
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/LocalizedAssetBodyInheritanceMode.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/LocalizedAssetBodyInheritanceMode.java
@@ -1,0 +1,29 @@
+package com.box.l10n.mojito.cli.command;
+
+import com.beust.jcommander.IStringConverter;
+import com.beust.jcommander.ParameterException;
+import com.box.l10n.mojito.rest.entity.LocalizedAssetBody;
+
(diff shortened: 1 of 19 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
```diff
@@ -62,6 +60,11 @@
     @Parameter(names = {Param.SOURCE_REGEX_LONG, Param.SOURCE_REGEX_SHORT}, arity = 1, required = false, description = Param.SOURCE_REGEX_DESCRIPTION)
     String sourcePathFilterRegex;
 
+    @Parameter(names = {"--inheritance-mode"}, required = false, description = "How to handle missing translations "
+            + "the parent (USE_PARENT for using parent). ",
+            converter = LocalizedAssetBodyInheritanceMode.class)
+    LocalizedAssetBody.InheritanceMode inheritanceMode = null;
```
aurambaj: Initialize with default value: USE_PARENT. That way the documentation shows the default value

cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
```diff
@@ -62,6 +60,11 @@
     @Parameter(names = {Param.SOURCE_REGEX_LONG, Param.SOURCE_REGEX_SHORT}, arity = 1, required = false, description = Param.SOURCE_REGEX_DESCRIPTION)
     String sourcePathFilterRegex;
 
+    @Parameter(names = {"--inheritance-mode"}, required = false, description = "How to handle missing translations "
```
aurambaj: description -> `Inheritance Mode. Used when there is no translations in the target locale for a text unit. (USE_PARENT to fallback to parent locale translation, REMOVE_TRANSLATED to remove the text unit from the file)`

cli/src/main/java/com/box/l10n/mojito/cli/command/LocalizedAssetBodyInheritanceMode.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public LocalizedAssetBody.InheritanceMode convert(String string) {
+
+        LocalizedAssetBody.InheritanceMode statusEqualTarget = null;
+
+        if (string != null) {
+            try {
+                statusEqualTarget = LocalizedAssetBody.InheritanceMode.valueOf(string.toUpperCase());
+            } catch (IllegalArgumentException iae) {
+                throw new ParameterException("Invalid \"inheritance type\" processing type [" + string + "]");
```
aurambaj: `inheritance type` -> `inheritance mode`

webapp/src/test/java/com/box/l10n/mojito/service/tm/TMServiceTest.java
```diff
@@ -1346,7 +1347,7 @@ public void testLocalizeXtb() throws Exception {
             logger.debug("source=[{}]", textUnitDTO.getSource());
         }
 
-        String localizedAsset = tmService.generateLocalized(asset, assetContent, repoLocale, "en-GB", null);
+        String localizedAsset = tmService.generateLocalized(asset, assetContent, repoLocale, "en-GB", null, InheritanceMode.USE_PARENT);
         logger.debug("localized=\n{}", localizedAsset);
         assertEquals(expectedContent, localizedAsset);
     }
```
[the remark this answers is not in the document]
dragosv: Added test for a few formats. Works except for the xtb one as it fails to output translationbundle (ignored that test for now)

aurambaj:
Thanks for all the tests! Let's keep `Xtb` the way you did with `@Ignore` shouldn't be a big deal.

If you can just fix the import. I'd rather stick the standard order and no change is needed in that case (sorry for being picky) and I'll merge!

aurambaj:
@dragosv Thanks!!

jeeyi:
@dragosv Awesome!! 👍 💯 🏆

PR #237, opened 2018-03-15
Fix release

jeeyi:
@aurambaj I removed jar signing and deployment that are specific to Box. This will make release independent. What do you think?


aurambaj:
Yes I think it is better to remove the Box specific deployment if it is not too many change on your side.

PR #239, opened 2018-03-16
Revert node upgrade

aurambaj:


aurambaj:
@jeeyi I'm also having issue with that update on our jenkins so rollbacking for now

PR #264, opened 2018-05-15
Fix workbench pagination

aurambaj:


webapp/src/main/resources/public/js/actions/workbench/SearchDataSource.js
```diff
@@ -49,19 +49,29 @@ const SearchDataSource = {
...
+
+                    let hasMore = false;
+
+                    if (textUnits.length === limit) {
+                        hasMore = true;
+                        textUnits = textUnits.slice(0, limit -1);
+                    }
 
-                    return results;
+                    return {textUnits: textUnits, hasMore: hasMore};
```
willyau: syntax shortcut:

return { textUnits, hasMore }

aurambaj: @willyau is back!! thanks!!

PR #287, opened 2018-07-09
Use Quartz for job scheduling to support cluster deployment

aurambaj:


aurambaj:
@jeeyi Moving job scheduling to use Quartz instead of spring basic support. This to be able to deploy a mojito cluster. It also makes some of or scheduling smoother. This is a relatively big change so heads up!

jeeyi:
@aurambaj If this is a big change, I'd like to create a release tag before you merge with the latest UI fixes from Edwin and async TM export.

aurambaj:
@jeeyi sure!

jeeyi:
@aurambaj FYI, I used V27. Now you have to rename V27__Quartz.sql to V28__Quartz.sql.

aurambaj:
@jeeyi @ehoogerbeets maybe this PR will help with plural support since it is re-enables `PluralFormUpdater` as one time job to run. 

Let me know if you want to cut the release before it or if i should push now. 

@jeeyi  If you don't want this included, can you do the release soon so that I can push that?

PR #288, opened 2018-07-09
MOJI-653: support multi-selection for repos while creating new drops

ehoogerbeets:
- allow multi-select for repos so you can make drops for various repos at the same time
- iterates through the set of repos and creates a new drop for each one

### DIFF
--- a/webapp/src/main/resources/public/js/components/drops/Drops.js
+++ b/webapp/src/main/resources/public/js/components/drops/Drops.js
@@ -73,7 +73,7 @@ let Drops = React.createClass({
     },
 
     /**
-     * Fetch drops from action given the fitler status.
+     * Fetch drops from action given the filter status.
      *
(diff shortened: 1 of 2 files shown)


aurambaj:
Looks good! Great to have a new contributor 🎉 !!

Can you merge or do I need to do it? Also, if you (or I) can squash and merge, to avoid having 2 commits that'd be great.

Related to that change, I was wondering if you noticed that too. I don't really use the drops anymore but it looks like the alert ("New Translation Request is being sent") is not removed anymore after the drop is created. there is this `dismissAfter={5000}` in the code but it sticks around.

ehoogerbeets:
I have requested to be added to the Box organization, but it has not happened yet, so I cannot merge or squash this PR.

Yes, I did notice that those the alerts do not go away. This also happens after you import files as well as when you cancel drops.

PR #290, opened 2018-07-10
TM Export made asynchronous

jeeyi:
@aurambaj @ehoogerbeets Please review.

There is one more column *output* added to pollable_task table so that we can persist output of pollable task. I updated cli/webapp to export TM xliff asynchronously. We currently have server timeout error while exporting large TM xliff files because it's running synchronously.


aurambaj:
Are you missing the schema update script to add the column?

I think the original idea was to keep the PollableTask to just do the task management and persist data in other tables. Adding big blobs in pollable might not be the best.

So my suggestion would be to add a table just to store the exported xliff instead of adding "output" to pollable.

webapp/src/main/java/com/box/l10n/mojito/rest/asset/AssetWS.java
```diff
@@ -238,6 +243,24 @@ public XliffExportBody xliffExport(@PathVariable("assetId") long assetId, @Reque
         return new XliffExportBody(normalizedContent);
```
aurambaj: I think that function should be refactored to get the content from the the database now.

webapp/src/main/java/com/box/l10n/mojito/rest/asset/AssetWS.java
```diff
@@ -238,6 +243,24 @@ public XliffExportBody xliffExport(@PathVariable("assetId") long assetId, @Reque
...
+     * @param assetId {@link Asset#id}
+     * @param bcp47tag bcp47 tag of translations to be exported
+     * @return a {@link PollableTask} that generates XLIFF asynchronously in a
+     * {@link XliffExportBody}
+     */
+    @RequestMapping(method = RequestMethod.POST, value = "/api/assets/{assetId}/xliffExport")
+    public XliffExportBody xliffExportAsync(@PathVariable("assetId") long assetId,
+            @RequestParam("bcp47tag") String bcp47tag,
+            @RequestBody XliffExportBody xliffExportBody) {
+        PollableFuture pollableFuture = tmService.exportAssetAsXLIFFAsync(assetId, bcp47tag, PollableTask.INJECT_CURRENT_TASK);
```
aurambaj: Similarly to `AssetWS#importSourceAsset` that WS should create the initial `TMXliff` entity and return it along with the pollable id. 

Then the GET WS would take the id of the `TMxliff` entity and return it.

webapp/src/main/java/com/box/l10n/mojito/rest/pollableTask/PollableTaskWS.java
```diff
@@ -29,4 +35,24 @@
...
         return pollableTaskService.getPollableTask(pollableTaskId);
     }
+
+    /**
+     * Returns asynchronously generated XLIFF by {@link PollableTask}
+     *
+     * @param pollableTaskId
+     * @return an XLIFF that was generated by {@link PollableTask}
+     */
+    @RequestMapping(method = RequestMethod.GET, value = "/api/pollableTasks/{pollableTaskId}/xliffExport")
```
aurambaj: i'd get ride of that with the refactor of the old GET

webapp/src/main/resources/db/migration/V26__Add_tm_xliff.sql
```diff
@@ -0,0 +1,4 @@
+create table tm_xliff (id bigint not null auto_increment, created_date datetime, last_modified_date datetime, content longtext, asset_id bigint not null, export_pollable_task_id bigint, locale_id bigint, primary key (id));
```
aurambaj: file name `V26` --> `V28`, I'm push stuff with schema update too

jeeyi:
Yes. Importing translations from localized files is super useful if we are setting up new repo with source and localized files. For existing repo, tm export can be used to export all used/unused translations.

PR #292, opened 2018-07-11
Add buttons to perform action based on data from selected text unit

magezil:
- Enable creating either global or locale experiments on a string copy of a text unit
- Sends information about the text unit (string, repository, and language) to experiment site

### DIFF
--- a/chromeextension/background.js
+++ b/chromeextension/background.js
@@ -3,7 +3,8 @@ var backgroundConfig = {
     enabled: false,
     headerName: 'X-Mojito-Ict',
     headerValue: 'on',
-    removeTagsBlock: true
+    removeTagsBlock: true,
+    actionButtons: []
 };
 
 chrome.storage.sync.get(backgroundConfig, (items) => {
--- a/webapp/src/main/resources/public/js/ict/IctModal.js
+++ b/webapp/src/main/resources/public/js/ict/IctModal.js
@@ -40,10 +40,19 @@ class ModalTextUnit extends React.Component {
 };
 
 class IctModal extends React.Component {
-      
+    showButtons(actionButtons) {
+        var buttons = "";
+        if (Array.isArray(actionButtons)) {
+            buttons = actionButtons.map((button) =>
+                <Button key={button.text} onClick={() => this.props.onOpenAction(button.url)}>
+                    {button['text']}
+                </Button>
+            );
+        }
+        return buttons;
+    }
     render() {
-        
-        // include <style> here because css can't have precendence over <style>
+        // include <style> here because css can't have precedence over <style>
         // from apps that override fonts, ect
         return (
             <div>   
@@ -64,6 +73,7 @@ class IctModal extends React.Component {
                         )}           
                     </Modal.Body>
                     <Modal.Footer>
+                        {this.showButtons(this.props.actionButtons)}
                         <Button bsStyle="primary" onClick={this.props.onOkay}>
                             <FormattedMessage id="label.okay"/>
                         </Button>
--- a/webapp/src/main/resources/public/js/ict/IctModalContainer.js
+++ b/webapp/src/main/resources/public/js/ict/IctModalContainer.js
@@ -8,7 +8,8 @@ class IctModalContainer extends React.Component {
         super(props);
         this.state = {
             showModal: false,
-            mojitoBaseUrl: this.props.defaultMojitoBaseUrl
+            mojitoBaseUrl: this.props.defaultMojitoBaseUrl,
+            actionButtons: this.props.actionButtons
         };
     }
 
@@ -32,6 +33,12 @@ class IctModalContainer extends React.Component {
         });
     }
 
+    setActionButtons(actionButtons) {
+        this.setState({
+            actionButtons: actionButtons
+        });
+    }
+
     openWorkbench() {
         var query = queryString.stringify({
             'repoNames[]': this.state.selectedTextUnit.repositoryName,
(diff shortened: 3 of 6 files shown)


chromeextension/background.js
```diff
@@ -1,9 +1,21 @@
 var backgroundConfig = {
     mojitoBaseUrl: 'http://localhost:8080/',
+    experimentUrl: 'http://localhost:8080/copytune/',
     enabled: false,
     headerName: 'X-Mojito-Ict',
     headerValue: 'on',
-    removeTagsBlock: true
+    removeTagsBlock: true,
+    experimentButtons: {
```
aurambaj: let's make this a little bit more generic. `experimentButtons` --> `actionButtons`

chromeextension/background.js
```diff
@@ -1,9 +1,21 @@
 var backgroundConfig = {
     mojitoBaseUrl: 'http://localhost:8080/',
+    experimentUrl: 'http://localhost:8080/copytune/',
```
aurambaj: in our case it makes sense to share the base url but we can imagine that each button may call different system. So I wouldn't share the "url" and just have it in the buttons config

chromeextension/background.js
```diff
@@ -1,9 +1,21 @@
...
     enabled: false,
     headerName: 'X-Mojito-Ict',
     headerValue: 'on',
-    removeTagsBlock: true
+    removeTagsBlock: true,
+    experimentButtons: {
+        local: {
```
aurambaj: let's comment that out since in the open source version it won't do anything

webapp/src/main/resources/public/js/ict/IctModal.js
```diff
@@ -64,6 +62,18 @@ class IctModal extends React.Component {
                         )}           
                     </Modal.Body>
                     <Modal.Footer>
+                        {this.props.experimentButtons['global'] !== undefined &&
```
aurambaj: you can introduce a function here to "render" an action button. The code is duplicated here (global and locale)

And then to keep things generic you should iterates on the button list and render each button. Potentially there could be only 1 or maybe more than 3. We shouldn't hardcode the logic with 2 buttons

webapp/src/main/resources/public/js/ict/IctModalContainer.js
```diff
@@ -43,13 +57,33 @@ class IctModalContainer extends React.Component {
...
+    createGlobalExperiment() {
+        var query = queryString.stringify({
+            'repoName': this.state.selectedTextUnit.repositoryName,
+            'searchText': this.state.selectedTextUnit.textUnitName
+        });
+        window.open(this.state.experimentUrl + this.state.experimentButtons['global']['urlCallback'] + '&' + query);
+    }
+
+    createLocalExperiment() {
+        var query = queryString.stringify({
```
aurambaj: so I guess in the generic version will end up passing an extra parameter for global but that won't be a big deal

webapp/src/main/resources/public/js/ict/chrome-ict-popup.js
```diff
@@ -25,6 +25,15 @@ const IctPopup = function (props) {
                 <FormControl type="text" value={props.mojitoBaseUrl} onChange={(e) => props.onMojitoBaseUrlChanged(e.target.value)} />
               </Col>
             </FormGroup>
+
+            <FormGroup controlId="formHorizontalMojitoUrl">
```
aurambaj: oh, you made that configurable?  I think it makes things a little bit more complicated and we don't really need it for now. I'd suggest removing it

webapp/src/main/resources/public/js/ict/IctModal.js
```diff
@@ -40,10 +40,19 @@ class ModalTextUnit extends React.Component {
 };
 
 class IctModal extends React.Component {
-      
+    showButtons(actionButtons) {
+        var buttons = "";
+        if (actionButtons !== []) {
+            buttons = actionButtons.map((button) =>
+                <Button bsStyle="primary" key={button.text} onClick={() => this.props.onOpenAction(button.url)}>
```
aurambaj: let's remove "primary" we should just have one primary button

webapp/src/main/resources/public/js/ict/IctModalContainer.js
```diff
@@ -43,13 +50,24 @@ class IctModalContainer extends React.Component {
         window.open(this.state.mojitoBaseUrl + "workbench?" + query);
     }
 
+    openActionInNewWindow(url) {
+        var query = queryString.stringify({
+            'repoName': this.state.selectedTextUnit.repositoryName,
+            'locales': this.state.selectedTextUnit.locale,
+            'searchText': this.state.selectedTextUnit.textUnitName
+        });
+        window.open(url + '&' + query);
```
aurambaj: '&' should be '?' which is delimiter between the base url and the query string, no?

webapp/src/main/resources/public/js/ict/IctModal.js
```diff
@@ -40,10 +40,19 @@ class ModalTextUnit extends React.Component {
 };
 
 class IctModal extends React.Component {
-      
+    showButtons(actionButtons) {
+        var buttons = "";
+        if (actionButtons !== []) {
```
aurambaj: can you try that instead: Array.isArray(actionButton)

webapp/src/main/resources/public/js/ict/IctModal.js
```diff
@@ -40,10 +40,19 @@ class ModalTextUnit extends React.Component {
...
+            buttons = actionButtons.map((button) =>
+                <Button bsStyle="primary" key={button.text} onClick={() => this.props.onOpenAction(button.url)}>
+                    <FormattedMessage id={button['text']}/>
```
aurambaj: you can drop `<FormattedMessage` since we won't have translation for that

webapp/src/main/resources/public/js/ict/IctModalContainer.js
magezil: The url I was using actually already had an initial query parameter in it. I'll think about updating it.

PR #293, opened 2018-07-11
MOJI-677: add a confirmation dialog box while cancelling a drop

ehoogerbeets:
Make sure the user does not accidentally cancel the drop by throwing up a simple confirmation dialog first.

### DIFF
--- a/webapp/src/main/resources/properties/en.properties
+++ b/webapp/src/main/resources/properties/en.properties
@@ -370,6 +370,18 @@ drops.controlbar.button.tooltip.import=Import Request
 # Tooltip on cancel drop button
 drops.controlbar.button.tooltip.cancel=Cancel Request
 
+# Title of the confirmation dialog box when a user attempts to cancel a review or translation request
+drops.confirm.cancel.model.title=Really Cancel?
+
+# Body of the confirmation dialog box when a user attempts to cancel a review or translation request
+drops.confirm.cancel.model.body=Are you sure you want to cancel this request?
+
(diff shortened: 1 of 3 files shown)


jeeyi:
@ehoogerbeets Could you attach the screenshot showing the dialog box?

ehoogerbeets:
screen shot 2018-07-11 at 13 42 14

webapp/src/main/resources/properties/en.properties
```diff
@@ -370,6 +370,18 @@ drops.controlbar.button.tooltip.import=Import Request
 # Tooltip on cancel drop button
 drops.controlbar.button.tooltip.cancel=Cancel Request
 
+# Title of the confirmation dialog box when a user attempts to cancel a review or translation request
+drops.confirm.cancel.model.title=Really Cancel?
```
jeeyi: Can we replace "Really Cancel?" with "Confirm cancel"? (to be consistent with the Confirm delete dialog in workbench).
confirmdelete

webapp/src/main/resources/public/js/components/drops/Drops.js
```diff
@@ -64,6 +65,13 @@ let Drops = React.createClass({
 
             /** @type {Drop} */
             "selectedDrop": null,
+
+            /** @type {Boolean} */
+            "showCancelModal": false,
+            /** @type {Number} */
+            "cancelDropId": null,
+            /** @type {Number} */
+            "cancelRepoId": null
```
jeeyi: Why do we need cancelRepoId?

PR #301, opened 2018-07-16
Fix typo that was causing asset_path to be undefined in query

magezil:
`asset_path` in the query should be assigned to `this.state.selectedTextUnit.assetName`, otherwise it will be assigned `undefined`


webapp/src/main/resources/public/js/ict/IctModalContainer.js
```diff
@@ -55,7 +55,7 @@ class IctModalContainer extends React.Component {
             'repoName': this.state.selectedTextUnit.repositoryName,
             'locales': this.state.selectedTextUnit.locale,
             'searchText': this.state.selectedTextUnit.textUnitName,
-            'asset_path': this.state.selectedTextUnit.asset_path
+            'asset_path': this.state.selectedTextUnit.assetName
```
aurambaj: Actually we should update the query parameter too for consistency and change the analytics app

magezil:
PR is not updating with most recent push.

PR #303, opened 2018-07-19
Multiple plurals

jeeyi:
Let's say the source file has plurals with only "other" item like below:
```
<plurals name="numberOfPeople">
    <item quantity="other">%1$d people</item>
</plurals>
```
The skeleton parts of the text unit for the above is:
part1 => `<plurals name="numberOfPeople">\n    <item quantity="other">`
part2 => `%1$d people`
part3 => `</item>`

If we create other items by copying skeleton of "other" item, we end up with something like below that ends up with malformed xml:
```
<plurals name="numberOfPeople">
    <item quantity="one">%1$d people</item>
<plurals name="numberOfPeople">
    <item quantity="other">%1$d people</item>
</plurals>
```

We should be updating skeleton so that any added items have three skeleton parts:
part1 => `<item quantity="few">`
part2 => `%1$d people`
part3 => `</item>`


webapp/src/main/java/com/box/l10n/mojito/okapi/filters/AndroidFilter.java
```diff
@@ -2,11 +2,13 @@
 
 import com.box.l10n.mojito.okapi.CopyFormsOnImport;
 import com.box.l10n.mojito.okapi.TextUnitUtils;
+import static com.box.l10n.mojito.okapi.filters.PluralsHolder.logger;
```
aurambaj: Want to remove that logger

jeeyi: Done

PR #311, opened 2018-08-06
MOJI-629: Keep track of the user that made a tm variant

ehoogerbeets:
... by adding an entity listener to the settable auditable entry.


ehoogerbeets:
No! Not sure! I tried this and it worked for me in this case, but I wasn't sure what the other consequences would be, as I am new to Spring and of course Mojito too. Jee said she would look into this more and try setting the user based on the currently-logged-in user info when storing the variant. I'll close this one.

PR #313, opened 2018-08-08
Added locales param to tm-export command

jeeyi:


cli/src/main/java/com/box/l10n/mojito/cli/command/TMExportCommand.java
```diff
@@ -43,6 +43,9 @@
     @Parameter(names = {Param.REPOSITORY_LONG, Param.REPOSITORY_SHORT}, arity = 1, required = true, description = Param.REPOSITORY_DESCRIPTION)
     String repositoryParam;
 
+    @Parameter(names = {"--locales", "-l"}, arity = 1, required = false, description = "List of locales to be exported, format: fr-FR,ja-JP")
```
aurambaj: you may want to introduce constant for those param since it is shared with `DropExportCommand`

jeeyi: Done

PR #320, opened 2018-08-31
Async Jobs improvement

aurambaj:
Move the in-memory async jobs to use Quartz. This allow to retry jobs if the application is stopped in the middle of their executions (deploy or crash). Quartz must be configured to use Mysql else it will run in memory and the jobs won't recover from a restart


webapp/src/main/resources/config/application.properties
```diff
@@ -54,9 +57,22 @@ cli.file=mojito-cli.jar
 
 # Quartz
 l10n.org.quartz.scheduler.skipUpdateCheck=true
-l10n.org.quartz.jobStore.useProperties=true
-l10n.org.quartz.scheduler.instanceId=AUTO
-l10n.org.quartz.jobStore.isClustered=true
+
+### for Mysql, cluster deployment
+#l10n.org.quartz.jobStore.useProperties=true
```
aurambaj: @jeeyi That's the configuration for Quartz that you need if you deploy a cluster or if you want the job to be retried if the server is restarted in the middle of the job execution. If no setting is added in the conf. it should use in-memory job scheduling and behave like with the old implementation (in memory is not working for a cluster though)

PR #323, opened 2018-09-13
support for ellipsis entity in Android resource file

jeeyi:
Android team has asked that we support `&#8230;` instead of &#8230; for elipsis. The benefits are:

- Accessibility: the talk back does not say: dot dot dot
- You are sure that the OS will not introduce a return carriage in the middle of the dots

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/okapi/filters/XMLEncoder.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/okapi/filters/XMLEncoder.java
@@ -64,6 +64,7 @@ public class XMLEncoder extends net.sf.okapi.common.encoder.XMLEncoder {
             replacement = "<$2$3>";
         }
         text = ANDROID_HTML.matcher(text).replaceAll(replacement);
+        text = text.replace("…", "&#8230;");
         text = text.replaceAll("\n", "\\\\n");
         text = text.replaceAll("\r", "\\\\r");
         text = escapeDoubleQuotes(text);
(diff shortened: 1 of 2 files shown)


jimmy-holzer-box:
@aurambaj @jeeyi 
Yes we changed '...' with '&#8230;' in the source file (https://git.dev.box.net/Mobile/android-app/commit/f96398afe56b797c564920ce236deae9a502843f)
 
Few days later we got this PR back: https://git.dev.box.net/Mobile/android-app/commit/3c3bfa3a25a767d0bbf8870481efac3d45719aa4
where '...' was still present in all translations. Minor detail but I also noticed that the spacing between the dots are not always the same (between languages or even inside the same language) E.g. In bn: Sort_by key and open_in key

Any way we can keep &#8230 when it's detected in the source file? 

Thanks for looking at it!

jeeyi:
@aurambaj, just to be clear, ellipsis character `…` is same as the unicode code point `&#8230;`. The ellipsis character `…` is a one character that is different from three dots `...`. I am not replacing three dots with the ellipsis character  `&#8230;`. I am just replacing the character representation with unicode code point.

I am checking with @jimmy-holzer-box if the use of ellipsis character `…` works the same way as using the unicode code point  `&#8230;`. If so, there is no need to force the unicode code point  `&#8230;` when we write localized resource files.

I agree with you to implement it in the integrity checker to prevent translators to use three dots  `...` instead of the ellipsis character  `…`.

aurambaj:
@jeeyi oh... got it now! You may want to mention in the documentation that it is the unicode character and not the 3 dot because reading the PR I was mislead!

But then I'm confused with the comment in the test. I'd be surprise if having unicode character or the entity makes a difference. I suspect the accessibility and carriage return problem would happen with a string having 3 '.'

So it looks like in that case it would be about replacing 3 '.' with unicode eclipse character in the translations, no?

jeeyi:
@aurambaj, @jimmy-holzer-box just comfirmed that having elippse character `…` works fine just as if we have `&#8230;` in the localized resource files. I am closing this pull request and will create a separate PR for the integrity checker.

PR #326, opened 2018-09-17
fixing intermittent screenshot tests failures

jeeyi:
@aurambaj I sometimes run into intermittent test failures from ScreenshotServiceTest. The failures are coming from tests that query screenshots for specific repository sometimes return two more screenshots than expected. 

It only happens when I run all unit tests in webapp. If I only run ScreenshotServiceTest, it works fine.

I also noticed that it works if I comment out the test in ScreenshotWSTest or assign the name for the screenshot run. What happens if ScreenshotRun name is null? Do you know why it causes intermittent failures?


aurambaj:
hm.. not sure at all. the query filters by repository id which should be specific to the test execution since it uses the `testIdWatcher`. The `screenshotRun` name should not matter much but we can try with that if it helps.

PR #330, opened 2018-09-19
Add git-blame functionality for cli

magezil:
git-blame command takes the repository, source file directory, and file type to extract the git blame information and save it with the text unit information. 
- GitBlameCommand extracts the author name, author email, commit id, and commit time and saves it with the text unit in a GitBlameWithUsage object.
- TextUnitWithUsageClient interacts with the TextUnitWS to extract the text units to blame and save the git-blame information into the db

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/GitBlameCommand.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/GitBlameCommand.java
@@ -1,0 +1,298 @@
+package com.box.l10n.mojito.cli.command;
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
+import com.box.l10n.mojito.cli.ConsoleWriter;
+import com.box.l10n.mojito.cli.command.param.Param;
+import com.box.l10n.mojito.cli.filefinder.FileMatch;
+import com.box.l10n.mojito.cli.filefinder.file.FileType;
+import com.box.l10n.mojito.rest.client.AssetClient;
+import com.box.l10n.mojito.rest.client.RepositoryClient;
+import com.box.l10n.mojito.rest.client.TextUnitWithUsageClient;
+import com.box.l10n.mojito.rest.client.exception.PollableTaskException;
+import com.box.l10n.mojito.rest.entity.*;
+import org.eclipse.jgit.api.BlameCommand;
+import org.eclipse.jgit.api.errors.GitAPIException;
+import org.eclipse.jgit.blame.BlameResult;
+import org.eclipse.jgit.lib.ObjectId;
+import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
+import org.fusesource.jansi.Ansi;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.text.MessageFormat;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * @author jaurambault
+ */
+@Component
+@Scope("prototype")
+@Parameters(commandNames = {"git-blame", "gb"}, commandDescription = "Git blame")
+public class GitBlameCommand extends Command {
+
+    /**
+     * logger
+     */
+    static Logger logger = LoggerFactory.getLogger(GitBlameCommand.class);
+
+    @Autowired
+    ConsoleWriter consoleWriter;
+
+    @Parameter(names = {Param.REPOSITORY_LONG, Param.REPOSITORY_SHORT}, arity = 1, required = true, description = Param.REPOSITORY_DESCRIPTION)
+    String repositoryParam;
+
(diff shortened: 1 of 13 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/GitBlameCommand.java
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    /**
+     * Reads in the lines of the file and runs git-blame on usage locations given by file
+     * @param gitBlameWithUsages
+     * @throws CommandException
+     */
+    void blameWithTextUnitUsages(List<GitBlameWithUsage> gitBlameWithUsages) throws CommandException {
+
+        for (GitBlameWithUsage textUnitWithUsage : gitBlameWithUsages) {
```
aurambaj: rename variable

cli/src/main/java/com/box/l10n/mojito/cli/command/GitBlameCommand.java
```diff
@@ -0,0 +1,298 @@
...
+    public void execute() throws CommandException {
+
+        commandDirectories = new CommandDirectories(sourceDirectoryParam);
+
+        consoleWriter.newLine().a("Git blame for repository: ").fg(Ansi.Color.CYAN).a(repositoryParam).println(2);
+
+        Repository repository = commandHelper.findRepositoryByName(repositoryParam);
+        List<PollableTask> pollableTasks = new ArrayList<>();
+
+        List<GitBlameWithUsage> textUnitsToBlame = textUnitWithUsageClient.getTextUnitToBlame(repository.getId());
```
aurambaj: Want to handle pagination here.

cli/src/main/java/com/box/l10n/mojito/cli/command/GitBlameCommand.java
```diff
@@ -0,0 +1,298 @@
...
+
+        if (fileType.getSourceFileExtension().equals("pot")) {
```
aurambaj: it is not the best to check the extension in that case, instead you can check the file type with something like `if (fileType instanceOf POFileType)`.

There will be other type that will behave like PO file so it would make sense to move that as an attribute in the FileType but it can be done in another PR

cli/src/main/java/com/box/l10n/mojito/cli/command/GitBlameCommand.java
```diff
@@ -0,0 +1,298 @@
...
+        }
+    }
+
+    /**
+     * Checks if the given line contains text unit(s)
+     * @param line
+     * @param gitBlameWithUsages
+     * @return list of GitBlameWithUsage objects that match current line
+     */
+    List<GitBlameWithUsage> getTextUnitNameFromLine(String line, List<GitBlameWithUsage> gitBlameWithUsages) {
```
aurambaj: the function should be rename since not it return a list of GitBlameWithUsage for that line and not a text unit name anymore

cli/src/main/java/com/box/l10n/mojito/cli/command/GitBlameCommand.java
```diff
@@ -0,0 +1,298 @@
...
+        }
+        return gitBlameWithUsagesWithLine;
+    }
+
+    /**
+     * Converts text unit name back to how name appears in source code
+     * @param textUnitName
+     * @return text unit name as string in source file
+     */
+    static String textUnitNameToStringInSourceFile(String textUnitName) {
```
aurambaj: let's not do `static` since we use Spring DI

cli/src/main/java/com/box/l10n/mojito/cli/command/GitBlameCommand.java
```diff
@@ -0,0 +1,298 @@
...
+        return stringInFile;
+
+    }
+
+    /**
+     * Extracts the file name from the given usage
+     * @param usage
+     * @return
+     */
+    static String getFileName(String usage) {
```
aurambaj: same

restclient/src/main/java/com/box/l10n/mojito/rest/client/TextUnitWithUsageClient.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author emagalindan
+ */
+@Component
+public class TextUnitWithUsageClient extends BaseClient {
```
aurambaj: Let's rename the client

restclient/src/main/java/com/box/l10n/mojito/rest/client/TextUnitWithUsageClient.java
```diff
@@ -0,0 +1,48 @@
...
+        return "textunits";
+    }
+
+    public List<GitBlameWithUsage> getTextUnitToBlame(Long repositoryId) {
+
+        Map<String, String> filterParams = new HashMap<>();
+
+        filterParams.put("repositoryIds[]", repositoryId.toString());
+
+        return authenticatedRestTemplate.getForObjectAsListWithQueryStringParams(
```
aurambaj: we need to handle pagination here else it will just fetch 10 entries.

aurambaj:
🎊 Thanks Liz

PR #335, opened 2018-09-27
Fix GitBlameCommand.getGitRepository function

magezil:
Fix NullPointerException thrown when -s parameter is not specified

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/GitBlameCommand.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/GitBlameCommand.java
@@ -225,7 +225,7 @@ public class GitBlameCommand extends Command {
 
             try {
                 gitRepository = builder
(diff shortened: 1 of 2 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/GitBlameCommand.java
```diff
@@ -225,7 +225,7 @@ private void getBlameResults(int lineNumber, BlameResult blameResultForFile, Git
 
             try {
                 gitRepository = builder
-                        .findGitDir(new File(sourceDirectoryParam))
+                        .findGitDir(new File(commandDirectories.sourceDirectoryPath.toString()))
```
aurambaj: please change to use the getter: `commandDirectories.getSourceDirectoryPath().toString()`

magezil: Oops 🤦‍♀️

PR #342, opened 2018-10-08
Add front-end for git-blame information in Workbench

magezil:
When the information button is clicked, displays information about the current text unit:
- text unit information
- git blame information
- debug/additional text unit information

### DIFF
--- a/webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
+++ b/webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
@@ -173,19 +173,18 @@ let GitBlameInfoModal = React.createClass({
         if (textUnit == null || usages == null)
             return null;
 
+        if (usages.length === 0) {
+            usages = [textUnit.getAssetPath()];
+        }
+
         if (this.props.appConfig.opengrok.server === null)
             return this.getUsages().join('\n');
 
         let repo = textUnit.getRepositoryName();
 
-        if (usages.length === 0) {
-            if (repo === "android")
-                usages = [textUnit.getAssetPath()];
-            else
-                return null;
-        }
         let links = [];
         for (let usage of usages) {
+            usage = usage.replace(this.props.appConfig.opengrok.extractedFilePrefix, "");
             let link = this.props.appConfig.opengrok.server;
             for (let repoKey in this.props.appConfig.opengrok.repositoryMapping){
                 if (repo === repoKey) {
--- a/webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
+++ b/webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
@@ -1,0 +1,255 @@
+import PropTypes from 'prop-types';
+import React from "react";
+import {FormattedMessage, injectIntl} from "react-intl";
+import {Button, Panel, Accordion, Modal} from "react-bootstrap";
+import {withAppConfig} from "../../utils/AppConfig";
+
+
+let GitBlameInfoModal = React.createClass({
+
+    propTypes() {
+        return {
+            "show": PropTypes.bool.isRequired,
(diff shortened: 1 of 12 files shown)


webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
```diff
@@ -0,0 +1,255 @@
...
+
+    /**
+     * @returns {*} Converts the list of usages of the text unit to a list of links to given text unit
+     */
+    getOpenGrokLocation() {
+        let textUnit = this.props.textUnit;
+        let usages = this.getUsages();
+
+        if (textUnit == null || usages == null)
+            return null;
```
aurambaj: I'd use `{ }`

webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
```diff
@@ -0,0 +1,255 @@
...
+
+        if (this.props.appConfig.opengrok.server === null)
+            return this.getUsages().join('\n');
+
+        let repo = textUnit.getRepositoryName();
+
+        if (usages.length === 0) {
+            if (repo === "android")
```
aurambaj: we can't have an hardcoded repository name. We need to find another solution. We can implement that in another PR

webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
```diff
@@ -0,0 +1,255 @@
...
+                usages = [textUnit.getAssetPath()];
+            else
+                return null;
+        }
+        let links = [];
+        for (let usage of usages) {
+            let link = this.props.appConfig.opengrok.server;
+            for (let repoKey in this.props.appConfig.opengrok.repositoryMapping){
```
aurambaj: probably want to introduce sub functions this getting long/complicated

webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * @param date   An integer representing a datetime
+     * @returns {*}  Human readable version of the given datetime
+     *
+     */
+    convertDateTime(date) {
+        if (date === null || isNaN(date))
+            return null;
+
+        return (this.props.intl.formatDate(date) + ", " + this.props.intl.formatTime(date));
```
aurambaj: isn't there a `formatDateTime` function?

webapp/src/main/resources/public/js/stores/workbench/GitBlameStore.js
```diff
@@ -0,0 +1,52 @@
...
+
+    openWithTextUnit(textUnit) {
+        console.log("GitBlameStore::openWithTextUnit");
+
+        this.show = true;
+        this.textUnit = textUnit;
+        this.gitBlameWithUsage = null;
+        this.loading = true;
+        this.getInstance().getGitBlameInfo(textUnit);
+
```
aurambaj: extra return line

webapp/src/main/resources/sass/mojito.scss
```diff
@@ -167,6 +167,7 @@ a.label-container:hover {
 /** MODAL */
 .modal-content {
     box-shadow: 0 2px 6px rgba(0, 0, 0, 0.1);
+    word-wrap: break-word;
```
aurambaj: isn't that the default value? Do we need it/can we remove it?

magezil: looks like it can be removed. There was text spilling out of the modal before, but it was probably caused by something else.

webapp/src/main/resources/public/js/stores/workbench/GitBlameStore.js
magezil: 🙀

PR #354, opened 2018-11-02
Pseudolocalization

jeeyi:


.gitignore
```diff
@@ -1,6 +1,7 @@
 /docker/.data
 /cli/target/
 /webapp/target/
+/webapp/package-lock.json
```
jeeyi: @aurambaj We don't need this file to be checked into git, correct?

aurambaj: yes it is needed it git, don't exclude it please.

webapp/src/main/java/com/box/l10n/mojito/service/assetintegritychecker/integritychecker/MessageFormatIntegrityChecker.java
```diff
@@ -23,13 +20,15 @@
...
  */
-public class MessageFormatIntegrityChecker implements TextUnitIntegrityChecker {
+public class MessageFormatIntegrityChecker extends AbstractTextUnitIntegrityChecker {
 
     /**
      * logger
      */
     static Logger logger = LoggerFactory.getLogger(MessageFormatIntegrityChecker.class);
 
+    private static final String REPLACEMENT_CHARS = "!!";
```
aurambaj: might want to use some special unicode character to make it safer, like `\x13`

.gitignore
jeeyi: Okay. I have reverted deletion of package-lock.json changes now.

webapp/src/main/java/com/box/l10n/mojito/service/assetintegritychecker/integritychecker/MessageFormatIntegrityChecker.java
jeeyi: Good point!

jeeyi:
@aurambaj I can't remember why I didn't use okapi code detection. Let me try with okapi. If it works, then it may be cleaner solution as you said.

jeeyi:
@aurambaj Named arguments like `{username}` are not detected as inline code. Only indexed arguments like `{1}` are detected as expected. In order to make it work, we need to customize rules in codeFinder. This isn't simple for plural message format.

PR #358, opened 2018-11-20
Fix import localized asset when some text unit are unused.

aurambaj:
The new implementation import the translation for the current/used text unit that has the given name. Previous implementation was not checking the used status and may have imported the translation on an unused text unit with same name.


webapp/src/main/java/com/box/l10n/mojito/okapi/ImportTranslationsFromLocalizedAssetStep.java
```diff
@@ -48,7 +48,9 @@
         REVIEW_NEEDED,
         TRANSLATION_NEEDED,
         APPROVED
-    };
+    }
+
+    ;
```
magezil: I don't know if this was intended, but it looks really weird.

aurambaj: nope, bad intellij auto-indent. Fixed

PR #360, opened 2018-11-21
Add support for SCM branches in Mojito.

aurambaj:
To support "usual" localization processes the following decision was made: branches share the tm text units. Changing the translation in one branch will change the translation in all branches. Note that for a given key if the source string is changed a new text unit is created and the translations will be independent. For normal branch support in applications this is the expected behavior: translation fix should apply to all branches but if changes are made to a source string or given identifier the translations can't be shared anymore since for example a change of placeholder could break the application.

If this behavior is not acceptable, the repository can be cloned instead but the TMs would diverge from that point and no tool are provided to merge branches.

To start working with branches just provide a "branch" parameter to the "push" command. Everything else stays the same.

For the reason provided before the "pull" command is not sensitive to branches. The output may differ if the branch contains a different string set but since the text units are shared there is no need to provide the branch name.

When working with branches, the text units seen in the workbench (= last asset successful asset extraction) are a merge of the text units coming from all branches. The origin of the text unit can be seen in the workbench. If a string is present in multiple branch, the "master" branch will be shown first and then it is based on time of creation of the branch.

An advantage of the merge logic is that branches may only push a subset of strings to a branch. This is useful with pull request to only work on new strings (if coming from an extractor run on the diff). They will just add up to the repository. When the pull request is merged the new strings will be picked up right away say for the master branch.

Detailed changes:
- Change "push" command to accept "branch parameter"
- Show the branch in the workbench text unit details
- Add WS to list branches: /api/repositories/{repositoryId}/branches and modify existing WS to accept branch filter
- Change rest clients: AssetClient and RepositoryClient to support new service and change to "push" command
- Database change to support branches: AssetContent (content is now store for each branch. before it was stored at the asset level), Branch (branch in the repository), AssetExtractionByBranch (keeps track of the last successful extraction per branch. This complete the last successful extraction at the asset level) entities
- Refactor JSON format for asset see view "AssetSummary" (to much was inlined before and the payload would get artificially huge) and improved JsonBackReference
- Change the Asset WS and service to support branches:
   - create asset now create an asset entity along with an asset content
   - AssetService#addOrUpdateAssetAndProcessIfNeeded() takes care of creating branches if needed an link the assetContent to a branch. If no branch is supplied a branch with name "null" will be created
   - AssetExtractionService#processAsset() will create asset extraction per branch and look if multiple branches exist for an asset. If so, it will create a merged asset extraction that is set as last successful asset extraction for the asset
   - The logic to delete/un-delete asset now takes in account branches
   - Cleanup asset extractions for each branch (before was only doing it at the asset level)


jeeyi:
@aurambaj This change backward compatible, correct? Branch parameter is optional and if we are not using it, there shouldn't be any differences. 

What about the existing content and content_md5 that are stored in asset table? Should we migrate them to the new asset_content table? Or do we just let push take care of it?

webapp/src/main/java/com/box/l10n/mojito/entity/Repository.java
```diff
@@ -58,7 +59,7 @@
     private DropExporterType dropExporterType;
 
     @JsonView(View.RepositorySummary.class)
-    @JsonManagedReference
+    @JsonManagedReference("repositoryLocales")
```
magezil: Just curious, why do you have a string after this annotation, but not some others, like the one in line 76 below?

magezil: Is it just so it will match the `@JsonBackReference` in `RepositoryLocale.java`?

aurambaj:
> @aurambaj This change backward compatible, correct? Branch parameter is optional and if we are not using it, there shouldn't be any differences.

@jeeyi Yes it should be backward compatible but I'll run more tests before merging to master. There shouldn't be any difference but after a new push the entities used to store the data will be different and the content/contentMd5 columns in the asset entity won't be used anymore. Ideally they could be cleanup. 

> 
> What about the existing content and content_md5 that are stored in asset table? Should we migrate them to the new asset_content table? Or do we just let push take care of it?

I kept those there for backward compatibility. They could be remove once the all the repo are migrated to the new entities. For now I just relaxed the constrain on the content column

PR #362, opened 2018-11-29
We should skip the non-existing text unit and continue importing tran…

jeeyi:
…slations

@aurambaj We ran into issues while importing exported tm xliffs because some of the xliffs had translation for non-existing translation units. I think skipping them is better than throwing run-time exception and aborting the whole import process. Let me know what you think.

Do you know why we have some translations in the exported localized tm xliffs for the text unit that is not in the source tm xliff?

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/tm/ImportExportedXliffStep.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/tm/ImportExportedXliffStep.java
@@ -271,47 +271,45 @@ public class ImportExportedXliffStep extends BasePipelineStep {
 
             Long tmTextUnitId = tmTextUnitIdsByMd5ForAsset.get(tmService.computeTMTextUnitMD5(name, sourceContent, importExportNote.getSourceComment()));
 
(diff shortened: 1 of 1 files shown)


aurambaj:
@jeeyi With the current implementation, I think the hard error make sense. We don't remove text units as far as I know so there must be a bug somewhere else. 

Did you check if the entry was in the DB?

jeeyi:
@aurambaj The text unit differences come from plurals. In English tm xliff, I have text unit `*_one` and `*_other`. In localized tm xliff for `ar-SA`, I have more text units according to the plural rules. `tm-import` gets terminated when it tries to import translation for `*_zero` because it is not in the English tm xliff and considered as adding a translation to a non-existing text unit.

aurambaj:
oh when importing plural strings for the source it should actually create all forms, that's the issue. Your fix can be a temporary solution, though

Why do still use that feature btw, what's the use case?

aurambaj:
@jeeyi btw, if you re-generate the source file they should have all the plural form, right?

jeeyi:
@aurambaj You mean, if I push the source file? Let me test it. I think it should work.

PR #364, opened 2018-12-04
Make the source locale configurable when creating a repository

aurambaj:
There is probably more work to make this fully usable but this first commit removes hard coded reference to 'en' as default locale in the backend. The `repo-create` CLI now takes a source language param


aurambaj:
@chreck FYI

PR #365, opened 2018-12-04
some refactoring to reuse code to escape/unescape double quotes

jeeyi:
@aurambaj JSFilter needs same double quote escaping as MacStringsFilter. So I refactored the code so that they share the same code.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/okapi/filters/RegexEscapeDoubleQuoteFilter.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/okapi/filters/RegexEscapeDoubleQuoteFilter.java
@@ -1,0 +1,47 @@
+package com.box.l10n.mojito.okapi.filters;
+
+import net.sf.okapi.common.Event;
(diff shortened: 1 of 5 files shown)


webapp/src/main/java/com/box/l10n/mojito/okapi/filters/RegexEscapeDoubleQuoteFilter.java
```diff
@@ -0,0 +1,47 @@
...
+            TextContainer source = new TextContainer(unescapedSourceString);
+            textUnit.setSource(source);
+        }
+        return event;
+    }
+
+    protected String unescape(String text) {
+        String unescapedText = text.replaceAll("(\\\\)(\")", "$2");
+        unescapedText = unescapedText.replaceAll("\\\\n", "\n");
+        unescapedText = unescapedText.replaceAll("\\\\r", "\r");
```
aurambaj: We could move those to a unescape `@Component` as well

PR #369, opened 2018-12-06
Keep track of the branch creator/owner.

aurambaj:
For now the creator comes form the security context. We'll provide and option in the "push" command later to override (use to push branch on behalf of a user).

This commit just adds the createdByUser attribute to the branch entity and adds logic to set it when processing an asset. It will creates a partial user if needed.

When automation jobs process a branch on behalf of a user, that user may not be in Mojito yet. It will create a partial user (since the name is only available at that point) which will be updated with more information the first time the user logs in.


webapp/src/main/java/com/box/l10n/mojito/service/asset/AssetService.java
```diff
@@ -146,7 +155,18 @@
             pollableFutureTaskResult.setExpectedSubTaskNumberOverride(1);
         }
 
-        Branch branch = branchService.getOrCreateBranch(asset.getRepository(), branchName);
+        //TODO pass the owner information, merge Xiaye's PR
```
aurambaj: @xiayewang can integrate with your PR here

PR #377, opened 2018-12-12
Only display a translated date if the text unit has been translated

magezil:
Previously, the translated date would show even if the text unit had not been translated


webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
```diff
@@ -354,6 +354,15 @@ let GitBlameInfoModal = React.createClass({
         return this.getCommitName();
     },
 
+    /**
+     * @returns {*} The translated date if the text unit has been translated
+     */
+    getTranslatedDate() {
+        if (this.props.textUnit.getTarget() != null && this.props.textUnit.getTmTextUnitCurrentVariantId != null) {
```
aurambaj: checking one is enough, let's keep `getTmTextUnitCurrentVariantId()`, need to add `()`

PR #379, opened 2018-12-14
Refactor extract usage from comment into component for mac file types

magezil:
Since locations are formatted in the comments in the same way for mac strings and mac stringsdict files, refactored the extract usage method into a component


webapp/src/main/java/com/box/l10n/mojito/okapi/ExtractUsagesFromTextUnitComments.java
```diff
@@ -0,0 +1,105 @@
...
+        textUnit.setAnnotation(new UsagesAnnotation(usagesFromComment));
+    }
+
+    /**
+     * Get and delete the usage locations from the comment in textUnit.
+     *
+     * @param textUnit to get the usages from
+     * @return the locations or empty set
+     */
+    protected Set<String> extractUsagesFromTextUnit(TextUnit textUnit) {
```
aurambaj: let's use the same method scope for all method, I like "package-protected" those days ... which is no keyword

`protected Set<String> extractUsagesFromTextUnit(TextUnit textUnit) {` -->` Set<String> extractUsagesFromTextUnit(TextUnit textUnit) {`

webapp/src/main/java/com/box/l10n/mojito/okapi/filters/MacStringsFilter.java
```diff
@@ -14,14 +20,17 @@
...
+
     @Override
     public String getName() {
         return FILTER_CONFIG_ID;
     }
 
     @Override
     public List<FilterConfiguration> getConfigurations() {
-        List<FilterConfiguration> list = new ArrayList<FilterConfiguration>();
+        List<FilterConfiguration> list = new ArrayList<>();
```
aurambaj: 👍

PR #404, opened 2019-02-22
Branch stats

aurambaj:


aurambaj:
@jeeyi heads up I'm going to merge that soon.

I've added JSON support and put the filter as one of the default supported filters. Since it is a common format, this may lead to some JSON files to be processed unexpectedly. I think it make sense to have it enabled by default similarly to what we do for other types but it could be a slight annoyance to reconfigure the filter types in the CLI calls. Thoughts?

PR #407, opened 2019-02-26
dedupe asset text unit in master when pushing new branch

xiayewang:
There can be lots of duplicate text units which already exist in master branch when new branch is being pushed. This is to dedupe asset text unit.


aurambaj:
as discussed we don't want the extraction step to be dependent of the db state. We cleanup the asset extraction to keep the table size small

PR #408, opened 2019-02-28
Add tmTextUnitIDs to search params

pperozo:
So that we can use these params in the api add them to the search params


webapp/src/main/java/com/box/l10n/mojito/rest/textunit/TextUnitWS.java
```diff
@@ -220,6 +220,7 @@ TextUnitSearcherParameters queryParamsToTextUnitSearcherParameters(
         textUnitSearcherParameters.setUsedFilter(usedFilter);
         textUnitSearcherParameters.setStatusFilter(statusFilter);
         textUnitSearcherParameters.setDoNotTranslateFilter(doNotTranslateFilter);
+        textUnitSearcherParameters.setTmTextUnitIds(tmTextUnitIds);
```
aurambaj: that's already set earlier, can remove

aurambaj:
got it pushed

PR #412, opened 2019-03-05
add git-diff param to push command to enable processing source asset …

xiayewang:
add git-diff param in push command.
if git diff is enabled, only process source asset contains git changes. Otherwise, process all source files.

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
@@ -32,8 +32,6 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
-import org.unix4j.Unix4j;
-import org.unix4j.unix.Grep;
 
 /**
  * @author jaurambault
@@ -186,6 +184,7 @@ public class PushCommand extends Command {
         String gitPath = gitRepository.getDirectory().getPath().replace(".git", "");
         Set<String> gitDiffFiles = new HashSet<>();
 
+        Pattern pattern = Pattern.compile(gitRegex);
         for(FileMatch fileMatch: sourceFileMatches) {
             ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
             List<DiffEntry> diff = git.
@@ -194,18 +193,13 @@ public class PushCommand extends Command {
                     setOutputStream(byteArrayOutputStream).
                     setPathFilter(PathFilter.create(fileMatch.getPath().toString().substring(gitPath.length()))).
                     call();
-            if(Integer.parseInt(Unix4j.
-                    fromString(byteArrayOutputStream.toString()).
-                    grep(Grep.Options.count, Pattern.compile(gitRegex)).
-                    toStringResult()) > 0) {
+            if(pattern.matcher(byteArrayOutputStream.toString()).find()) {
                 gitDiffFiles.add(gitPath + diff.get(0).getNewPath());
             }
-
-
         }
         return sourceFileMatches.
                 stream().
-                filter(fileMatch -> gitDiffFiles.contains(fileMatch.getPath().toString().substring(gitPath.length()))).
+                filter(fileMatch -> gitDiffFiles.contains(fileMatch.getPath().toString())).
                 collect(Collectors.toCollection(ArrayList::new));
     }
 }
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
@@ -14,26 +14,26 @@ import com.box.l10n.mojito.rest.entity.*;
 import com.box.l10n.mojito.rest.entity.Repository;
 import com.google.common.collect.Sets;
 
-import java.io.*;
-import java.util.*;
+import java.io.ByteArrayOutputStream;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-
-import static org.grep4j.core.Grep4j.grep;
-import static org.grep4j.core.Grep4j.regularExpression;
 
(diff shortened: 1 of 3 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
```diff
@@ -87,6 +96,16 @@ public void execute() throws CommandException {
         List<PollableTask> pollableTasks = new ArrayList<>();
 
         ArrayList<FileMatch> sourceFileMatches = commandHelper.getSourceFileMatches(commandDirectories, fileType, sourceLocale, sourcePathFilterRegex);
+
+        // filter out source files which do not contain diff changes if git diff is enabled.
```
aurambaj: use a log statement instead of the comment please.

cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
```diff
@@ -87,6 +96,16 @@ public void execute() throws CommandException {
...
+        if (gitDiff) {
+            sourceFileMatches = getGitChangeFiles(sourceFileMatches);
+            if (sourceFileMatches.isEmpty()) {
+                logger.debug("No new string has been created. No need to create branch and process asset.");
+                return;
```
aurambaj: i'd rather keep function with a single return

cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
```diff
@@ -146,4 +164,28 @@ public void execute() throws CommandException {
...
+     * @param sourceFileMatches
+     * @return
+     * @throws CommandException
+     */
+    private ArrayList<FileMatch> getGitChangeFiles(ArrayList<FileMatch> sourceFileMatches) throws CommandException {
+        GitRepository gitRepository = new GitRepository();
+        gitRepository.init(commandDirectories.getSourceDirectoryPath().toString());
+        Git git = new Git(gitRepository.jgitRepository);
+        String gitPath = gitRepository.getDirectory().getPath().replace(".git", "");
+        Set<String> gitDiffFiles = new HashSet<>();
```
aurambaj: can we have logic to skip some diff, the equivalent of ``` if [ `git diff | grep "^\+" | grep -c -v -E "([+]{3})|(^\+\"POT-Creation-Date:)"` != '0' ]; then``` for the POT files

cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
```diff
@@ -146,4 +164,28 @@ public void execute() throws CommandException {
...
+        try {
+            List<DiffEntry> diffEntries = git.diff().setOldTree(gitRepository.getHeadTree()).call();
+            gitDiffFiles.addAll(diffEntries.stream().map(DiffEntry::getNewPath).collect(Collectors.toSet()));
+        } catch (IOException | GitAPIException e) {
+            logger.debug(e.getMessage());
```
aurambaj: when is this happening? is it fine to silently fail?

cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
```diff
@@ -146,4 +164,28 @@ public void execute() throws CommandException {
...
+        }
+        return sourceFileMatches.
+                stream().
+                filter(fileMatch -> fileMatch.getPath().toString().startsWith(gitPath) ? gitDiffFiles.contains(fileMatch.getPath().toString().substring(gitPath.length())) : false).
+                collect(Collectors.toCollection(ArrayList::new));
```
aurambaj: shorter version: `collect(Collectors.toList())`

cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
```diff
@@ -146,4 +164,28 @@ public void execute() throws CommandException {
...
+        Set<String> gitDiffFiles = new HashSet<>();
+        try {
+            List<DiffEntry> diffEntries = git.diff().setOldTree(gitRepository.getHeadTree()).call();
+            gitDiffFiles.addAll(diffEntries.stream().map(DiffEntry::getNewPath).collect(Collectors.toSet()));
+        } catch (IOException | GitAPIException e) {
+            logger.debug(e.getMessage());
+        }
+        return sourceFileMatches.
+                stream().
+                filter(fileMatch -> fileMatch.getPath().toString().startsWith(gitPath) ? gitDiffFiles.contains(fileMatch.getPath().toString().substring(gitPath.length())) : false).
```
aurambaj: when is the sourceFileMatch not in the gitPath?

cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
xiayewang: going to add

cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
xiayewang: moving to log statement

cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
xiayewang: remove line 103 - 105 to keep single return

cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
xiayewang: sourceFileMatches is initiated as ArrayList in execute(). Abstract List is not compatible with ArrayList.

cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
xiayewang: you are right. soruceFile should always under gitPath. We can remove the condition check here.

xiayewang: the initial purpose is to prevent indexOutOfBoundary exception in case of some weird cases.

xiayewang:
Move to https://github.com/box/mojito/pull/415

PR #414, opened 2019-03-05
Add screenshots to workbench

chidinmae:


webapp/src/main/resources/public/js/components/workbench/GitBlameScreenshotViewerModal.js
```diff
@@ -0,0 +1,113 @@
+import React from "react";
+import PropTypes from 'prop-types';
+import {Col, Glyphicon, Image, Modal} from "react-bootstrap";
+import keycode from "keycode";
+import ClassNames from "classnames";
+import {injectIntl} from "react-intl";
+
+class GitBlameScreenshotViewerModal extends React.Component {
```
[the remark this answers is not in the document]
chidinmae: the store is different, but will delete the modal

webapp/src/main/resources/public/js/stores/workbench/GitBlameScreenshotViewerStore.js
```diff
@@ -0,0 +1,54 @@
+import alt from "../../alt";
+import GitBlameScreenshotViewerActions from "../../actions/workbench/GitBlameScreenshotViewerActions";
+
+class GitBlameScreenshotViewerStore {
```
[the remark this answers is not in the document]
chidinmae: it doesnt make sense to extend the BranchesScreenshotViewerStore, they dont share code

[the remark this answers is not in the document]
aurambaj: I've check the code more and yes it is almost the same thing. The only difference at the moment is the screenshot list that is passed instead of the `branchStatisticId`.

The base store can be similar to this one and `BranchesScreenshotViewerStore` changed with the open function be something like 
```javascript
open(branchStatisticId) {
    let branchStatisticScreenshots = BranchesStore.getBranchStatisticById(branchStatisticId).branch.screenshots
    super.open(branchStatisticScreenshots);
}
```

webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
```diff
@@ -43,6 +44,28 @@ let GitBlameInfoModal = React.createClass({
...
+     */
+    displayScreenshotLink(label, data, branchScreenshots) {
+        if (data == null || !branchScreenshots || !branchScreenshots.length) {
+            return this.displayInfo(label,null)
+        }
+        return (
+            <div className={"row git-blame"}>
+                <label className={"col-sm-3 git-blame-label"}>{label}</label>
+                <Link
+                    onClick={() => GitBlameScreenshotViewerActions.open(branchScreenshots)}
```
aurambaj: would you mind having a prop for the "onClick" callback? 

Trying to keep the UI component "dumb" and let the top component do all the wiring

webapp/src/main/resources/public/js/stores/workbench/GitBlameScreenshotViewerStore.js
```diff
@@ -0,0 +1,20 @@
...
+class GitBlameScreenshotViewerStore extends ScreenshotViewerStore {
+
+    constructor() {
+        super();
+        this.bindActions(GitBlameScreenshotViewerActions);
+    }
+
+    open(branchScreenshots) {
+        this.branchStatisticScreenshots = branchScreenshots;
+        this.loadScreenshot(1);
```
aurambaj: same here

webapp/src/main/resources/public/js/stores/branches/BranchesScreenshotViewerStore.js
```diff
@@ -1,57 +1,22 @@
...
-        this.number = 0;
-        this.total = 0;
-        this.branchStatisticId = null;
-        this.textUnits = [];
-    }
-
     open(branchStatisticId) {
-        this.loadScreenshotByIndex(branchStatisticId, 1);
+        this.branchStatisticScreenshots = BranchesStore.getBranchStatisticById(branchStatisticId).branch.screenshots;
+        this.loadScreenshot(1);
```
aurambaj: move those 2 line in parent `open()`
```
this.loadScreenshot(1);
this.show = true;
```

webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
```diff
@@ -43,6 +44,28 @@ let GitBlameInfoModal = React.createClass({
...
     },
 
+    /**
+     * @param label                 The label for the data to display
+     * @param data                  The link text
+     * @param branchScreenshots     The screenshots for this branch
+     * @returns {*}                 The row of label:link to display in the modal
+     */
+    displayScreenshotLink(label, data, branchScreenshots) {
+        if (data == null || !branchScreenshots || !branchScreenshots.length) {
```
aurambaj: looks like the label is always `"Screenshot"` and data is `"View"` all the time. The `branchScreenshots` can also be fetch in the function.

We usually internationalize the UI strings (though this page might be missing some)

```
displayScreenshotLink() {
   const label = this.props.intl.formatMessage({ id:  "...screenshot"});
...
onClick={() => this.props.onViewScreenshotClick(branchScreenshots)}
                    className="col-sm-9 git-blame-screenshot clickable">
                   <FormattedMessage id="....view" />
}
```

webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
```diff
@@ -43,6 +44,28 @@ let GitBlameInfoModal = React.createClass({
         );
     },
 
+    /**
+     * @returns {*} The row of label:link to display in the modal
+     */
+    displayScreenshotLink() {
+        const label = this.props.intl.formatMessage({ id: "textUnit.gitBlameModal.screenshots"});
+        const branchScreenshots = this.getBranchScreenshots();
```
aurambaj: you can make `getBranchScreenshots()` return non `null` result so that you don't have to test it

webapp/src/main/resources/public/js/stores/ScreenshotViewerStore.js
```diff
@@ -0,0 +1,48 @@
...
+    constructor() {
+        this.setDefaultState();
+    }
+
+    setDefaultState() {
+        this.show = false;
+        this.src = null;
+        this.number = 0;
+        this.total = 0;
+        this.branchStatisticScreenshots = null;
```
aurambaj: safer to initialize that to `[]` to prevent pointer issue in `loadScreenshot`

webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
```diff
@@ -165,6 +189,14 @@ let GitBlameInfoModal = React.createClass({
         }
     },
 
+    getBranchScreenshots() {
+        try {
+            return this.props.gitBlameWithUsage.branch.screenshots;
+        } catch(e) {
+            return null;
```
aurambaj: from previous comment return `[]`

webapp/src/main/resources/public/js/stores/ScreenshotViewerStore.js
```diff
@@ -0,0 +1,48 @@
...
+        this.textUnits = [];
+    }
+
+    open() {
```
aurambaj: Probably make more sense to have `open(branchStatisticScreenshots)` so it makes it clearer what is required for the parent class to work and then you can remove one of the override

webapp/src/main/resources/public/js/stores/workbench/GitBlameScreenshotViewerStore.js
```diff
@@ -0,0 +1,19 @@
...
+
+class GitBlameScreenshotViewerStore extends ScreenshotViewerStore {
+
+    constructor() {
+        super();
+        this.bindActions(GitBlameScreenshotViewerActions);
+    }
+
+    open(branchScreenshots) {
+        this.branchStatisticScreenshots = branchScreenshots;
```
aurambaj: with previous comment, this can be removed

webapp/src/main/resources/public/js/stores/ScreenshotViewerStore.js
chidinmae: but one open takes an id and the other takes the list of screenshots, wouldn't this be misleading for the one that takes id?

aurambaj: good point the one with Id, should be renamed to `openWithBranchStatisticId ` or something like that

PR #422, opened 2019-03-15
Import dup name

aurambaj:


aurambaj:
yes I will but I may do some other implementation. We shouldn't hard fail the import but that implementation is random. Going to think more about it

PR #430, opened 2019-03-29
Add optional third party TMS links to the text unit info

aurambaj:


webapp/src/main/resources/config/application.properties
```diff
@@ -147,6 +147,11 @@ l10n.PluralFormUpdater=true
 #l10n.link.{repository}.commit.url=http://github.com/${gitHubRepository}
 #l10n.link.{repository}.commit.label=${gitHubRepository}
 
+#l10n.link.{repository}.thirdParty.url=http://sometool.com/?
+#l10n.link.{repository}.thirdParty.urlComponent=queryParamToEncode=${assetPath}
```
magezil: Did you intend to have two equal signs here? (One between `urlComponent` and `queryParamToEncode`, the other between `queryParamToEncode` and `${assetPath}`

webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
```diff
@@ -184,19 +185,53 @@ let GitBlameInfoModal = React.createClass({
...
+        let url = LinkHelper.renderUrl(urlTemplate, urlComponentTemplate, params);
+        let label = LinkHelper.renderTemplate(labelTemplate, params);
+        return <a href={url}>{label}</a>;
+    },
+
+    getThirdPartyLink() {
+        let params = this.getParamsForLinks();
+        let url = LinkHelper.renderUrl(this.getThirdPartyUrlTemplate(), this.getThirdPartyUrlComponentTemplate(), params);
+        let label = LinkHelper.renderTemplate(this.getThirdPartyLabelTemplate(), params);
+        return <a href={url}>{label}</a>;
```
magezil: you can just reuse the logic of `renderLink` to create the link: `return this.renderLink(this.getThirdPartyUrlTemplate(), this.getThirdPartyUrlComponentTemplate(), this.getThirdPartyLabelTemplate(), params)`

webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
```diff
@@ -271,7 +293,7 @@ let GitBlameInfoModal = React.createClass({
      */
     getLocationExtractorPrefix() {
         try {
-            return this.props.appConfig.link.link[this.props.textUnit.getRepositoryName()].location.extractorPrefix;
```
magezil: does this work without the second `link`? 😮

webapp/src/main/resources/config/application.properties
aurambaj: yes, the first is the delimiter of the property file and after that it can be anything. For a query param `=` is used between the key and the value

webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
aurambaj: :D introduced the function for this... Thanks for catching!

webapp/src/main/java/com/box/l10n/mojito/react/ReactAppConfig.java
```diff
@@ -1,11 +1,13 @@
 package com.box.l10n.mojito.react;
 
+import com.fasterxml.jackson.annotation.JsonUnwrapped;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 @Component
 public class ReactAppConfig {
 
+    @JsonUnwrapped
```
aurambaj: @magezil remove a level of "link" with the `@JsonUnwrapped` annotation. I didn't make sense to have two, I think it was just there as a result have having the map in an object

webapp/src/main/resources/public/js/components/workbench/GitBlameInfoModal.js
aurambaj: see comment on `@JsonUnwrapped`

webapp/src/main/java/com/box/l10n/mojito/react/ReactAppConfig.java
magezil: Awesome! I agree that it doesn't make sense to have two.

PR #434, opened 2019-04-05
inital smartling PR with working API calls to authenticate, refresh and pull source strings

chidinmae:


webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    public enum RequestType {
+        AUTHENTICATION,
+        REFRESH,
+        SOURCE_STRINGS,
+        WITH_TOKEN
+    }
+
+    Map<String, Object> getPayloadMap(RequestType requestType) {
```
aurambaj: probably better to have 2 independent functions and not pass the `requestType` in that case

webapp/src/main/java/com/box/l10n/mojito/entity/ThirdPartyTranslations.java
```diff
@@ -0,0 +1,39 @@
+package com.box.l10n.mojito.entity;
+
+import javax.persistence.*;
+
+/**
+ * Entity used to store the mapping between third party translation ids and mojito text units.
```
aurambaj: nitpick: it is mapping text units ("string") not really translations

webapp/src/main/java/com/box/l10n/mojito/entity/ThirdPartyTranslations.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @author chidinmae
+ */
+@Entity
+@Table(name = "third_party_translations")
```
aurambaj: so `third_party_translations` -> `third_party_text_unit` (we don't use 's' in the table names)

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,199 @@
...
+            map.put("refreshToken", refreshToken);
+        }
+        return map;
+    }
+
+    HttpHeaders getHttpHeadersForJson(RequestType requestType) {
+        HttpHeaders headers = new HttpHeaders();
+        headers.setContentType(MediaType.APPLICATION_JSON);
+        if (requestType == RequestType.SOURCE_STRINGS) {
+            headers.set("Authorization", "Bearer " + authToken);
```
aurambaj: why do we set only the auth header for "source string", looks like we should have a function for unauth and one for auth request. So following previous comment, probably better to split in 2 functions.

The logic for unauth is probably gonna be unique and might be inlined. vs auth that will be shared by all API calls

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,199 @@
...
+            authToken = authenticationResponse.getResponse().getData().getAccessToken();
+            refreshToken = authenticationResponse.getResponse().getData().getRefreshToken();
+        }
+        return authenticationResponse;
+    }
+
+    AuthenticationResponse postForAuthentication(HttpEntity<Map<String, Object>> httpEntity, String authApiV2) throws SmartlingClientException {
+        String postForObject = restTemplate.postForObject(
+                authApiV2, httpEntity, String.class);
+        ObjectMapper json = new ObjectMapper();
```
aurambaj: no need to re-instanciate the `ObjectMapper` all the time, should be moved as an attribute. there is shared one in the app that can be autowired as well

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,199 @@
...
+        try {
+            authenticationResponse = json.readValue(postForObject, AuthenticationResponse.class);
+        } catch (IOException e) {
+            e.printStackTrace();
+            throw new SmartlingClientException(e.getMessage());
+        }
+        return authenticationResponse;
+    }
+
+    public AuthenticationResponse refresh() throws SmartlingClientException {
```
aurambaj: this is never called?

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,199 @@
...
+
+    AuthenticationResponse postForAuthentication(HttpEntity<Map<String, Object>> httpEntity, String authApiV2) throws SmartlingClientException {
+        String postForObject = restTemplate.postForObject(
+                authApiV2, httpEntity, String.class);
+        ObjectMapper json = new ObjectMapper();
+        AuthenticationResponse authenticationResponse;
+        try {
+            authenticationResponse = json.readValue(postForObject, AuthenticationResponse.class);
+        } catch (IOException e) {
+            e.printStackTrace();
```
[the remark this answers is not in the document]
chidinmae: thanks! this was from testing

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClientConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConfigurationProperties("l10n.smartling")
+public class SmartlingClientConfiguration {
+
+    String userIdentifier;
+    String userSecret;
+
+    @ConditionalOnProperty("${l10n.smartling.userIdentifier:true} && ${l10n.smartling.userSecret:true}")
```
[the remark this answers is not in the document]
chidinmae: https://stackoverflow.com/a/51257660

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
chidinmae: we set no auth header for auth and refresh because they are not required, I don't think it needs to be split https://api-reference.smartling.com/#tag/Authentication

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
chidinmae: as I said when you requested the PR, the work is incomplete

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
aurambaj: my comment was unclear. I got that it doesn't set the header since it is fetching the information for it. 

what I was trying to say is:
1) if more APIs get added to the client, for each of those, the "if" statement will have to be updated with the new request type and that doesn't scale. It is also hard to test and overall won't be clean
2) there is no point during the auth flow to call a function that has logic specific to the API calls that need the authorization header. 

So there should be a base method that gets the header for JSON (used by the auth flow), then have a method that gets header for requests that needs authentication (get source string and new ones later). The later method calls the method to get header for JSON and then adds the authorization header. 

And this was related to the previous comment. there shouldn't be a single `getPayloadMap` method with a if statement for each request type. Instead there should be a function for each request type.

webapp/src/main/java/com/box/l10n/mojito/smartling/request/AuthenticationData.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public void setAccessToken(String accessToken) {
+        this.accessToken = accessToken;
+    }
+
+    public void setTokenType(String tokenType) {
+        this.tokenType = tokenType;
+    }
+
+    public AuthenticationData() {}
```
aurambaj: no need for the empty constructor. same in other classes

xiayewang:
Looks good to me overall. Nitpicking suggestion, looks like there are common logics shared between AuthenticationResponse and SourceStringsResponse. Can we create BaseResponse to extract the common logic like formatError and response code check?

webapp/src/main/java/com/box/l10n/mojito/entity/ThirdPartyTextUnit.java
```diff
@@ -0,0 +1,48 @@
...
+       indexes = {@Index(name = "UK__TP_TEXT_UNIT", columnList = "third_party_text_unit_id", unique = true)})
+public class ThirdPartyTextUnit extends AuditableEntity {
+
+    @Column(name = "third_party_text_unit_id")
+    private String thirdPartyTextUnitId;
+
+    @Column(name = "mapping_key")
+    private String mappingKey;
+
+    @JoinColumn(name = "tm_text_unit_id", foreignKey = @ForeignKey(name = "FK__THIRD_PARTY_TRANSLATION_TEXT_UNIT__TM_TEXT_UNIT__ID"))
```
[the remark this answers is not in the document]
chidinmae: ok could you add that naming convention somewhere I can reference?

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,206 @@
...
+            String msg = formatAuthenticationErrorMessage(authenticationResponse);
+            throw new SmartlingClientException(msg);
+        } else {
+            Instant now = Instant.now();
+            lastAuthentication = now;
+            lastRefresh = now;
+            nextAuthentication = now.plus(Duration.ofHours(authenticationDurationHours));
+            nextRefresh = now.plus(Duration.ofMinutes(refreshDurationMinutes));
+            authToken = authenticationResponse.getResponse().getData().getAccessToken();
+            logger.info(authToken);
```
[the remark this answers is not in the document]
chidinmae: yes thanks left over from testing now that is voided

webapp/src/main/java/com/box/l10n/mojito/entity/ThirdPartyTextUnit.java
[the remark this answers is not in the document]
aurambaj: that's knowledge we share during code reviews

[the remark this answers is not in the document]
aurambaj: also since this PR doesn't contain any "service" logic, it is better to have only the Smartling client code here and move this entity in the PR for the service where it will be used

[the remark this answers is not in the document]
chidinmae: sure thing, will remove

chidinmae:
> Looks good to me overall. Nitpicking suggestion, looks like there are common logics shared between AuthenticationResponse and SourceStringsResponse. Can we create BaseResponse to extract the common logic like formatError and response code check?

Thanks, I've added a BaseResponse

chidinmae:
> Can you please add a test for the Smartling client? something similar to the SlackClientTests. Even if it can't be run during the standard test suite, we'll have it to run one off tests.
> 
> We may want to have a mechanism to invalidate the tokens. If the server return an authentication error it looks like it will keep failing by using the same token until it is expired.

ok to add test in next PR?

aurambaj:
> > Can you please add a test for the Smartling client? something similar to the SlackClientTests. Even if it can't be run during the standard test suite, we'll have it to run one off tests.
> > We may want to have a mechanism to invalidate the tokens. If the server return an authentication error it looks like it will keep failing by using the same token until it is expired.
> 
> ok to add test in next PR?

Please add the tests at the time as the code, so in this PR.

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+        return authenticationResponse;
+    }
+
+    public void refresh() throws SmartlingClientException {
+        Instant now = Instant.now();
+        Instant nextCalculatedRefresh = now.plus(Duration.ofMinutes(refreshDurationMinutes));
+        if (nextRefresh.isAfter(now) && nextAuthentication.isAfter(now)) {
+            return;
+        } else if (nextCalculatedRefresh.isAfter(nextAuthentication)) {
```
aurambaj: why is do we need `nextCalculatedRefresh`? Shouldn't `nextAuthentication` be enough to know when to do authentication again?

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,188 @@
...
+            authenticate();
+            return;
+        }
+        Map<String, Object> payload = getPayloadMapRefresh();
+        AuthenticationResponse authenticationResponse;
+        try {
+            authenticationResponse = postForAuthentication(payload, AUTH_API_V2_REFRESH);
+        } catch (SmartlingClientException e) {
+            throw new SmartlingClientException(e.getMessage());
```
aurambaj: can just re-throw:

```
} catch (SmartlingClientException e) {
     throw e;
}
```

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        if (authenticationResponse.isAuthenticationErrorResponse()) {
+            authenticate();
```
aurambaj: what's the different between those 3 error cases? why do we sometime re-authenticate and sometimes throw an exception?

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,188 @@
...
+
+    RestTemplate restTemplate = new RestTemplate();
+    ObjectMapper objectMapper = new ObjectMapper();
+
+    String userIdentifier;
+    String userSecret;
+    String authToken = null;
+    String refreshToken = null;
+    Instant lastAuthentication = null;
+    Instant lastRefresh = null;
```
aurambaj: same, it looks like this variable is set but not used anywhere

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,188 @@
...
+    static final String API_BIND_CONTEXT_TO_HASHCODE = DEFAULT_BASE_HOST + "context-api/v2/projects/{projecId}/bindings";
+
+    RestTemplate restTemplate = new RestTemplate();
+    ObjectMapper objectMapper = new ObjectMapper();
+
+    String userIdentifier;
+    String userSecret;
+    String authToken = null;
+    String refreshToken = null;
+    Instant lastAuthentication = null;
```
aurambaj: It looks like this variable is set but not used anywhere

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,188 @@
...
+        return new HttpEntity<>(payload, getHttpHeadersForJsonWithAuthorization());
+    }
+
+    public void authenticate() throws SmartlingClientException {
+        Map<String, Object> payload = getPayloadMapAuthentication();
+        AuthenticationResponse authenticationResponse;
+        try {
+            authenticationResponse = postForAuthentication(payload, AUTH_API_V2_AUTHENTICATE);
+        } catch (SmartlingClientException e) {
+            throw new SmartlingClientException(e.getMessage());
```
aurambaj: here, you should just re-throw the exception, not re-create one.

```
catch (SmartlingClientException e) {
            throw e;
```

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,188 @@
...
+    AuthenticationResponse postForAuthentication(Map<String, Object> payload, String authApiV2) throws SmartlingClientException {
+        HttpEntity<Map<String, Object>> httpEntity = getHttpEntityForPayload(payload);
+        String postForObject = restTemplate.postForObject(
+                authApiV2, httpEntity, String.class);
+        AuthenticationResponse authenticationResponse;
+        try {
+            authenticationResponse = objectMapper.readValue(postForObject, AuthenticationResponse.class);
+        } catch (IOException e) {
+            logger.error(e.getMessage());
+            throw new SmartlingClientException(e.getMessage());
```
aurambaj: In general, it is is better to have an explicit message about what is failing and log/chain the exception.
```
} catch (IOException e) {
            String msg = "Could not deserialize the authentication response";
            logger.error(msg, e); 
            throw new SmartlingClientException(msg, e);
}
```

but in that case you should be able to let the restTemplate do the Jackson deserialization instead of getting a String and then converting with the object mapper.

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,188 @@
...
+        Map<String, Object> payload = getPayloadMapPullSourceString();
+
+        HttpEntity<Map<String, Object>> httpEntity = getHttpEntityForPayloadWithAuthorization(payload);
+
+        Map<String, String> pathVariables = new HashMap<>();
+        pathVariables.put("projectId", projectId);
+        pathVariables.put("fileUri", file);
+        pathVariables.put("offset", offset.toString());
+        UriTemplate template = new UriTemplate(API_PULL_SOURCE_STRINGS);
+        ResponseEntity<String> exchange = restTemplate.exchange(
```
aurambaj: same as before, the restTemplate do the Jackson deserialization instead of getting a String and then converting with the object mapper.

webapp/src/test/java/com/box/l10n/mojito/smartling/SmartlingClientTest.java
```diff
@@ -0,0 +1,44 @@
...
+    @Value("${test.l10n.smartling.offset:0}")
+    Integer offset;
+
+    @Test
+    public void testClient() throws SmartlingClientException {
+        Assume.assumeNotNull(userIdentifier, userSecret);
+        SmartlingClient smartlingClient = new SmartlingClient(userIdentifier, userSecret);
+
+        smartlingClient.authenticate();
+        smartlingClient.refresh();
```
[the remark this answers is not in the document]
chidinmae: Its not needed, I can remove them from the test.

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
chidinmae: restTemplate doesn't do well with nested classes which is why I convert from string to the class with jackson

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
chidinmae: this is to pre-emptively authenticate if this refresh is close to the next time we need to authenticate

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
aurambaj: hm, I'm surprised since it should be using Jackson too and we've been using it without issue. Can you show the exact failure with nested classes? it should probably be fixable.

chidinmae: I don't have the logs anymore so I'd have to recreate the conditions, can do so later if yo'd like

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
chidinmae: 1. authentication error means we should authenticate 
2. any other response code that is not success and that is not related to authentication should be surfaced 
3. not authentication error and not not success code means it is a success code and we should set the refresh token.

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
chidinmae: but will also add a check for if nextAuthentication is after now

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    <T> HttpEntity<T> getHttpEntityForPayloadWithAuthorization(T payload) {
+        return new HttpEntity<>(payload, getHttpHeadersForJsonWithAuthorization());
+    }
+
+    public void authenticate() throws SmartlingClientException {
+        Map<String, Object> payload = getPayloadMapAuthentication();
+        AuthenticationResponse authenticationResponse;
+        try {
```
aurambaj: `try catch` is not needed since the method `throws SmartlingClientException`

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,165 @@
...
+    RestTemplate restTemplate = new RestTemplate();
+
+    String userIdentifier;
+    String userSecret;
+    String authToken = null;
+    String refreshToken = null;
+    Instant nextAuthentication = Instant.now();
+    Instant nextRefresh = Instant.now();
+    // Refresh tokens are valid for up to 1 hour
+    Integer refreshDurationMinutes = 50;
```
aurambaj: should this be constant? doesn't seem to be changed

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,165 @@
...
+    // Authentication tokens can be refreshed for up to 24 hours
+    Integer authenticationDurationHours = 23;
```
aurambaj: same, should this be constant? doesn't seem to be changed

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -0,0 +1,165 @@
...
+    public void refresh() throws SmartlingClientException {
+        Instant now = Instant.now();
+        Instant nextCalculatedRefresh = now.plus(Duration.ofMinutes(refreshDurationMinutes));
+        if (now.isAfter(nextAuthentication) || nextCalculatedRefresh.isAfter(nextAuthentication)) {
+            authenticate();
+        } else if (!(nextRefresh.isAfter(now) && nextAuthentication.isAfter(now))) {
+            Map<String, Object> payload = getPayloadMapRefresh();
+            AuthenticationResponse authenticationResponse;
+            try {
+                authenticationResponse = postForAuthentication(payload, AUTH_API_V2_REFRESH);
```
aurambaj: same, try catch is not needed since the method `throws SmartlingClientException`

aurambaj:
looks great! 

Can you rebase (sounds like there is conflict) and squash into 1 commit? Then I'll merge

PR #438, opened 2019-04-16
Implemented partially imported drop

jeeyi:
Sometimes some languages are delayed and the drop is partially imported. Current behavior is that the drop is considered to have `Imported` status as long as there is at least one xliff that was imported successfully. It's hard to distinguish between the fully completed drop and the partially completed drop. 

Partially completed drop should appear under `In Progress` filter with `Partially Imported` status. Also the detail should show which language was imported.

Screen Shot 2019-04-16 at 10 52 10 AM

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/entity/Drop.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/entity/Drop.java
@@ -96,6 +96,10 @@ public class Drop extends AuditableEntity {
     @JsonView(View.DropSummary.class)
     protected Boolean importFailed;
 
+    @JsonView(View.DropSummary.class)
+    @Column(name = "partiallyImported")
+    private Boolean partiallyImported = false;
+
     public User getCreatedByUser() {
         return createdByUser;
     }
@@ -191,6 +195,14 @@ public class Drop extends AuditableEntity {
 
     public void setImportFailed(Boolean importFailed) {
         this.importFailed = importFailed;
+    }
+
+    public Boolean getPartiallyImported() {
+        return partiallyImported;
+    }
+
+    public void setPartiallyImported(Boolean partiallyImported) {
+        this.partiallyImported = partiallyImported;
     }
 
 }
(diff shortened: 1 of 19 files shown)


jeeyi:
Do not merge yet, one more change is coming - a button to force the `Partially Imported` status to `Imported` status.

webapp/src/main/java/com/box/l10n/mojito/entity/Drop.java
```diff
@@ -96,6 +96,10 @@
     @JsonView(View.DropSummary.class)
     protected Boolean importFailed;
 
+    @JsonView(View.DropSummary.class)
+    @Column(name = "partiallyImported")
```
aurambaj: `partially_imported`?

webapp/src/main/resources/db/migration/V44__Add_imported.sql
```diff
@@ -0,0 +1,2 @@
+alter table `drop` add column partiallyImported bit not null default false;
```
aurambaj: same as in the entity `partially_imported`

webapp/src/main/java/com/box/l10n/mojito/entity/Drop.java
jeeyi: Done

webapp/src/main/resources/db/migration/V44__Add_imported.sql
jeeyi: Done

jeeyi:
This is the button to mark partially imported drop to completed. This only shows up for the partially imported drops.

Screen Shot 2019-04-16 at 4 04 01 PM

webapp/src/main/resources/public/js/components/drops/Drops.js
```diff
@@ -173,6 +173,15 @@ let Drops = React.createClass({
         this.delayGetNewRequests(500);
     },
 
+    /**
+     * Handle complete onclick event
+     */
+    onClickComplete(dropId) {
+        DropActions.completeRequest(dropId);
+
```
aurambaj: extra line?

jeeyi: Removed :)

PR #440, opened 2019-04-24
Don't escape "annotation" in Android

aurambaj:


webapp/src/main/java/com/box/l10n/mojito/okapi/filters/AndroidXMLEncoder.java
```diff
@@ -2,41 +2,56 @@
...
+@Configurable
 public class AndroidXMLEncoder extends net.sf.okapi.common.encoder.XMLEncoder {
 
+    /**
+     * logger
+     */
+    static Logger logger = LoggerFactory.getLogger(AndroidXMLEncoder.class);
+
     // trying to match variables between html tags, for example, <b>%d</b>, <i>%1$s</i>, <u>%2$s</u>
-    private static final Pattern ANDROID_VARIABLE_WITHIN_HTML = Pattern.compile("(&lt;[b|i|u]&gt;)((.*?)%(([-0+ #]?)[-0+ #]?)((\\d\\$)?)(([\\d\\*]*)(\\.[\\d\\*]*)?)[dioxXucsfeEgGpn](.*?))+(&lt;/[b|i|u]&gt;)");
```
aurambaj: @jeeyi not sure to understand that requirement. Do you remember what it was for?

webapp/src/main/java/com/box/l10n/mojito/okapi/filters/AndroidXMLEncoder.java
```diff
@@ -2,41 +2,56 @@
...
      * New escaping, should be come default but keep it as an option for backward compatibility. Can invert the option
      * later to old escaping or just remove it
      */
     boolean newEscaping = false;
 
+    @Autowired
+    UnescapeUtils unescapeUtils;
+
     public AndroidXMLEncoder(boolean newEscaping) {
         this.newEscaping = newEscaping;
```
aurambaj: @jeeyi If you get a chance to test the new escaping, I think it will be nice to turn it on by default.

to use it right now: `mojito push -r android -fo newEscaping=true`

jeeyi: Let me test it and get back to you

jeeyi: @aurambaj With the new escaping, some strings need re-translation because they do not match anymore. I am okay to make the new escaping the default as long as there is a way to use the old one. I can update our push command for Android.

webapp/src/main/java/com/box/l10n/mojito/okapi/filters/AndroidXMLEncoder.java
jeeyi: I can't remember it anymore. As long as all unit tests work, we can remove it.

aurambaj: `testLocalizeAndroidStringsWithEscapedHTMLAndCDATA` seems to test that case, just not sure to understand in that case why it should behave like. I won't touch it

webapp/src/main/java/com/box/l10n/mojito/okapi/filters/AndroidXMLEncoder.java
aurambaj: Yes that make sense since the behavior is slightly different. I believe that accepting the old string should be fine since it should be backward compatible in term of Android app but I haven't done extensive testing. 

I'm gonna swap the option for `oldEscaping` then. Usage will be `-fo oldEscaping=true` and if not set it will use the new escaping by default

jeeyi: Got it!

jeeyi:
@aurambaj please fix the test failures.

webapp/src/main/java/com/box/l10n/mojito/okapi/filters/AndroidXMLEncoder.java
aurambaj: Also I think most of the change should be backward compatible (only changing how quotes are handled). The things that might be different are line feed. They used to show escaped in the translation and now, it is real line feed. Are those the main differences you see?

I'm thinking a good way could be to just push+import for the current files with the new escaping. It will re-import the small diff and then you don't have to use `-fo oldEscaping=true` at all.

jeeyi:
Thank you!

PR #444, opened 2019-05-03
screenshots: adding ThirdPartTextUnit entity

chidinmae:


chidinmae:
> When adding an entity it is also need to add a migration script with it: http://www.mojito.global/docs/guides/open-source-contributors/#to-generate-the-final-migration-script-before-committing-code
> 
> Are we going to have another table to keep track of the screenshots that needs to be uploaded? We can put both table in one commit to have a single migration script.

would prefer to make that a separate diff once more of the screenshot methods are in place, API calls, etc

PR #450, opened 2019-05-16
 screenshots: update SmartlingClient and test

chidinmae:
Access tokens expire every 5-minutes, but you can get a new token by calling the Refresh Authentication API, using the refresh token returned by a successful Authentication. Refresh tokens are valid for 1 hour. You can refresh your Authentication for up to 24-hours before you need to log in again with an Authenticate request.

https://help.smartling.com/hc/en-us/articles/360007789914

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
@@ -1,16 +1,14 @@
 package com.box.l10n.mojito.smartling;
 
-import com.box.l10n.mojito.smartling.response.AuthenticationResponse;
 import com.box.l10n.mojito.smartling.response.SourceStringsResponse;
 import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


aurambaj:
Also I'm wondering if it is a regular OAuth2 flow, can [OAuth2RestTemplate](https://docs.spring.io/spring-security/oauth/apidocs/org/springframework/security/oauth2/client/OAuth2RestTemplate.html) be reused? Similar to this: https://www.baeldung.com/spring-security-oauth2-authentication-with-reddit

chidinmae:
> Also I'm wondering if it is a regular OAuth2 flow, can [OAuth2RestTemplate](https://docs.spring.io/spring-security/oauth/apidocs/org/springframework/security/oauth2/client/OAuth2RestTemplate.html) be reused? Similar to this: https://www.baeldung.com/spring-security-oauth2-authentication-with-reddit

according to their docs it is Oauth2, will check this out and update

PR #456, opened 2019-05-23
screenshots: add DTO classes for matching and searching

chidinmae:


webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitMatchingService.java
```diff
@@ -0,0 +1,173 @@
...
+    SmartlingClient smartlingClient;
+
+    @Autowired
+    TMTextUnitRepository tmTextUnitRepository;
+
+    void getSmartlingThirdPartyTextUnits(String projectId) {
+        FilesResponse files = smartlingClient.getFiles(projectId);
+        for (File file : files.getResponse().getData().getItems()) {
+            String fileUri = file.getFileUri();
+            if (fileUri.matches(".*\\d+_(singular|plural)_source\\.xml")) {
```
aurambaj: Can you please move the regex to a pattern that is created only once and replace this with something like `pattern.match(fileUri).matches()`

webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitMatchingService.java
```diff
@@ -0,0 +1,173 @@
...
+                Integer offset = 0;
```
aurambaj: following logic should go in a function with the logic to process 1 file

webapp/src/test/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitMatchingServiceTest.java
```diff
@@ -0,0 +1,144 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.annotation.Transactional;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.mockito.Mockito.*;
+
+public class ThirdPartyTextUnitMatchingServiceTest extends ServiceTestBase {
+    @Autowired
+    private ThirdPartyTextUnitMatchingService thirdPartyTextUnitMatchingService = new ThirdPartyTextUnitMatchingService();
```
aurambaj: can you remove "private" to be consistent with the rest?

Since `@Autowired` remove ` = new ThirdPartyTextUnitMatchingService(`

webapp/src/test/java/com/box/l10n/mojito/service/smartling/SmartlingTestData.java
```diff
@@ -0,0 +1,98 @@
+package com.box.l10n.mojito.service.smartling;
+
+import com.box.l10n.mojito.smartling.request.*;
+import com.box.l10n.mojito.smartling.response.FilesResponse;
+import com.box.l10n.mojito.smartling.response.SourceStringsResponse;
+
+import java.util.Arrays;
+import java.util.Collections;
+
+class SmartlingTestData {
```
aurambaj: this seems to be weird to have nested classes like this. Why not just have regular classes?

webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitMatchingService.java
```diff
@@ -0,0 +1,173 @@
...
+        AtomicInteger partialMatch = new AtomicInteger(0);
+        AtomicInteger newThirdPartyTextUnit = new AtomicInteger(0);
+        AtomicInteger newTPTWithoutMatchingMojitoTU = new AtomicInteger(0);
+        AtomicInteger missingVariant = new AtomicInteger(0);
+        String delimiter = "#@#";
+        // partition the list of StringData into groups of 500
+        List<List<StringData>> batchedStringData = Lists.partition(stringDataList, 500);
+
+        batchedStringData.forEach(
+                stringDataBatch -> {
```
aurambaj: I just had a quick look and I need to do more digging... but here is my first feedback to start the discussion

can we split in 2 steps: "mapping" then "saving"? The batching logic here useful only when saving. Having all done at once is.

Also can you introduce functions to make the code clearer. At first glance, it uses "streams" but it is pretty much imperative loops so in that case it should be just standard "for loops". Or it could be rewritten fully in a more functional style but that'd probably start with `getSmartlingThirdPartyTextUnits` as well

webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitMatchingService.java
```diff
@@ -0,0 +1,173 @@
...
+                                            stringdebug -> {
+                                                String hashcode = stringdebug.getHashcode();
+                                                String mappingKey = stringdebug.getStringVariant();
+
+                                                if (mappingKey != null) {
+                                                    String[] mappingKeySplit = mappingKey.split(delimiter);
+                                                    String assetPath = mappingKeySplit[0];
+                                                    String name = mappingKeySplit[1];
+
+                                                    // Check to see if we already have a match for this hashcode and mapping key, if so we can skip it
```
aurambaj: I am worried about performances here as it is making 1 query per text unit. In other places we batch fetch the info to perform the mapping. I'd have to look at it in more detail, just a first feedback.

webapp/src/main/java/com/box/l10n/mojito/entity/ThirdPartyTextUnit.java
```diff
@@ -47,4 +47,32 @@ public void setTmTextUnit(TMTextUnit tmTextUnit) {
         this.tmTextUnit = tmTextUnit;
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof ThirdPartyTextUnit)) return false;
```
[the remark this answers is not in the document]
chidinmae: Yes it is needed to dedupe the insert list to prevent a SQL error

webapp/src/test/java/com/box/l10n/mojito/service/smartling/SmartlingTestData.java
chidinmae: This is just test data and it didn't seem worth it to make a separate file for each test case, similar to TMTestData (current class for test data)

webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitMatchingService.java
chidinmae: will await your further review

chidinmae: I have now refactored the service

webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitDTONativeObjectMapper.java
```diff
@@ -0,0 +1,19 @@
+package com.box.l10n.mojito.service.smartling;
+
+import com.github.pnowy.nc.core.CriteriaResult;
+import com.github.pnowy.nc.core.mappers.NativeObjectMapper;
+
+public class ThirdPartyTextUnitDTONativeObjectMapper implements NativeObjectMapper<ThirdPartyTextUnitDTO> {
```
aurambaj: we're gonna use Native SQL queries?

webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitDTO.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    public void setTmTextUnitName(String tmTextUnitName) {
+        this.tmTextUnitName = tmTextUnitName;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof ThirdPartyTextUnitDTO)) return false;
```
aurambaj: I think I mentioned that in another review. It is hard to really have meaningful "equals" implementation for those kind of entities: should it be based on the "id", the whole set of fields, etc. Since it is not obvious, we haven't been implementing them in the project. Instead we'd use an explicit comparator next to the code that needs it.

webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitDTONativeObjectMapper.java
chidinmae: similar set up to TextUnitDTONativeObjectMapper, uses NativeCriteria

webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitDTO.java
chidinmae: This is for adding to a Set<ThirdPartyTextUnitDTO>. There is a meaningful equals here where each ThirdPartyTextUnit has unique hashcode, mapping key and text unit it matches to.

webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitDTONativeObjectMapper.java
aurambaj: not sure why NativeCriteria is needed here? right now it is only used for searching text units which require complicated queries. 

For the rest we use standard queries, even when using DTOs (`AssetMappingDTO`, `TextUnitIdMd5DTO`, etc)

chidinmae: Got it, I will remove this file since NativeCriteria is overkill

webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitDTO.java
```diff
@@ -0,0 +1,100 @@
+package com.box.l10n.mojito.service.smartling;
+
+import com.box.l10n.mojito.entity.ThirdPartyTextUnit;
+
+public class ThirdPartyTextUnitDTO {
+    Long id;
```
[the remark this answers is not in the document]
chidinmae: id is passed from the database, it is unique and immutable in the db

webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitDTO.java
```diff
@@ -0,0 +1,100 @@
...
+    public String getTmTextUnitName() {
+        return tmTextUnitName;
+    }
+
+    public void setTmTextUnitName(String tmTextUnitName) {
+        this.tmTextUnitName = tmTextUnitName;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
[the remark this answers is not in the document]
chidinmae: should not be the case as id is passed from db, ids are unique in db

webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitDTO.java
```diff
@@ -0,0 +1,100 @@
...
+        if (thirdPartyTextUnitId != null ? !thirdPartyTextUnitId.equals(thirdPartyTextUnitDTO.thirdPartyTextUnitId) : thirdPartyTextUnitDTO.thirdPartyTextUnitId != null) return false;
+        if (repositoryName != null ? !repositoryName.equals(thirdPartyTextUnitDTO.repositoryName) : thirdPartyTextUnitDTO.repositoryName != null) return false;
+        if (assetPath != null ? !assetPath.equals(thirdPartyTextUnitDTO.assetPath) : thirdPartyTextUnitDTO.assetPath != null) return false;
+        if (tmTextUnitName != null ? !tmTextUnitName.equals(thirdPartyTextUnitDTO.tmTextUnitName) : thirdPartyTextUnitDTO.tmTextUnitName != null) return false;
+
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
```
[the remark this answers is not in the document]
chidinmae: its used for the equals comparison https://www.geeksforgeeks.org/equals-hashcode-methods-java/

webapp/src/main/java/com/box/l10n/mojito/service/smartling/ThirdPartyTextUnitDTO.java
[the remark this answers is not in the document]
xiayewangatpin: make `id` final and private if possible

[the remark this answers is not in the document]
chidinmae: makes sense!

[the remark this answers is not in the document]
chidinmae: I pushed an update

PR #465, opened 2019-06-13
Added import-fetched parameter for drop import

### LINKED ISSUE #462: Import drop non-interactive
In order to automate usage of mojito drop import I need to be able to just import the last drop if it exists, without any interaction. Something like -y or --force would be my idea.

Willing to take on this, any suggestion/guidance?

dragosv:
This fixes #462

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/DropImportCommand.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/DropImportCommand.java
@@ -158,13 +158,18 @@ public class DropImportCommand extends Command {
      * of available {@link Drop}s
      */
     private List<Long> getSelectedDropIds(Map<Long, Drop> numberedAvailableDrops) throws CommandException {
+        List<Long> selectedDropIds;
+
         if (importFetchedParam) {
-            return numberedAvailableDrops.entrySet().stream()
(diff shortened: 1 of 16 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/DropImportCommand.java
```diff
@@ -147,15 +157,24 @@ private Boolean getImportedFilter() {
      * @throws CommandException if the input doesn't match a number from the map
      * of available {@link Drop}s
      */
-    private Long getSelectedDropId(Map<Long, Drop> numberedAvailableDrops) throws CommandException {
+    private List<Long> getSelectedDropIds(Map<Long, Drop> numberedAvailableDrops) throws CommandException {
+        if (importFetchedParam) {
+            return numberedAvailableDrops.entrySet().stream()
+                    .filter(x -> !Boolean.TRUE.equals(x.getValue().getCanceled()) && x.getValue().getLastImportedDate() == null)
```
aurambaj: I don't remember so just asking. if not using `-all` the already imported won't be shown right? And you may want to re-import the drop. So, s hould `x.getValue().getLastImportedDate() == null` be removed?

dragosv: Make sense, especially that the all parameter exists.

PR #470, opened 2019-07-10
Json comments are not processed

jeeyi:
@aurambaj I noticed that comment like this in JSON is not being processed as expected:
```
{
  // comment
  "hello": "Hello!"
}
```
How should we handle when we have both comment `// comment` and `"note": "comment"`?


jeeyi:
This doesn't work. Let me fix it again.

PR #473, opened 2019-07-30
Json with comments

jeeyi:
@aurambaj Let me know if you have any comments/concerns.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/okapi/filters/JSONFilter.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/okapi/filters/JSONFilter.java
@@ -39,6 +39,7 @@ public class JSONFilter extends net.sf.okapi.filters.json.JSONFilter {
     XLIFFNoteAnnotation xliffNoteAnnotation;
     UsagesAnnotation usagesAnnotation;
     String currentKeyName;
+    String comment = null;
 
     @Override
     public String getName() {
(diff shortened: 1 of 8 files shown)


webapp/src/main/java/com/box/l10n/mojito/okapi/filters/JSONFilter.java
```diff
@@ -94,6 +101,15 @@ public void handleValue(String value, JsonValueTypes valueType) {
         extractNoteIfMatch(value);
         extractUsageIfMatch(value);
         super.handleValue(value, valueType);
+        if (comment != null) {
+            addXliffNote(comment);
```
aurambaj: can this be in the "handleComment" function?

webapp/src/main/java/com/box/l10n/mojito/okapi/filters/JSONFilter.java
```diff
@@ -94,6 +101,15 @@ public void handleValue(String value, JsonValueTypes valueType) {
...
+            ITextUnit textUnit = getEventTextUnit();
+            if (textUnit != null) {
+                textUnit.setAnnotation(xliffNoteAnnotation);
+            }
+            comment = null;
```
aurambaj: maybe put that logic in a function that says it handle "json comment" .

I suppose if you reset here it is because this is called after `handleObjectEnd`?

webapp/src/main/java/com/box/l10n/mojito/okapi/filters/JSONFilter.java
jeeyi: No, because when comment is being processed, textUnit is not yet created. TextUnit gets created in `handleValue`. The order of call is `handleComment`, `handleKey` and `handleValue` for the following code: 
```
{
  // Greeting from Main UI
  "hello": "Hello!"
}
```

webapp/src/main/java/com/box/l10n/mojito/okapi/filters/JSONFilter.java
jeeyi: Done.

jeeyi:
@aurambaj Test is fixed. Test resources are updated.

PR #478, opened 2019-08-17
Additional option --fully-translated for pull command

jeeyi:
This option is used to prevent partially translated localized files. When this option is specified, it will check the number of untranslated strings for the locale. If there are untranslated strings, the localized file is not generated. 

With this option, localized files are generated for partially translated locales (child locales that inherit translations from the parent locale) only if their parent locales are fully translated.

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
@@ -13,6 +13,8 @@ import com.box.l10n.mojito.rest.entity.Asset;
 import com.box.l10n.mojito.rest.entity.LocalizedAssetBody;
 import com.box.l10n.mojito.rest.entity.Repository;
 import com.box.l10n.mojito.rest.entity.RepositoryLocale;
+import com.box.l10n.mojito.rest.entity.RepositoryLocaleStatistic;
+import com.box.l10n.mojito.rest.entity.RepositoryStatistic;
 import org.fusesource.jansi.Ansi.Color;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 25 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
```diff
@@ -98,13 +103,23 @@
...
+
     @Override
     public void execute() throws CommandException {
 
         consoleWriter.newLine().a("Pull localized asset from repository: ").fg(Color.CYAN).a(repositoryParam).println(2);
 
         repository = commandHelper.findRepositoryByName(repositoryParam);
 
+        if (onlyIfFullyTranslated) {
+            checkFullyTranslated(repository);
```
aurambaj: rename to `initLocaleFullyTranslatedMap`?

cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
```diff
@@ -164,8 +179,23 @@ void generateLocalizedFilesWithLocaleMaping(Repository repository, FileMatch sou
     }
 
     void generateLocalizedFile(Repository repository, FileMatch sourceFileMatch, List<String> filterOptions, String outputBcp47tag, RepositoryLocale repositoryLocale) throws CommandException {
-        LocalizedAssetBody localizedAsset = getLocalizedAsset(repository, sourceFileMatch, repositoryLocale, outputBcp47tag, filterOptions);
-        writeLocalizedAssetToTargetDirectory(localizedAsset, sourceFileMatch);
+        boolean skip = false;
+
+        if (onlyIfFullyTranslated) {
```
aurambaj: move that block to a function?

cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
jeeyi: Done

cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
jeeyi: Done

PR #480, opened 2019-08-21
JS/TS filter updated to support template strings in backquotes

jeeyi:
This is to support strings enclosed in backquote. For example:
```
// example1
"string1": `multi-line text
with \`escaped\` backquote
for testing`,

// example2
"string2": "this is `backquote` in \"double-quoted\" string\n",
```


jeeyi:
@aurambaj FYI, I referenced other okapi filters like PropertiesFilter but the implementation is different.

PR #490, opened 2019-09-25
Remove @DisallowConcurrentExecution since it is making the job store …

aurambaj:
…durably too and we don't care about concurrent execution anyway


webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/job/BranchNotificationJob.java
```diff
@@ -15,7 +15,6 @@
  * Job that sends notifications for a branch.
  */
 @Component
-@DisallowConcurrentExecution
 public class BranchNotificationJob extends SchedulableJob {
```
aurambaj: also remove BranchNotificationMissingScreenshotsJob

PR #497, opened 2019-10-14
Implement translation history UI

ehoogerbeets:
Now that the BE exposes the translation history of a text unit, this PR introduces the UI on the front end so that you can see it easily. The history is exposed as a clock button on each text unit:

Screen Shot 2019-10-14 at 14 00 19

When you click it, a dialog comes up with the translation history in it:

Screen Shot 2019-10-17 at 13 30 59

You can only view the history for the locale of the unit that you click on. You cannot (yet?) see the history of all translations for a source string.

The table is ordered in reverse chronological order. (ie. the latest is on the top of the table). The first/oldest entry in the list (ie. the bottom of the table) is information about when the text unit was initially created and it has no translation yet. 

In the screen shot, the 2nd & 3rd row have no translation because I deleted the text to make sure it still works. When there are many items in the list, the dialog grows dynamically and you can scroll the page to see them all.

Imports show up with the username of the person who performed the import.

Currently, you cannot do anything with this UI other than view the history. Later in a separate PR, we can implement some sort of "revert" functionality where you can go back to an old translation, or any other behaviours we would like to do.

### DIFF
--- a/webapp/src/main/resources/properties/en.properties
+++ b/webapp/src/main/resources/properties/en.properties
@@ -707,16 +707,13 @@ textUnit.gitBlameModal.screenshotModalOpen=View
 workbench.translationHistoryModal.info=Text unit history
 
 # Title of the translation history modal dialog
-workbench.translationHistoryModal.title=Text Unit Translation History
+workbench.translationHistoryModal.title=Translation History for Locale {locale}
 
 # Label for the button to close the translation history modal dialog
 textUnit.translationHistoryModal.close=Close
 
 # Text in the section in the translation history dialog where info about the translation unit is given
-textUnit.translationHistoryModal.locale=Translation history for locale: {locale}
-
-# Text in the section in the translation history dialog where info about the translation unit is given
-textUnit.translationHistoryModal.source=Source string: "{source}"
+textUnit.translationHistoryModal.source=Source: "{source}"
 
 # Label for the column that contains the username of the user that made the translation
 textUnit.translationHistoryModal.User=User Name
--- a/webapp/src/main/resources/properties/en.properties
+++ b/webapp/src/main/resources/properties/en.properties
@@ -712,8 +712,11 @@ workbench.translationHistoryModal.title=Text Unit Translation History
 # Label for the button to close the translation history modal dialog
 textUnit.translationHistoryModal.close=Close
 
-# Label for the section in the translation history dialog where the table of historical translations appear
-textUnit.translationHistoryModal.header=Translation history for locale: {locale}\nSource string: "{source}"
+# Text in the section in the translation history dialog where info about the translation unit is given
+textUnit.translationHistoryModal.locale=Translation history for locale: {locale}
+
+# Text in the section in the translation history dialog where info about the translation unit is given
+textUnit.translationHistoryModal.source=Source string: "{source}"
 
 # Label for the column that contains the username of the user that made the translation
 textUnit.translationHistoryModal.User=User Name
@@ -729,3 +732,6 @@ textUnit.translationHistoryModal.Status=Status
 
 # Body of the translation history modal dialog if there are no translations on this text unit yet
 textUnit.translationHistoryModal.NoTranslations=No translation history yet.
+
+# String shown in the translation field when the text unit is first submitted to mojito and there is no translation yet
+textUnit.translationHistoryModal.InitialPush=<no translation yet>
--- a/webapp/src/main/resources/properties/en.properties
+++ b/webapp/src/main/resources/properties/en.properties
@@ -702,3 +702,6 @@ textUnit.gitBlameModal.screenshots=Screenshots
 
 # Text for link to open screenshot modal in git-blame modal
 textUnit.gitBlameModal.screenshotModalOpen=View
+
+# Label for button to open the translation history modal
+workbench.translationHistoryModal.info=Text unit history
--- a/webapp/src/main/resources/properties/en.properties
+++ b/webapp/src/main/resources/properties/en.properties
@@ -703,5 +703,29 @@ textUnit.gitBlameModal.screenshots=Screenshots
 # Text for link to open screenshot modal in git-blame modal
 textUnit.gitBlameModal.screenshotModalOpen=View
 
-# Label for button to open the translation history modal
+# Tooltip for button to open the translation history modal
 workbench.translationHistoryModal.info=Text unit history
+
+# Title of the translation history modal dialog
+workbench.translationHistoryModal.title=Text Unit Translation History
+
+# Label for the button to close the translation history modal dialog
+textUnit.translationHistoryModal.close=Close
+
+# Label for the section in the translation history dialog where the table of historical translations appear
+textUnit.translationHistoryModal.header=Translation history for locale: {locale}\nSource string: "{source}"
+
+# Label for the column that contains the username of the user that made the translation
+textUnit.translationHistoryModal.User=User Name
+
+# Label for the column that contains the translation
+textUnit.translationHistoryModal.Translation=Translation
+
+# Label for the column that contains the date that the translation was submitted
+textUnit.translationHistoryModal.Date=Date
+
+# Label for the column that contains the status of the translation
+textUnit.translationHistoryModal.Status=Status
+
+# Body of the translation history modal dialog if there are no translations on this text unit yet
+textUnit.translationHistoryModal.NoTranslations=No translation history yet.
--- a/webapp/src/main/resources/public/js/components/workbench/TextUnit.js
+++ b/webapp/src/main/resources/public/js/components/workbench/TextUnit.js
@@ -830,14 +830,11 @@ let TextUnit = createReactClass({
                         <span className="textunit-gitInfo glyphicon glyphicon-info-sign mls"
                               onClick={this.onTextUnitInfoClick} />
                     </OverlayTrigger>
-                    {
-                        (this.props.textUnit.getTmTextUnitVariantId()) ?
-                            <OverlayTrigger placement="top" overlay={assetPathTranslationHistoryTooltip}>
-                                <span className="textunit-translation-history glyphicon glyphicon-time mls"
-                                      onClick={this.onTranslationHistoryClick} />
-                            </OverlayTrigger> :
-                            ""
-                    }
+
+                    <OverlayTrigger placement="top" overlay={assetPathTranslationHistoryTooltip}>
+                        <span className="textunit-translation-history glyphicon glyphicon-time mls"
+                              onClick={this.onTranslationHistoryClick} />
+                    </OverlayTrigger>
                 </span>
         );
     },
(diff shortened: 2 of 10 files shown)


webapp/src/main/resources/public/js/components/workbench/TextUnit.js
```diff
@@ -805,18 +813,28 @@ let TextUnit = createReactClass({
         let assetPathWithGitInfoTooltip =
             <Tooltip id="{this.props.textUnit.getId()}-gitInfo">{this.props.intl.formatMessage( {id: 'workbench.gitBlameModal.info'} )}</Tooltip>;
 
+        let assetPathTranslationHistoryTooltip =
+            <Tooltip id="{this.props.textUnit.getId()}-translation-history">{this.props.intl.formatMessage( {id: 'workbench.translationHistoryModal.info'} )}</Tooltip>;
```
aurambaj: I'm looking at the current code: `"{this.props.textUnit.getId()}-gitInfo"` and it doesn't seem to work. So I suspect this doesn't work either. Could you fix both?

webapp/src/main/resources/public/js/components/workbench/TranslationHistoryModal.js
```diff
@@ -0,0 +1,122 @@
...
+        return textUnit ? (
+            <Modal className={"git-blame-modal"} show={show} onHide={this.closeModal}>
+                <Modal.Header closeButton>
+                    <Modal.Title><FormattedMessage id="workbench.translationHistoryModal.title"/></Modal.Title>
+                </Modal.Header>
+                <Modal.Body>
+                    <div className={"row"}>
+                        <div className={"col-sm-4"}>
+                            <div>
+                                <FormattedMessage id={"textUnit.translationHistoryModal.locale"} values={{
```
aurambaj: Actually, can we merge this into the title since it is almost the same string?

If not merging I'd have a table to align everything, also see comment bellow

webapp/src/main/resources/public/js/components/workbench/TranslationHistoryModal.js
```diff
@@ -0,0 +1,122 @@
...
+        this.props.onCloseModal();
+    };
+
+    render() {
+        const { textUnit, show } = this.props;
+
+        return textUnit ? (
+            <Modal className={"git-blame-modal"} show={show} onHide={this.closeModal}>
+                <Modal.Header closeButton>
+                    <Modal.Title><FormattedMessage id="workbench.translationHistoryModal.title"/></Modal.Title>
```
aurambaj: I believe "Translation History" is enough, I'd remove the "Text Unit"

webapp/src/main/resources/properties/en.properties
```diff
@@ -701,4 +701,37 @@ textUnit.gitBlameModal.close=Close
...
+workbench.translationHistoryModal.title=Text Unit Translation History
+
+# Label for the button to close the translation history modal dialog
+textUnit.translationHistoryModal.close=Close
+
+# Text in the section in the translation history dialog where info about the translation unit is given
+textUnit.translationHistoryModal.locale=Translation history for locale: {locale}
+
+# Text in the section in the translation history dialog where info about the translation unit is given
+textUnit.translationHistoryModal.source=Source string: "{source}"
```
aurambaj: I'd have just "Source" and remove "string"

webapp/src/main/resources/properties/en.properties
```diff
@@ -701,4 +701,37 @@ textUnit.gitBlameModal.close=Close
...
 # Text for link to open screenshot modal in git-blame modal
-textUnit.gitBlameModal.screenshotModalOpen=View
\ No newline at end of file
+textUnit.gitBlameModal.screenshotModalOpen=View
+
+# Tooltip for button to open the translation history modal
+workbench.translationHistoryModal.info=Text unit history
+
+# Title of the translation history modal dialog
+workbench.translationHistoryModal.title=Text Unit Translation History
```
aurambaj: Maybe just have "Translation History" here? also comment later that we could merge the title and the first line.

webapp/src/main/resources/public/js/components/workbench/TranslationHistoryModal.js
```diff
@@ -0,0 +1,122 @@
...
+                </Modal.Header>
+                <Modal.Body>
+                    <div className={"row"}>
+                        <div className={"col-sm-4"}>
+                            <div>
+                                <FormattedMessage id={"textUnit.translationHistoryModal.locale"} values={{
+                                    locale: textUnit.getTargetLocale()
```
aurambaj: layout wise maybe we want to copy what's in the info modal here for consistency?
I can't find a way to attach a screenshot easily but basically the table with the first column in bold and the value in the second column

webapp/src/main/resources/public/js/components/workbench/TranslationHistoryModal.js
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * @returns {*} Generated content for the git blame information section
+     */
+    rendertranslationHistory = () => {
+        const { translationHistory, textUnit, intl } = this.props;
+
+        return (
+            <Table striped bordered hover responsive="sm">
+                <thead>
```
aurambaj: the style looks different from other tables. Can we keep similar CSS as the repository/drop pages?

webapp/src/main/resources/public/js/stores/workbench/TranslationHistoryStore.js
```diff
@@ -0,0 +1,51 @@
...
+        this.translationHistory = null;
+        this.loading = false;
+    }
+
+    close() {
+        this.show = false;
+    }
+
+    openWithTextUnit(textUnit) {
+        console.log("TranslationHistoryStore::openWithTextUnit");
```
aurambaj: remove :)

webapp/src/main/resources/public/js/stores/workbench/TranslationHistoryStore.js
```diff
@@ -0,0 +1,51 @@
...
+
+        this.show = true;
+        this.textUnit = textUnit;
+        this.translationHistory = null;
+        this.loading = true;
+        this.getInstance().getTranslationHistory(textUnit);
+    }
+
+    onGetTranslationHistorySuccess(translationHistory) {
+        console.log("TranslationHistoryStore::onGetTranslationHistorySuccess");
```
aurambaj: same

webapp/src/main/resources/public/js/stores/workbench/TranslationHistoryStore.js
```diff
@@ -0,0 +1,51 @@
...
+        this.translationHistory = translationHistory;
+        this.loading = false;
+    }
+
+    onGetTranslationHistoryError(errorResponse) {
+        console.log("TranslationHistoryStore::onGetTranslationHistoryError");
```
aurambaj: same

webapp/src/main/resources/public/js/components/workbench/TranslationHistoryModal.js
```diff
@@ -0,0 +1,122 @@
...
+            </Table>
+        );
+    };
+
+    /**
+     * @param date   An integer representing a datetime
+     * @returns {*}  Human readable version of the given datetime
+     *
+     */
+    convertDateTime = (date) => {
```
aurambaj: should this be in a helper class?

webapp/src/main/resources/public/js/components/workbench/TranslationHistoryModal.js
```diff
@@ -0,0 +1,122 @@
...
+            "textUnit": PropTypes.object.isRequired,
+            "translationHistory": PropTypes.object.isRequired
+        };
+    }
+
+    renderHistoryItem = (item) => {
+        return item ?
+            (
+                <tr>
+                    <td>{item.createdByUser.username}</td>
```
aurambaj: vertical alignment (text - icon) seems off? maybe because of the icon size?

webapp/src/main/resources/sass/mojito.scss
```diff
@@ -479,6 +488,10 @@ div.targetstring-container textarea {
     width: 60%;
 }
 
+.history-initial {
+    color: $brand-primary;
```
aurambaj: I guess the primary color should be used for "important" things. Here it seems to be a minor information, should we use: light gray + italic?

aurambaj:
> Screen Shot 2019-10-14 at 14 00 19
👍

jeeyi:
Could we polish the UI so that it looks consistent with other table (repository page or project request page)? No bold text in the header, consistent heights and alignment, etc.

Screen Shot 2019-10-14 at 11 13 45 PM

webapp/src/main/resources/public/js/components/workbench/TextUnit.js
ehoogerbeets: Done

webapp/src/main/resources/public/js/components/workbench/TranslationHistoryModal.js
ehoogerbeets: The icon beside it has the tooltip "Text unit info", so I made this one "Text unit history" to parallel that.

webapp/src/main/resources/public/js/components/workbench/TextUnit.js
aurambaj: thanks!

webapp/src/main/resources/public/js/components/workbench/TranslationHistoryModal.js
ehoogerbeets: Done

webapp/src/main/resources/properties/en.properties
ehoogerbeets: Done

webapp/src/main/resources/properties/en.properties
ehoogerbeets: Merged with title

webapp/src/main/resources/public/js/stores/workbench/TranslationHistoryStore.js
ehoogerbeets: done

webapp/src/main/resources/public/js/stores/workbench/TranslationHistoryStore.js
ehoogerbeets: done

webapp/src/main/resources/public/js/components/workbench/TranslationHistoryModal.js
ehoogerbeets: Probably...

webapp/src/main/resources/public/js/components/workbench/TranslationHistoryModal.js
ehoogerbeets: After addressing Jee's comment below, it is aligned now

webapp/src/main/resources/sass/mojito.scss
ehoogerbeets: done

aurambaj:
I've tried it, it is pretty cool! Thanks @ehoogerbeets !

PR #501, opened 2019-11-01
Avoid empty xliffs

ehoogerbeets:
Modify the BE code to not generate an xliff file for a locale in a drop if there are no translation units with translations needed for that particular locale. Also, changed the import so that if there are no translations needed for a locale, it does not mark the imported drop as partially imported. It only checks the locales where translations were needed.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/translationkit/TranslationKitService.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/translationkit/TranslationKitService.java
@@ -18,6 +18,7 @@ import com.box.l10n.mojito.service.locale.LocaleService;
 import com.box.l10n.mojito.service.repository.RepositoryLocaleRepository;
 import com.box.l10n.mojito.service.tm.TMTextUnitRepository;
 import com.box.l10n.mojito.service.tm.search.StatusFilter;
+import com.box.l10n.mojito.service.tm.search.TextUnitAndWordCount;
 import com.box.l10n.mojito.service.tm.search.TextUnitDTO;
 import com.box.l10n.mojito.service.tm.search.TextUnitSearcher;
 import com.box.l10n.mojito.service.tm.search.TextUnitSearcherParameters;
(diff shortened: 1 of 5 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/translationkit/TranslationKitService.java
```diff
@@ -102,6 +103,13 @@ public TranslationKitAsXliff generateTranslationKitAsXLIFF(Long dropId, Long tmI
         logger.debug("Get translation kit for in tmId: {} and locale: {}", tmId, localeId);
         TranslationKit translationKit = addTranslationKit(dropId, localeId, type);
 
+        Long repositoryId = translationKit.getDrop().getRepository().getId();
+        StatusFilter statusFilter = TranslationKit.Type.TRANSLATION.equals(type) ? StatusFilter.FOR_TRANSLATION : StatusFilter.REVIEW_NEEDED;
+
+        if (isEmpty(repositoryId, localeId, statusFilter, useInheritance)) {
+            return null;
```
aurambaj: I don't think it is a good pattern to return "null" to convey the notion of an empty translation kit. I would keep creating the object as before but have an attribute on TranslationKitAsXliff that says it is empty which can be used to export or not the file

webapp/src/main/java/com/box/l10n/mojito/service/translationkit/TranslationKitService.java
```diff
@@ -396,4 +404,44 @@ public void checkForPartiallyImported(Long dropId) {
 
         return textUnitDTOs;
     }
+
+    private boolean isEmptyForStatus(Long repositoryId, Long localeId, StatusFilter statusFilter) {
+        TextUnitAndWordCount textUnitAndWordCount;
```
aurambaj: that seems to be doing extra queries and potentially may get out of sync the TranslationKitStep logic. 

TranslationKitStep computes the number of words (`wordCount` attribute) that that can be used to check for empty kit. It is also saved in the translationKit DB entry. I would rely one the already existing word count and add it to TranslationKitAsXliff.

Not sure what is the best between re-reading from the DB entity or getting from the step (or potentially use an annotation). It is pretty similar.

PR #504, opened 2019-11-21
Added filter by count of sub branches

SergeyDemyanenko:


webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticSpecification.java
```diff
@@ -77,4 +77,16 @@ public Predicate toPredicate(Root<BranchStatistic> root, CriteriaQuery<?> query,
         };
     }
 
+    public static SingleParamSpecification<BranchStatistic> empty(final Boolean empty) {
```
aurambaj: that can be made a `boolean`

webapp/src/main/resources/properties/be.properties
```diff
@@ -46,6 +46,9 @@ branches.deleted=Deleted
 # Branches search filter option label to indicate that the branch is deleted
 branches.searchstatusDropdown.deleted=Deleted
```
aurambaj: just modify the "en.properties" file, the other or auto-generated

PR #510, opened 2019-11-22
Specify user importing translations

jeeyi:
Now that we have translation history, we'd like to be able to distinguish changes done through the workbench and drops. 

Current implementation sets the user who pressed the import button as the one responsible for all translations in the drop. This changes provide a way to specify user for translations imported by the drop.

```
l10n.dropService.dropImporterUsername=vendor
```

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/translationkit/TranslationKitService.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/translationkit/TranslationKitService.java
@@ -292,7 +292,7 @@ public class TranslationKitService {
         List<TranslationKit> translationKits = translationKitRepository.findByDropId(drop.getId());
         boolean partiallyImported = false;
         for (TranslationKit translationKit : translationKits) {
-            if (!translationKit.getImported()) {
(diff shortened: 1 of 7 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/translationkit/TranslationKitService.java
```diff
@@ -292,7 +292,7 @@ public void checkForPartiallyImported(Long dropId) {
         List<TranslationKit> translationKits = translationKitRepository.findByDropId(drop.getId());
         boolean partiallyImported = false;
         for (TranslationKit translationKit : translationKits) {
-            if (!translationKit.getImported()) {
+            if (translationKit.getNumTranslationKitUnits() > 0 && !translationKit.getImported()) {
```
jeeyi: @ehoogerbeets This fixes a bug I found for the partially imported drop.

ehoogerbeets: 👍

PR #512, opened 2019-12-03
Added download method to SmartlingClient

SergeyDemyanenko:


webapp/src/test/java/com/box/l10n/mojito/smartling/SmartlingClientTest.java
```diff
@@ -10,10 +10,7 @@
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.io.ByteStreams;
-import org.junit.Assert;
```
aurambaj: let's avoid wildcard import. To configure Intellij: https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports

webapp/src/test/java/com/box/l10n/mojito/smartling/SmartlingClientTest.java
```diff
@@ -102,6 +99,22 @@ public void testUploadFile() {
 
     }
 
+    @Test
+    public void testDownloadFile() {
+        Assume.assumeNotNull(projectId);
```
aurambaj: can you do an upload in this test too, so that the test can be ran in isolation

PR #514, opened 2019-12-06
Add a mode to copy translations between text units with the leveraging command

aurambaj:
With mode "TUDS" it is possible to copy translation between text units of any projects. 
--tuids-mapping option is used to provide the mapping between the source and target tm text unit ids. All translations that are not rejected are copied (regardless of locale supported by the target project)

Example: leveraging-copy-tm -m TUID --tuids-mapping "12:33;32:23"


aurambaj:
@jeeyi  that might be useful for you as well :)

PR #515, opened 2019-12-10
EnumConverter class added

SergeyDemyanenko:


cli/src/main/java/com/box/l10n/mojito/cli/command/CopyTmConfigModeConverter.java
```diff
@@ -0,0 +1,15 @@
+package com.box.l10n.mojito.cli.command;
+
+import com.box.l10n.mojito.rest.entity.CopyTmConfig;
+
+/**
+ *
+ * @author jaurambault
```
aurambaj: that should be your name 😄

PR #516, opened 2019-12-10
Add an in-memory implementation of ThirdPartyTMS interface which is n…

aurambaj:
…ow default impl.

This allow to have a fully working webapp and to run the thirdparty-sync commands in demo mode. Strictly speaking, that has no real interest since the data won't be accessible but at least it is working configuration (before the backend would throw NPE and the command fail).

This will be useful for testing though.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSInMemory.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSInMemory.java
@@ -1,0 +1,66 @@
+package com.box.l10n.mojito.service.thirdparty;
+
+import com.box.l10n.mojito.entity.Repository;
+import com.box.l10n.mojito.smartling.request.Binding;
+import com.box.l10n.mojito.smartling.request.Bindings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.UUID;
+import java.util.regex.Pattern;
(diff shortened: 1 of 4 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSInMemory.java
```diff
@@ -0,0 +1,66 @@
...
+public class ThirdPartyTMSInMemory implements ThirdPartyTMS {
+
+    static Logger logger = LoggerFactory.getLogger(ThirdPartyTMSInMemory.class);
+
+    HashMap<String, byte[]> images = new HashMap<>();
+    HashMap<String, HashMap<String, String>> imageToTextUnitMappings = new HashMap<>();
+
+    /**
+     * Storing the uploaded images, that could be useful for testing but at the moment is useless...
+     * TODO remove if we don't need for testing
```
aurambaj: @SergeyDemyanenko storing the data might be useful for testing but if not, it can be removed... doesn't hurt for now

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSInMemory.java
```diff
@@ -0,0 +1,66 @@
...
+        ThirdPartyTMSImage thirdPartyTMSImage = new ThirdPartyTMSImage();
+        thirdPartyTMSImage.setId(imageId);
+        return thirdPartyTMSImage;
+    }
+
+    @Override
+    public List<ThirdPartyTextUnit> getThirdPartyTextUnits(Repository repository, String projectId) {
+        logger.debug("Get third party text units for repository: {} and project id: {}", repository.getId(), projectId);
+
+        // TODO return empty for now but later the interface will support adding strings and so this can retrun them
```
aurambaj: that would be more useful for actual testing. As the rest of script is migrated, 2 methods should be added to the ThirdPartyTMS interface (something like pushSourceTextUnits() and pushTranslations()). 

So this method will have a meaningful implementation later for tests.

PR #517, opened 2019-12-10
Added console command for sync

SergeyDemyanenko:


cli/src/main/java/com/box/l10n/mojito/cli/command/ThirdPartySyncCommand.java
```diff
@@ -38,6 +43,12 @@
     @Parameter(names = {"--project", "-p"}, arity = 1, required = true, description = "Third party project to synchronize with")
     String thirdPartyProjectId;
 
+    @Parameter(names = {"--actions", "-a"}, arity = 1, required = false, description = "Actions to synchronize", listConverter = ThirdPartySyncActionsConverter.class)
+    List<ThirdPartySync.Action> actions = Arrays.asList(ThirdPartySync.Action.MAP_TEXTUNIT, ThirdPartySync.Action.PUSH_SCREENSHOT);
+
+    @Parameter(names = {"--options", "-o"}, arity = 1, required = false, description = "Options to synchronize")
```
aurambaj: i think it should be  `variableArity = true,` since it is a list

cli/src/main/java/com/box/l10n/mojito/cli/command/ThirdPartySyncCommand.java
```diff
@@ -38,6 +43,12 @@
     @Parameter(names = {"--project", "-p"}, arity = 1, required = true, description = "Third party project to synchronize with")
     String thirdPartyProjectId;
 
+    @Parameter(names = {"--actions", "-a"}, arity = 1, required = false, description = "Actions to synchronize", listConverter = ThirdPartySyncActionsConverter.class)
```
aurambaj: i think it should be variableArity = true, since it is a list. Or do you want to convert a single param into the list

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyService.java
```diff
@@ -81,18 +89,38 @@
...
     }
 
-    void syncMojitoWithThirdPartyTMS(Long repositoryId, String thirdPartyProjectId) {
+    void syncMojitoWithThirdPartyTMS(Long repositoryId, String thirdPartyProjectId, List<Action> actions, List<String> options) {
         logger.debug("thirdparty TMS: {}", thirdPartyTMS);
         Repository repository = repositoryRepository.findOne(repositoryId);
-        mapMojitoAndThirdPartyTextUnits(repository, thirdPartyProjectId);
-        uploadScreenshotsAndCreateMappings(repository, thirdPartyProjectId);
+
+        actions.sort((o1, o2) -> Integer.compare(o1.ordinal(), o2.ordinal()));
```
aurambaj: order might be clearer with a "if" statement. 

if someone changes the order in enum it won't be obvious that this is impacted and will break silently. Or document the order in the enum + test to enforce order.

cli/src/main/java/com/box/l10n/mojito/cli/command/ThirdPartySyncCommand.java
```diff
@@ -38,6 +43,12 @@
     @Parameter(names = {"--project", "-p"}, arity = 1, required = true, description = "Third party project to synchronize with")
     String thirdPartyProjectId;
 
+    @Parameter(names = {"--actions", "-a"}, variableArity = true, required = false, description = "Actions to synchronize", converter = ThirdPartySyncActionsConverter.class)
+    List<ThirdPartySync.Action> actions = Arrays.asList(ThirdPartySync.Action.MAP_TEXTUNIT, ThirdPartySync.Action.PUSH_SCREENSHOT);
+
+    @Parameter(names = {"--options", "-o"}, variableArity = true, required = false, description = "Options to synchronize")
+    List<String> options = new ArrayList<>();
```
aurambaj: nit pick, but for other list like "filterOptions" we don't have an empty array as default value. It makes the CLI doc look different. I'd rather remove it to have something more consistent.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyService.java
```diff
@@ -81,18 +89,36 @@
...
 
-    void syncMojitoWithThirdPartyTMS(Long repositoryId, String thirdPartyProjectId) {
+    void syncMojitoWithThirdPartyTMS(Long repositoryId, String thirdPartyProjectId, List<Action> actions, List<String> options) {
         logger.debug("thirdparty TMS: {}", thirdPartyTMS);
         Repository repository = repositoryRepository.findOne(repositoryId);
-        mapMojitoAndThirdPartyTextUnits(repository, thirdPartyProjectId);
-        uploadScreenshotsAndCreateMappings(repository, thirdPartyProjectId);
+
+        if (actions.contains(Action.PUSH)) {
+            // TODO
```
aurambaj: could throw `UnsupportedOperationException` until it is implemented

PR #521, opened 2019-12-19
refactoring, deprecated IO method

SergeyDemyanenko:


webapp/src/main/java/com/box/l10n/mojito/boxsdk/BoxSDKService.java
```diff
@@ -156,7 +157,7 @@ public BoxFile uploadFile(String folderId, String filename, String filecontent)
                 logger.debug("Uploaded new file, id: " + uploadFile.getID() + ", name: " + filename);
             } else {
                 logger.debug("Upload a new version of file named: {} to folder: {}", filename, folderId);
-                uploadFile.uploadVersion(IOUtils.toInputStream(filecontent));
+                uploadFile.uploadVersion(IOUtils.toInputStream(filecontent, Charset.defaultCharset()));
```
aurambaj: we must use utf8, not rely on possible different system encoding

SergeyDemyanenko: I think the default charset is utf8 on the mojito server.
I did just that, because I saw the following code:

```java    
    @Deprecated
    public static InputStream toInputStream(final String input) {
            return toInputStream(input, Charset.defaultCharset());
    }
```

PR #523, opened 2019-12-19
small log improvements

SergeyDemyanenko:


webapp/pom.xml
```diff
@@ -16,7 +16,7 @@
 
     <properties>
         <start-class>com.box.l10n.mojito.Application</start-class>
-        <frontend.maven.plugin.version>1.0</frontend.maven.plugin.version>
+        <frontend.maven.plugin.version>1.8.0</frontend.maven.plugin.version>
```
aurambaj: @SergeyDemyanenko is that on purpose? asking because the PR name is about "logging"

SergeyDemyanenko: Yes, it is. The new version fixes some issues with logging.

SergeyDemyanenko: Yes, it is. The new version fixes some issues with logging.

PR #527, opened 2020-01-09
Improve matching for plural text units

aurambaj:
Before a text unit with same prefix as the name of a singular text unit could be match to it. Now we match first the plural entries and then match by name. Also match unused plural text units.


aurambaj:
FYI @SergeyDemyanenko

PR #529, opened 2020-01-14
New common module, and command to perform local extractions and diffs

aurambaj:


aurambaj:
@jeeyi Having "common" module makes the CLI a little bit bigger (37M vs 22M) but gives the advantage to be able to share okapi/extraction logic between the Webapp and CLI. Let me know if this is a problem on your side. Thanks!

PR #530, opened 2020-01-16
drop-export command to handle empty drop

jeeyi:
When creating a drop from CLI, we are now checking number of strings for translation. If there is nothing to be translated, the drop is not created.

```
$ mojito drop-export -r Demo

Export a Drop from repository: Demo


Repository is already fully translated
```


cli/src/main/java/com/box/l10n/mojito/cli/command/DropExportCommand.java
```diff
@@ -132,24 +140,48 @@ public void execute() throws CommandException {
...
-
+    
+    private boolean shouldCreateDrop(Repository repository) {
+        boolean createDrop = false;
+        Set<String> localesforTranslation = getLocalesForTranslation(repository);
+        RepositoryStatistic repoStat = repository.getRepositoryStatistic();
+        if (repoStat != null) {
+            for (RepositoryLocaleStatistic repoLocaleStat : repoStat.getRepositoryLocaleStatistics()) {
+                if (localesforTranslation.contains(repoLocaleStat.getLocale().getBcp47Tag()) && repoLocaleStat.getForTranslationCount() > 0L) {
+                    return true;
```
aurambaj: replace with: `createDrop = true` and `break;`?

cli/src/main/java/com/box/l10n/mojito/cli/command/DropExportCommand.java
```diff
@@ -132,24 +140,48 @@ public void execute() throws CommandException {
...
+                }
+            }
+        }
+        return createDrop;
+    }
+    
+    private Set<String> getLocalesForTranslation(Repository repository) {
+        Set<String> localesForTranslation = new HashSet<>();
```
aurambaj: isn't it the same logic as `getBcp47TagsForExportFromRepository()` maybe just have `Sets.newHashSet(getBcp47TagsForExportFromRepository(repository));` ?

cli/src/test/java/com/box/l10n/mojito/cli/command/DropExportCommandTest.java
```diff
@@ -32,24 +32,43 @@ public void export() throws Exception {
...
 
         Asset asset = assetClient.getAssetByPathAndRepositoryId("source-xliff.xliff", repository.getId());
         importTranslations(asset.getId(), "source-xliff_", "fr-FR");
         importTranslations(asset.getId(), "source-xliff_", "ja-JP");
 
-        Asset asset2 = assetClient.getAssetByPathAndRepositoryId("source2-xliff.xliff", repository.getId());
-        importTranslations(asset2.getId(), "source2-xliff_", "fr-FR");
-        importTranslations(asset2.getId(), "source2-xliff_", "ja-JP");
+        // wait for stats to be updated
+        Thread.sleep(1000);
```
aurambaj: can you replace with `waitForCondition` please?

like in 

```
    waitForCondition("b4 should become untranslated when stats are computed", () -> {
            List<String> branches = repositoryClient.getBranches(repository.getId(), null, null, null,
                    false, false, null).stream().map(Branch::getName).sorted().collect(Collectors.toList());
            return branches.equals(asList("b4"));
        });
```

cli/src/main/java/com/box/l10n/mojito/cli/command/DropExportCommand.java
jeeyi: Done

cli/src/test/java/com/box/l10n/mojito/cli/command/DropExportCommandTest.java
jeeyi: See waitForRepositoryToHaveStringsForTranslations. Thank you!

PR #536, opened 2020-01-23
Third party sync (smartling)

SergeyDemyanenko:


webapp/src/main/java/com/box/l10n/mojito/service/tm/search/TextUnitSearcherParameters.java
```diff
@@ -50,235 +50,266 @@ public String getName() {
         return name;
     }
 
-    public void setName(String name) {
+    public TextUnitSearcherParameters setName(String name) {
         this.name = name;
+        return this;
```
aurambaj: This class is just a simple POJO and should stay that way. We could have a builder but let's not mix both style. For now we can just to do with the normal setters.

PR #549, opened 2020-02-22
Support local extractions with diff computation and push of asset with extracted content

aurambaj:


cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionDiffCommand.java
```diff
@@ -2,46 +2,54 @@
...
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 
+import java.nio.file.Path;
+import java.util.stream.Stream;
+
 /**
  * Computes a diff between to local extractions.
- *
+ * <p>
```
aurambaj: @jeeyi FYI, this is a new command that is handy to process pull-request/diffs without calling the server at all. Eventually, if there are new translation units it can sync the delta with the server

jeeyi: This is super helpful. Thank you @aurambaj!

PR #559, opened 2020-03-11
Move docs for gh-pages branch into master

aurambaj:
Github now supports docs from a specific directory in master, no need for the gh-pages.
It will make doc versioning easier.


aurambaj:
@jeeyi Moving the docs in master. When this is landed, could you reconfigure the project to use the directory on master (`Settings > GitHub Pages > Source : master branch/docs folder`)? Thanks!

PR #572, opened 2020-04-20
Async localized files generation and blob storage

aurambaj:


aurambaj:
@jeeyi FYI, big changes in this PR. We can now execute an async job and have an output for it, all tracked via PollableTask. This change has no impact on current tools. It will be used only if passing an option to the `pull` command. It has a default implementation backed by a database but to scale there is also a S3 implementation. That's what we're gonna use so the database implementation will be less tested. 

As it proves stable though we can start using that more widely: making the default `pull` command async, TMXliff export could use that too, and more.

PR #576, opened 2020-05-01
Update translation history UI

ehoogerbeets:
- highlight the current translation
- add explanatory text for that highlight
- add explanatory text when the latest translation is not the current one

### DIFF
--- a/webapp/src/main/resources/properties/en.properties
+++ b/webapp/src/main/resources/properties/en.properties
@@ -719,10 +719,10 @@ workbench.translationHistoryModal.title=Translation History
 textUnit.translationHistoryModal.close=Close
 
 # Text telling the user why one of the rows is highlighted
-textUnit.translationHistoryModal.currentUnitExplanation=The highlighted row is the currently approved translation.
+textUnit.translationHistoryModal.currentUnitExplanation=The highlighted row is the current translation as shown in the workbench.
 
 # Text explaining why the currently approved translation is not the latest
-textUnit.translationHistoryModal.currentNotLatest=In some cases, the currently approved translation is not the most recent one.
+textUnit.translationHistoryModal.currentNotLatest=Note that some edits happened in parallel. More recent translations from a review request were imported but not set as current translation.
 
 # Label for the column that contains the username of the user that made the translation
 textUnit.translationHistoryModal.User=User Name
--- a/webapp/src/main/resources/properties/en.properties
+++ b/webapp/src/main/resources/properties/en.properties
@@ -718,6 +718,12 @@ workbench.translationHistoryModal.title=Translation History
 # Label for the button to close the translation history modal dialog
 textUnit.translationHistoryModal.close=Close
 
+# Text telling the user why one of the rows is highlighted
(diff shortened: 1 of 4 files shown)


webapp/src/main/resources/properties/en.properties
```diff
@@ -718,6 +718,12 @@ workbench.translationHistoryModal.title=Translation History
 # Label for the button to close the translation history modal dialog
 textUnit.translationHistoryModal.close=Close
 
+# Text telling the user why one of the rows is highlighted
+textUnit.translationHistoryModal.currentUnitExplanation=The highlighted row is the currently approved translation.
```
jeeyi, line 722: Instead of `currently approved translation`, use `current translation`. We sometimes have translation with `Needs Review` or `Rejected` status.

webapp/src/main/resources/properties/en.properties
```diff
@@ -718,6 +718,12 @@ workbench.translationHistoryModal.title=Translation History
...
+
+# Text explaining why the currently approved translation is not the latest
+textUnit.translationHistoryModal.currentNotLatest=In some cases, the currently approved translation is not the most recent one.
```
jeeyi, line 725: ditto

aurambaj:
just saw that Jee commented before I had time to finish the review... so basically what she said for the translation status. 

Could you share a screenshot too?

webapp/src/main/resources/public/js/components/workbench/TranslationHistoryModal.js
```diff
@@ -87,7 +90,9 @@ class translationHistoryModal extends React.Component {
     };
 
     render() {
-        const {textUnit, show} = this.props;
+        const {textUnit, show, translationHistory} = this.props;
+        const currentVariantExists = translationHistory && translationHistory.length && textUnit.getTmTextUnitId();
```
jeeyi, line 94: Could we rename the variable to `translationExists`?

webapp/src/main/resources/properties/en.properties
aurambaj: I guess in that case we can be specific since it should not show often: "Some edits happened in parallel: more recent translations from a drop were imported but not set as current translation."

webapp/src/main/resources/public/js/components/workbench/TranslationHistoryModal.js
```diff
@@ -87,7 +90,9 @@ class translationHistoryModal extends React.Component {
...
+        const currentVariantIsNotLatest = currentVariantExists && textUnit.getTmTextUnitVariantId() !== translationHistory[0].id;
```
jeeyi, line 95: Rename the variable to `translationIsNotLatest`.

ehoogerbeets:
When there is no translation yet:

Screen Shot 2020-05-01 at 16 18 38

When the latest is the current:

Screen Shot 2020-05-01 at 16 18 10

When the latest is not the current:

Screen Shot 2020-05-01 at 16 17 57

aurambaj:
Looks good! 

For the regular case I wouldn't show "The highlighted row is the current translation as shown in the workbench." If we really want a flag, we could have a "label" like other places but for now I'd just remove it.

PR #586, opened 2020-06-09
Enable StatsD metrics names customization through configuration

comunacho:
Out of the box, Micrometer's statsd reporter doesn't allow to customize metric names beyond the flavor (Etsy, Datadog, etc) that the reporter uses for it.

With this change, we'll use `MeterRegistryCustomizer` to inject a customized `NamingConvention` whenever it is required. This customized `NamingConvention` will allow adding custom prefixes to each metric name, as well as custom tag name prefixes when required.

This custom metric naming schema can be enabled with:

`l10n.management.metrics.export.statsd.transform-name=true`

And the three configuration parameters we would be able to use are:

**NamingConvention:**
Defines the [naming convention](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java) that would be applied to the metric name (defaults to `identity`)
`
l10n.management.metrics.export.statsd.naming-convention=snakeCase|camelCase|upperCamelCase|slashes
`

**Name prefix:**
Defines the name prefix (or namespace) to attach to the metric name:
`
l10n.management.metrics.export.statsd.name-prefix=mojito
`

**Tag key prefix:**
Defines the tag key prefix to prefix to the tag key:
`
l10n.management.metrics.export.statsd.tag-key-prefix=_t_
`


webapp/src/main/java/com/box/l10n/mojito/rest/cli/CliWS.java
```diff
@@ -43,6 +45,7 @@
      * @throws IOException
      */
     @RequestMapping(value = "/cli/mojito-cli.jar", method = RequestMethod.GET)
+    @Timed
```
aurambaj, line 48: > What do I get out of the box?
> Spring Boot 2 autoconfigures quite a few metrics for you, including:
> 
> JVM, report utilization of:
> 
> Various memory and buffer pools
> 
> Statistics related to garbage collection
> 
> Thread utilization
> 
> Number of classes loaded/unloaded
> 
> CPU usage
> 
> Spring MVC and WebFlux request latencies

do we really need to add the `@Timed` annotations? 

> 
> RestTemplate latencies

webapp/pom.xml
```diff
@@ -214,6 +214,12 @@
         <!--            <version>2.2.3</version>-->
         <!--        </dependency>-->
 
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-registry-statsd</artifactId>
+            <version>1.5.1</version>
```
aurambaj, line 220: is there a default version defined in springboot?

webapp/src/main/java/com/box/l10n/mojito/utils/MetricsConfig.java
```diff
@@ -0,0 +1,96 @@
...
+            public String tagKey(String key) {
+                return tagKeyPrefix + transformation.transform(key);
+            }
+        };
+    }
+
+    @Bean
+    MeterRegistryCustomizer<MeterRegistry> meterRegistryCustomizer() {
+
+        return registry -> {
```
aurambaj, line 86: this applies to all implementation? Is there an impact on other backends?

webapp/src/main/java/com/box/l10n/mojito/utils/NamingTransformation.java
```diff
@@ -0,0 +1,66 @@
+package com.box.l10n.mojito.utils;
+
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import io.micrometer.core.instrument.util.StringUtils;
+
+public enum NamingTransformation {
```
aurambaj, line 10: what does NamingTransformation do exactly? Why do we need this? Can you add some tests?

webapp/src/main/java/com/box/l10n/mojito/utils/NamingTransformation.java
```diff
@@ -0,0 +1,70 @@
...
+
+    NONE(Function.identity()),
+
+    CAMEL_CASE(s -> {
+        String[] parts = s.split("\\.");
+        StringBuilder conventionName = new StringBuilder(s.length());
+        for (int i = 0; i < parts.length; i++) {
```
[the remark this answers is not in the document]
comunacho: I've used it for the case conversion on line 29. It can't be applied globally as Guava's `CaseFormat` doesn't have the input format (`some.metric.name`) specified in the enum

webapp/src/main/java/com/box/l10n/mojito/utils/StatsDMetricsConfig.java
```diff
@@ -0,0 +1,94 @@
...
+            @Override
+            public String tagValue(String value) {
+                return tagValuePrefix + nameTransformation.transform(value);
+            }
+        };
+    }
+
+    @Bean
+    @ConditionalOnProperty(value="l10n.management.metrics.export.statsd.transformName", havingValue="true")
+    MeterRegistryCustomizer<MeterRegistry> meterRegistryCustomizer() {
```
aurambaj, line 84: to target statds only registry: `MeterRegistryCustomizer<StatsdMeterRegistry>`? (I fear this would apply to all backends)

webapp/src/main/java/com/box/l10n/mojito/utils/StatsDMetricsConfig.java
```diff
@@ -0,0 +1,94 @@
...
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.config.NamingConvention;
+import io.micrometer.core.lang.Nullable;
+
+/**
+ * StatsD metrics related config
+ */
+@Component
+@ConfigurationProperties("l10n.management.metrics.export.statsd")
```
aurambaj, line 18: Can you keep the @ConfigurationProperties in its own class? no extra logic

webapp/src/main/java/com/box/l10n/mojito/utils/StatsDMetricsConfig.java
```diff
@@ -0,0 +1,94 @@
...
+@ConditionalOnProperty(value="management.metrics.export.statsd.enabled", havingValue="true")
```
aurambaj, line 19: Can you move the rest of the logic (conditional bean creation) into its own class?

webapp/src/main/java/com/box/l10n/mojito/utils/StatsDMetricsConfig.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private NamingTransformation nameTransformation = NamingTransformation.NONE;
+
+    private String tagKeyPrefix = "";
+
+    private String tagValuePrefix = "";
+
+    private String namePrefix = "";
+
+    public NamingTransformation getNameTransformation() {
```
aurambaj, line 30: do we really need to support name transformation? can't we just do adding prefix on top of an existing flavors?

webapp/src/main/java/com/box/l10n/mojito/utils/StatsDMetricsConfig.java
```diff
@@ -0,0 +1,94 @@
...
+
+    public String getNamePrefix() {
+        return namePrefix;
+    }
+
+    public void setNamePrefix(String namePrefix) {
+        this.namePrefix = namePrefix;
+    }
+
+    public NamingConvention buildNamingConvention() {
```
aurambaj, line 62: looks good, can you put it in its own class + tests

isn't it simpler to just reuse existing naming convention instead of having `NameTransformation` class?

something similar too

```
new NamingConvention() {
                @Override
                public String name(String name, Meter.Type type, String baseUnit) {
                    return camelCase.name("mojito" + name, type, baseUnit);
                }

                @Override
                public String tagKey(String key) {
                    return "_t_" + camelCase.tagKey(key);
                }
            }
```

webapp/pom.xml
```diff
@@ -213,6 +220,12 @@
         <!--            <version>2.2.3</version>-->
         <!--        </dependency>-->
 
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-registry-statsd</artifactId>
+            <version>${micrometer.version}</version>
```
aurambaj, line 226: de-dup

webapp/src/main/java/com/box/l10n/mojito/metrics/StatsdMetricsConfig.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+/**
+ * StatsD metrics related config
+ */
+@Component
+@ConfigurationProperties("l10n.management.metrics.export.statsd")
+@ConditionalOnProperty(value="management.metrics.export.statsd.enabled", havingValue="true")
```
aurambaj, line 12: can you split to have a class that only takes care of reading properties: @ConfigurationProperties

webapp/src/main/java/com/box/l10n/mojito/metrics/NamingConventionConverter.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.common.base.Strings;
+
+import org.springframework.boot.context.properties.ConfigurationPropertiesBinding;
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.stereotype.Component;
+
+import io.micrometer.core.instrument.config.NamingConvention;
+
+@Component
+@ConfigurationPropertiesBinding
```
comunacho, line 12: `@ConfigurationPropertiesBinding` allows Spring to use this Converter for configuration parameters. It'll be used for `NamingConvention` within `StatsdNamingConventionConfig`

webapp/src/main/java/com/box/l10n/mojito/metrics/NamingConventionBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+            namePrefix = config.getNamePrefix();
+        } else {
+            namePrefix = config.getNamePrefix() + ".";
+        }
+
+        return new NamingConvention() {
+
+            @Override
+            public String name(String name, Meter.Type type, @Nullable String baseUnit) {
+                return namePrefix + convention.name(name, type, baseUnit);
```
comunacho, line 26: With `NamingConvention` being defined via configuration, and without any re-direction, we can use the convention directly from here.

webapp/src/main/java/com/box/l10n/mojito/metrics/NamingConventionConverter.java
```diff
@@ -0,0 +1,42 @@
...
+public class NamingConventionConverter implements Converter<String, NamingConvention> {
+
+    @Override
+    public NamingConvention convert(String from) {
```
aurambaj, line 16: I've done something similar using reflection here: https://github.com/box/mojito/commit/1531cd746db5d6e7f1e64f3f0b4a74f42616c22c.

A little more concise and won't need to be extended if the NamingConvention changes

webapp/src/main/java/com/box/l10n/mojito/metrics/NamingConventionBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+import io.micrometer.core.lang.Nullable;
+
+public class NamingConventionBuilder {
+
+    public static NamingConvention build(StatsdNamingConventionConfig config) {
+
+        NamingConvention convention = config.getNamingConvention();
+        String namePrefix;
+
+        if (Strings.isNullOrEmpty(config.getNamePrefix()) || config.getNamePrefix().endsWith(".")) {
```
aurambaj, line 16: minor: I wouldn't do any treatment on the prefix (removing last "."), people should type exactly what they want there or else the behavior should be documented clearly to the user

webapp/src/main/java/com/box/l10n/mojito/metrics/NamingConventionBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.google.common.base.Strings;
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.config.NamingConvention;
+import io.micrometer.core.lang.Nullable;
+
+public class NamingConventionBuilder {
+
+    public static NamingConvention build(StatsdNamingConventionConfig config) {
```
aurambaj, line 11: can you avoid static methods and use spring component?

webapp/src/main/java/com/box/l10n/mojito/metrics/StatsdMeterRegistryCustomizer.java
```diff
@@ -0,0 +1,23 @@
...
+
+@Component
+public class StatsdMeterRegistryCustomizer {
+
+    @Bean
+    @ConditionalOnProperty(value="l10n.management.metrics.export.statsd.transform-name", havingValue="true")
+    public MeterRegistryCustomizer<StatsdMeterRegistry> customizer(StatsdNamingConventionConfig config) {
+
+        return registry -> {
+            if (config.changesNaming()){
```
aurambaj, line 18: do you still need that check since you're targetting "statsd" now with `MeterRegistryCustomizer<StatsdMeterRegistry>` and @ConditionalOnProperty(value="l10n.management.metrics.export.statsd.transform-name", havingValue="true") ?

webapp/src/test/java/com/box/l10n/mojito/metrics/HierarchicalNameMapperWithCustomNamingConventionTest.java
```diff
@@ -0,0 +1,322 @@
...
+import org.junit.jupiter.api.Test;
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.config.NamingConvention;
+import io.micrometer.core.instrument.util.HierarchicalNameMapper;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class HierarchicalNameMapperWithCustomNamingConventionTest {
```
aurambaj, line 13: `public class` ?

why is the class named `HierarchicalNameMapperWithCustomNamingConventionTest`? usually the unit test that goes with a class `ClassA` would be called `ClassATest`. I don't see `HierarchicalNameMapperWithCustomNamingConvention`

webapp/src/test/java/com/box/l10n/mojito/metrics/HierarchicalNameMapperWithCustomNamingConventionTest.java
```diff
@@ -0,0 +1,322 @@
...
+    @BeforeEach
+    void setUp() {
+        config = new StatsdNamingConventionConfig();
+
+        tags = Tags.of("tagA", "valA").and("tagB", "valB");
+        noTags = Tags.empty();
+    }
+
+    @Test
+    void withNamePrefixAndKeyPrefixAndIdentityDoesntChangesNames() {
```
aurambaj, line 30: I think `@Test` must be placed on with function that are "public"

webapp/src/test/java/com/box/l10n/mojito/metrics/StatsdNamingConventionConfigTest.java
```diff
@@ -0,0 +1,56 @@
+package com.box.l10n.mojito.metrics;
+
+import org.junit.jupiter.api.Test;
+
+import io.micrometer.core.instrument.config.NamingConvention;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.*;
+
+class StatsdNamingConventionConfigTest {
```
aurambaj, line 10: knit: Thank you for adding test, this is great. Usually i'm fine with no test for plain POJO with no business logic, including `@ConfigurationProperites`

webapp/src/main/java/com/box/l10n/mojito/metrics/StatsdNamingConventionConfig.java
```diff
@@ -0,0 +1,53 @@
...
+import io.micrometer.core.instrument.config.NamingConvention;
+
+/**
+ * StatsD naming convention config
+ *
+ * This configuration will be applied to statsd metric names when the top-level statsd reporter is enabled
+ */
+@Component
+@ConfigurationProperties("l10n.management.metrics.export.statsd")
+@ConditionalOnProperty(value="management.metrics.export.statsd.enabled", havingValue="true")
```
aurambaj, line 16: that's not needed is it?

aurambaj:
at this point, functionally wise speaking, this seems pretty much like doing what's here: https://github.com/box/mojito/commit/1531cd746db5d6e7f1e64f3f0b4a74f42616c22c. 
 
I don't see the benefits of the naming convention over `HierarchicalNameMapper`, even I think `HierarchicalNameMapper` is more appropriate in that case. 

we may want to merge your tests to have a more exhaustive coverage though.

webapp/src/test/java/com/box/l10n/mojito/metrics/StatsdNamingConventionConfigTest.java
comunacho: I agree, I tend not to add tests for units without business logic, but this one has the `changesNaming` method and I wanted to make sure it was doing what was intended.

webapp/src/test/java/com/box/l10n/mojito/metrics/HierarchicalNameMapperWithCustomNamingConventionTest.java
comunacho: I'm using JUnit 5 here, so there's no need to declare it as public.

Issue #591, opened 2020-06-17
Travis build status are not reported in Github


aurambaj:
We used to have travis build status reported on the branch/commit that were pushed to Github. Same was true for pull-request. Now, It is not working anymore. Looks like the build is done but reporting back to github doesn't happen

@jeeyi Looks like I don't have full access to travis anymore. Could you have a look please?

<img width="1074" alt="Screen Shot 2020-06-17 at 4 32 26 PM" src="https://user-images.githubusercontent.com/9702182/84961037-23d73980-b0b8-11ea-99d4-763e2451f471.png">

jeeyi:
It works for me.
<img width="621" alt="Screen Shot 2020-06-18 at 9 01 27 AM" src="https://user-images.githubusercontent.com/17402500/85044357-59c10000-b142-11ea-80e1-49dc6810140b.png">

Check your settings and check if Travis CI has access to your account. Settings -> Applications -> Authorized OAuth Apps:
<img width="827" alt="Screen Shot 2020-06-18 at 9 02 34 AM" src="https://user-images.githubusercontent.com/17402500/85044478-89700800-b142-11ea-87a2-5378b33305bb.png">

aurambaj:
seems to show up for me this morning too. Thanks for "the fix" :D

I've read on google that some people had intermittent issue like this... closing for now.

aurambaj:
@jeeyi this is happening again. One thing I don't remember exactly also is how the initial setup was done. I think we use the "pro" travis but in your screenshot there is both "Travis CI/pro" and "Travis CI for Open Source". As far as I remember one is on `travis-ci.org` and the other one is on `travis-ci.com`. When I go to `travis-ci.com` i don't see anything anymore which i guess make sense since I"m not in the Box organization anymore. I see the build in `travis-ci.org`, things are built it is just github that is not updated

aurambaj:
Setup Github action to build and run the test. It seems to have bring back the report for travis too. Will keep both for now

PR #595, opened 2020-06-19
Adding JobListener to report Quartz metrics

comunacho:
On top of what we've added to publish instrumentation metrics using micrometer and statsd, this change adds two classes:

- A Quartz `JobListener` that runs upon completion of each job defined within the codebase and publishes timing metrics for each job's execution time (in milliseconds) in the `quartz.jobs.execution` timer, including `jobGroup` and `jobClass` tags. To enable this metric, the `l10n.management.metrics.quartz.enabled` property must be `true`.
- A `Scheduled` task, which collects "queue" sizes and publishes them as gauges in the `quartz.pending.jobs` gauge, including `jobGroup` and `jobClass` tags. This task runs only when MySQL is being used as the datasource. To enable this metric, the `l10n.management.metrics.quartz.queue-monitoring.enabled` property must be `true`. Also, a `l10n.management.metrics.quartz.queue-monitoring.execution-rate` property is provided to execute the task with a fixed period in milliseconds between invocations. It can be defined in milliseconds, of following a [Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-) compliant string.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzMetricsReportingJobListener.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzMetricsReportingJobListener.java
@@ -1,0 +1,45 @@
+package com.box.l10n.mojito.monitoring;
+
+import org.quartz.JobExecutionContext;
+import org.quartz.JobExecutionException;
+import org.quartz.JobListener;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+
+import java.util.concurrent.TimeUnit;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+
+@ConditionalOnProperty(value = "l10n.management.metrics.quartz.enabled", havingValue = "true")
+@Component
+public class QuartzMetricsReportingJobListener implements JobListener {
+
+    private MeterRegistry meterRegistry;
+
+    public QuartzMetricsReportingJobListener(@Autowired MeterRegistry meterRegistry) {
+        this.meterRegistry = meterRegistry;
+    }
+
+    @Override
+    public String getName() {
+        return getClass().getSimpleName();
+    }
+
+    @Override
+    public void jobWasExecuted(JobExecutionContext context, JobExecutionException jobException) {
+        Tags timerTags = Tags.of("jobGroup", context.getJobDetail().getKey().getGroup(),
+                                 "jobClass", context.getJobDetail().getJobClass().getSimpleName());
+
+        meterRegistry.timer("quartz.jobs.execution", timerTags)
+                     .record(context.getJobRunTime(), TimeUnit.MILLISECONDS);
+    }
+
+    @Override
+    public void jobToBeExecuted(JobExecutionContext context) { }
+
+    @Override
+    public void jobExecutionVetoed(JobExecutionContext context) { }
+
+}
--- a/webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
@@ -1,0 +1,69 @@
+package com.box.l10n.mojito.monitoring;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Profile;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Component;
+
+import java.util.List;
+
+import javax.sql.DataSource;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+
+@Profile("!disablescheduling")
+@Component
+@ConditionalOnProperty(value = "l10n.management.metrics.quartz.sql-queue-monitoring.enabled", havingValue = "true")
+public class QuartzPendingJobsReportingTask {
+
+    private JdbcTemplate jdbcTemplate;
+    private MeterRegistry meterRegistry;
+
+    public QuartzPendingJobsReportingTask(@Autowired DataSource dataSource,
+                                          @Autowired MeterRegistry meterRegistry) {
+        this.meterRegistry = meterRegistry;
+        this.jdbcTemplate = new JdbcTemplate(dataSource);
+    }
+
+    @Scheduled(fixedRateString = "${l10n.management.metrics.quartz.sql-queue-monitoring.execution-rate}")
+    public void reportPendingJobs() {
+        fetchResults().forEach(this::reportResults);
+    }
+
+    private void reportResults(PendingJob pendingJob){
+        meterRegistry.gauge("quartz.pending.jobs",
+                            Tags.of("jobClass", pendingJob.jobClass, "jobGroup", pendingJob.jobGroup),
+                            pendingJob.count);
+    }
+
+    List<PendingJob> fetchResults(){
+        return jdbcTemplate.query(
+            "SELECT job_class_name, job_group, COUNT(*) FROM QRTZ_JOB_DETAILS GROUP BY job_class_name, job_group",
+            (rs, num) -> new PendingJob(extractClassName(rs.getString(1)), rs.getString(2), rs.getLong(3))
+        );
+    }
+
+    static String extractClassName(String input) {
+        String[] parts = input.split("\\.");
+        return parts.length > 0 ? parts[parts.length - 1] : "";
+    }
+
+    /*
+    * This class represents data associated with a group of jobs pending to be executed in our Quartz instance
+    * */
+    static class PendingJob {
+        public String jobClass;
+        public String jobGroup;
+        public Long   count;
+
+        public PendingJob(String jobClass, String jobGroup, Long count) {
+            this.jobClass = jobClass;
+            this.jobGroup = jobGroup;
+            this.count = count;
+        }
+    }
+
+}
--- a/webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
+++ b/webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
@@ -1,0 +1,215 @@
+package com.box.l10n.mojito.monitoring;
+
+import com.box.l10n.mojito.service.DBUtils;
+import com.box.l10n.mojito.service.assetExtraction.ServiceTestBase;
+import com.box.l10n.mojito.test.TestIdWatcher;
+
+import org.awaitility.core.ConditionFactory;
+import org.junit.Assume;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.quartz.Job;
+import org.quartz.JobDetail;
+import org.quartz.JobExecutionContext;
+import org.quartz.Scheduler;
+import org.quartz.SchedulerException;
+import org.quartz.Trigger;
+import org.quartz.TriggerBuilder;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.test.context.TestPropertySource;
+
+import java.time.Duration;
+import java.time.ZonedDateTime;
+import java.util.Date;
+import java.util.List;
+
+import javax.sql.DataSource;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.search.MeterNotFoundException;
+import io.micrometer.core.instrument.search.RequiredSearch;
+
+import static com.box.l10n.mojito.monitoring.QuartzPendingJobsReportingTask.extractClassName;
+import static com.box.l10n.mojito.quartz.QuartzConfig.DYNAMIC_GROUP_NAME;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+import static org.quartz.JobBuilder.newJob;
+
+public class QuartzPendingJobsReportingTaskTest extends ServiceTestBase {
+
+    public static final String MY_GROUP = "MY_GROUP";
+
+    @Autowired
+    MeterRegistry meterRegistry;
+
+    @Autowired
+    DataSource dataSource;
+
+    @Autowired
+    Scheduler scheduler;
+
+    @Autowired
+    DBUtils dbUtils;
+
+    @Value("${l10n.management.metrics.quartz.sql-queue-monitoring.enabled:false}")
+    Boolean monitoringEnabled;
+
+    QuartzPendingJobsReportingTask task;
+
+    /*
+     * This sets up the condition we'll use for await() later on. Given jobs can take some time
+     * to start and be processed, we wait at most 5 seconds and we ignore exceptions of Meters not
+     * found while we search for them (at t0 we might not have a specific meter given by the
+     * RequiredSearch conditions, but then at t1 the process that populates that
+     * Meter might have completed and populated that search)
+     */
+    ConditionFactory await = await().atMost(Duration.ofSeconds(5))
+                                    .ignoreException(MeterNotFoundException.class);
+
+    @Rule
+    public TestIdWatcher testIdWatcher = new TestIdWatcher();
+
+    @Before
+    public void setUp() throws SchedulerException {
+
+
+        task = new QuartzPendingJobsReportingTask(dataSource, meterRegistry);
+
+        scheduler.clear();
+    }
+
+    @Test
+    public void testGaugesCountPendingJobsForNonDynamicGroup() throws Exception {
+
+        Assume.assumeTrue(dbUtils.isMysql() && monitoringEnabled);
+
+        JobDetail job;
+        TriggerBuilder<Trigger> builder;
+
+        job = newJob(Test1Job.class).withIdentity(testIdWatcher.getEntityName("Test1Job"), MY_GROUP).build();
+        builder = TriggerBuilder.newTrigger()
+                                .forJob(job)
+                                .startAt(Date.from(ZonedDateTime.now().plusMinutes(5).toInstant()));
+        scheduler.scheduleJob(job, builder.build());
+
+        job = newJob(Test2Job.class).withIdentity(testIdWatcher.getEntityName("Test2Job"), MY_GROUP).build();
+        builder = TriggerBuilder.newTrigger()
+                                .forJob(job)
+                                .startAt(Date.from(ZonedDateTime.now().plusMinutes(5).toInstant()));
+        scheduler.scheduleJob(job, builder.build());
+
+        task.reportPendingJobs();
+
+        RequiredSearch search1 = meterRegistry.get("quartz.pending.jobs")
+                                              .tag("jobGroup", MY_GROUP)
+                                              .tag("jobClass", "QuartzPendingJobsReportingTaskTest$Test1Job");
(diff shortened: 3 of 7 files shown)


webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzMetricsReportingJobListener.java
```diff
@@ -0,0 +1,127 @@
...
+    @Override
+    public String getName() {
+        return getClass().getSimpleName();
+    }
+
+    @Override
+    public void jobWasExecuted(JobExecutionContext context, JobExecutionException jobException) {
+        try {
+            registerExecutionTime(context);
+            registerPendingJobs(context);
```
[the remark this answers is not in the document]
comunacho: I agree. I do have mixed feelings as well. As we discussed, the other option here would be to run this method during `jobToBeExecuted`, but at the end of the day, that'll have the exact same effect.

As we discussed, other options would be:

1) Using a `ScheduledTask` to run this query periodically (but we'll need to dedup and probably sanitize metrics being pushed through micrometer)
2) Using a Quartz job to pull these metrics within a scheduled job running in Quartz as well (which can lead to stale metrics or not having metrics at all if quartz is busy processing other jobs)
2.a) On top of this idea, we could play around with job priorities. But with slow & long-running jobs even priorities would create a bottleneck.
2.b) Instead of using priorities, we could create a dedicated scheduler on where we could execute jobs collecting metrics. This would ensure we would collect metrics periodically, but we would still using Quartz to collect metrics.

Another approach would encompass using

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzMetricsReportingJobListener.java
```diff
@@ -0,0 +1,127 @@
...
+        } catch (SchedulerException e){
+            logger.error("An exception occurred when collecting quartz metrics", e);
+        }
+    }
+
+    private void registerPendingJobs(JobExecutionContext context) throws SchedulerException {
+        groupJobsByClass(context.getScheduler().getJobKeys(GroupMatcher.anyJobGroup()))
```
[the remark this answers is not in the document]
comunacho: This runs one query which targets indexed columns. For my setup, this is the prepared statement it runs for that:

`SELECT JOB_NAME, JOB_GROUP FROM QRTZ_JOB_DETAILS WHERE SCHED_NAME = 'TestScheduler' AND JOB_GROUP = ?`

So considering this runs on top of indexed columns, it shouldn't be really expensive.

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzMetricsReportingJobListener.java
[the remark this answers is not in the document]
aurambaj: yes, let's do 1. for now since we don't have that many servers. If we're sending the size queue, we can just do an average.

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzMetricsReportingJobListener.java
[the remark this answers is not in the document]
aurambaj: looks like it is fetching data so I'd say that's not ok. Potentially replacing with a similar `count` query and check the explain plan. But yeah as discussed maybe having a schedule task is better.

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private void reportResults(PendingJob pendingJob){
+        meterRegistry.gauge("quartz.pending.jobs",
+                            Tags.of("jobClass", pendingJob.jobClass, "jobGroup", pendingJob.jobGroup),
+                            pendingJob.count);
+    }
+
+    List<PendingJob> fetchResults(){
+        return jdbcTemplate.query(
+            "SELECT job_class_name, job_group, COUNT(*) FROM QRTZ_JOB_DETAILS d GROUP BY d.job_class_name, d.job_group",
```
aurambaj, line 39: they don't have any abstraction on top of the quartz datasource to get the count? you can only list jobs?

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
```diff
@@ -0,0 +1,64 @@
...
+            (rs, num) -> new PendingJob(extractClassName(rs.getString(1)), rs.getString(2), rs.getLong(3))
+        );
+    }
+
+    private String extractClassName(String input) {
```
aurambaj, line 44: can you have a test for this?

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
```diff
@@ -0,0 +1,64 @@
...
+        String[] parts = input.split("\\.");
+        return parts[parts.length-1];
```
aurambaj, line 46: knit: indentation seems weird, do you apply intellij reformat?

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
```diff
@@ -0,0 +1,64 @@
...
+@ConditionalOnExpression("'${spring.datasource.url:}'.toLowerCase().contains('mysql')")
+public class QuartzPendingJobsReportingTask {
+
+    @Autowired
+    JdbcTemplate jdbcTemplate;
+
+    @Autowired
+    MeterRegistry meterRegistry;
+
+    @Scheduled(fixedRateString = "${l10n.management.metrics.quartz-metrics.execution-rate}")
```
aurambaj, line 26: I suspect we can't use @ConfigurationProperties since it is an annotation?

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
```diff
@@ -0,0 +1,143 @@
...
+import io.micrometer.core.instrument.search.MeterNotFoundException;
+import io.micrometer.core.instrument.search.RequiredSearch;
+
+import static com.box.l10n.mojito.quartz.QuartzConfig.DYNAMIC_GROUP_NAME;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+import static org.quartz.JobBuilder.newJob;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = {Application.class},
```
aurambaj, line 34: can we either piggy back on ServiceTestBase or have a simple test that doesn't bootstrap the whole application? 

I don't know if they changed the implementation since then, but at the time we use to have slow builds because of different test contexts (and/or using dirties context). So unless having a test context that is really limited in scope and quick to start, I'd rather avoid re-introducing that.

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
```diff
@@ -0,0 +1,143 @@
...
+    @Autowired
+    Scheduler scheduler;
+
+    ConditionFactory await;
+    SecureRandom random = new SecureRandom();
+
+    @Before
+    public void setUp() throws Exception {
+        if (!scheduler.isStarted()){
+            scheduler.start();
```
aurambaj, line 54: in what case would this happen?

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
```diff
@@ -0,0 +1,143 @@
...
+        }
+        await = await().atMost(Duration.ofSeconds(5))
+                       .ignoreException(MeterNotFoundException.class);
```
aurambaj, line 57: what is this doing?

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
```diff
@@ -0,0 +1,64 @@
...
+    private void reportResults(PendingJob pendingJob){
+        meterRegistry.gauge("quartz.pending.jobs",
+                            Tags.of("jobClass", pendingJob.jobClass, "jobGroup", pendingJob.jobGroup),
+                            pendingJob.count);
+    }
+
+    List<PendingJob> fetchResults(){
+        return jdbcTemplate.query(
+            "SELECT job_class_name, job_group, COUNT(*) FROM QRTZ_JOB_DETAILS d GROUP BY d.job_class_name, d.job_group",
+            (rs, num) -> new PendingJob(extractClassName(rs.getString(1)), rs.getString(2), rs.getLong(3))
```
aurambaj, line 40: can you have a simple test just for this function?

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
```diff
@@ -0,0 +1,143 @@
...
+                                .startAt(Date.from(ZonedDateTime.now().plusMinutes(5).toInstant()));
+        scheduler.scheduleJob(job, builder.build());
+
+        job = newJob(Test2Job.class).withIdentity("Test2Job", MY_GROUP).build();
+        builder = TriggerBuilder.newTrigger()
+                                .forJob(job)
+                                .startAt(Date.from(ZonedDateTime.now().plusMinutes(5).toInstant()));
+        scheduler.scheduleJob(job, builder.build());
+
+        await().until(() -> scheduler.isStarted());
```
aurambaj, line 78: related to @Before?

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
comunacho: Correct. The spring docs suggest using this format if a property is required.

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
comunacho: Correct

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
comunacho: I had a couple of failed situations when running my tests locally, so to avoid any failures it was better to have this as part of the setup.

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
comunacho: Yeah, its just intellij format

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
comunacho: I've spent some time on this yesterday and there's no abstraction for quartz that we could use. One option is to use the `JobStore` method `getJobKeys` along with `getJobGroupNames` but at the end of the day that'd be the same I had built initially.

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
comunacho: Updated

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
comunacho: Done

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
comunacho: This sets up the condition we'll use for awaitility later on. Given jobs can take some time to start and be processed, we wait at most 5 seconds and we ignore exceptions of Meters not found while we search for them (at t0 we might not have a specific meter given by the `RequiredSearch` conditions, but then at t1 the process that populates that Meter might have completed and populated that search)

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzMetricsReportingJobListener.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+
+import java.util.concurrent.TimeUnit;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+
+@ConditionalOnProperty(value = "l10n.management.metrics.quartz-metrics.enabled", havingValue = "true")
```
aurambaj, line 15: knit: `l10n.management.metrics.quartz-metrics.enabled` --> `l10n.management.metrics.quartz.enabled` , no need to repeat metric, no?

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.List;
+
+import javax.sql.DataSource;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+
+@Component
+@ConditionalOnProperty(value = {"l10n.management.metrics.quartz-metrics.enabled",
```
aurambaj, line 17: aren't those two monitoring system independent? since it is opt-in, should it just check a single flag?

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
aurambaj: strange, mine add space around `-`

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
aurambaj: can you add this comment in the code?

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
aurambaj: sad, we don't get auto-complete on properties with this I beleive.

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
```diff
@@ -0,0 +1,196 @@
...
+    @Test
+    public void testFetchResults() throws Exception {
+
+        JobDetail job;
+        TriggerBuilder<Trigger> builder;
+        int test1Count = random.nextInt(5) + 2;
+        int emptyCount = random.nextInt(5) + 2;
+
+        for (int i = 0; i < test1Count; i++) {
+            job = newJob(Test1Job.class).withIdentity("Test1Job_" + i, DYNAMIC_GROUP_NAME).build();
```
aurambaj, line 150: we've been using 
```
    @Rule
    public TestIdWatcher testIdWatcher = new TestIdWatcher();
```

to isolate test data between run and tests. Should probably used here too?

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
```diff
@@ -0,0 +1,196 @@
...
+        await.untilAsserted(() -> assertThat(test1Gauges.gauge().value()).isEqualTo(test1Count));
+        await.untilAsserted(() -> assertThat(test2Gauges.gauge().value()).isEqualTo(emptyCount));
+    }
+
+    @Test
+    public void testFetchResults() throws Exception {
+
+        JobDetail job;
+        TriggerBuilder<Trigger> builder;
+        int test1Count = random.nextInt(5) + 2;
```
aurambaj, line 146: why random numbers here?

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
```diff
@@ -0,0 +1,196 @@
...
+        assertThat(pendingJobs).extracting("jobClass")
+                               .containsExactlyInAnyOrder("QuartzPendingJobsReportingTaskTest$Test1Job", "EmptyTestJob");
+        assertThat(pendingJobs).extracting("jobGroup")
+                               .contains(DYNAMIC_GROUP_NAME);
+        assertThat(pendingJobs).extracting("count")
+                               .containsExactlyInAnyOrder((long) test1Count, (long) emptyCount);
+    }
+
+    @Test
+    public void testExtractClassName(){
```
aurambaj, line 177: indentation is weird again here.

`task.extractClassName("."))` may fail? i think that why I first asked for a test but then removed my comment because I supposed it is a edge case

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
comunacho: Done

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
comunacho: Done

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
comunacho: Ahh, it was a leftover for when I wasn't acting upon the `QuartzPendingJobsReportingTask` directly. Updated

webapp/src/test/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTaskTest.java
```diff
@@ -0,0 +1,205 @@
...
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.search.MeterNotFoundException;
+import io.micrometer.core.instrument.search.RequiredSearch;
+
+import static com.box.l10n.mojito.quartz.QuartzConfig.DYNAMIC_GROUP_NAME;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+import static org.quartz.JobBuilder.newJob;
+
+@Profile("!disablescheduling")
```
aurambaj, line 39: oh, I wasn't clear, this is not meant to be on the test itself. But it is set on scheduled jobs. The profile is activated during testing so that those job don't run during testing.

aurambaj:
Looks like the [issue with the checks ](https://github.com/box/mojito/issues/591)is happening again @jeeyi 

@comunacho the build seems to be failing: https://travis-ci.org/github/box/mojito/builds/701418962

PR #596, opened 2020-06-20
First pass for adding North American indigenous locales

patcon:
Gitter chat context: https://gitter.im/box/mojito?at=5eed1a307ba3965373b936c1

Hey @aurambaj! No pressure to work with this PR if it's stepping on your toes, but I was eager to give it a shot!

Any thoughts on reviewing this? (I'm _sure_ there are things wrong with this approach, but thought it better to have something to talk over 🙂 )

### DIFF
--- a/docs/_docs/refs/mojito-locales.md
+++ b/docs/_docs/refs/mojito-locales.md
@@ -32,9 +32,12 @@ permalink: /docs/refs/mojito-locales/
 | bn-IN	        | Bengali (India) |
 | bs-BA	        | Bosnian (Bosnia and Herzegovina) |
 | ca-ES	        | Catalan (Spain) |
+| chr-021       | Cherokee (Northern America) |
+| cr-021        | Cree (Northern America) |
 | cs-CZ	        | Czech (Czech Republic) |
 | cy-GB	        | Welsh (United Kingdom) |
 | da-DK	        | Danish (Denmark) |
+| dak-021       | Dakota (Northern America) |
 | de-AT	        | German (Austria) |
 | de-CH	        | German (Switzerland) |
 | de-DE	        | German (Germany) |
@@ -56,7 +59,7 @@ permalink: /docs/refs/mojito-locales/
 | en-US	        | English (United States) |
 | en-ZA	        | English (South Africa) |
 | en-ZW	        | English (Zimbabwe) |
-| en-419	        | Spanish (Latin America) |
+| es-419        | Spanish (Latin America) |
 | es-AR	        | Spanish (Argentina) |
 | es-BO	        | Spanish (Bolivia) |
 | es-CL	        | Spanish (Chile) |
@@ -99,6 +102,7 @@ permalink: /docs/refs/mojito-locales/
 | is-IS	        | Icelandic (Iceland) |
 | it-CH	        | Italian (Switzerland) |
 | it-IT	        | Italian (Italy) |
+| iu-021        | Inuktitut (Northern America) |
 | ja-JP	        | Japanese (Japan) |
 | ka-GE	        | Georgian (Georgia) |
 | kk-KZ	        | Kazakh (Kazakhstan) |
@@ -115,12 +119,15 @@ permalink: /docs/refs/mojito-locales/
 | ms-BN	        | Malay (Brunei Darussalam) |
 | ms-MY	        | Malay (Malaysia) |
 | mt-MT	        | Maltese (Malta) |
+| mus-021       | Creek (Northern America) |
 | nb-NO	        | Norwegian (Norway) |
 | nl-BE	        | Dutch (Belgium) |
 | nl-NL	        | Dutch (Netherlands) |
 | nn-NO	        | Norwegian (Nynorsk) (Norway) |
 | ns-ZA	        | Northern Sotho (South Africa) |
+| nv-003        | Navajo (North America) |
 | ny-MW	        | Nyanja; Chewa; Chichewa (Malawi) |
+| oj-021        | Ojibwa (Northern America) |
 | pa-IN	        | Punjabi (India) |
 | pl-PL	        | Polish (Poland) |
 | ps-AR	        | Pashto (Afghanistan) |
@@ -157,6 +164,7 @@ permalink: /docs/refs/mojito-locales/
 | uz-UZ	        | Uzbek (Uzbekistan) |
 | vi-VN	        | Vietnamese (Viet Nam) |
 | xh-ZA	        | Xhosa (South Africa) |
+| ypk-021       | Yupik (Northern America) |
 | zh-CN	        | Chinese (Simplified) |
 | zh-HK	        | Chinese (Hong Kong) |
 | zh-MO	        | Chinese (Macau) |
--- a/webapp/src/main/resources/db/hsql/data.sql
+++ b/webapp/src/main/resources/db/hsql/data.sql
@@ -152,6 +152,14 @@ insert into locale (id, bcp47_tag) values (781, 'zh-MO');
 insert into locale (id, bcp47_tag) values (785, 'zh-SG'); 
 insert into locale (id, bcp47_tag) values (153, 'zh-TW'); 
 insert into locale (id, bcp47_tag) values (793, 'zu-ZA'); 
+insert into locale (id, bcp47_tag) values (804, 'nv-003');
+insert into locale (id, bcp47_tag) values (805, 'oj-021');
+insert into locale (id, bcp47_tag) values (806, 'iu-021');
+insert into locale (id, bcp47_tag) values (807, 'cr-021');
+insert into locale (id, bcp47_tag) values (808, 'mus-021');
+insert into locale (id, bcp47_tag) values (809, 'dak-021');
+insert into locale (id, bcp47_tag) values (810, 'chr-021');
+insert into locale (id, bcp47_tag) values (811, 'ypk-021');
 
 
 insert into plural_form (id, name) values (1, 'one');
@@ -549,4 +557,12 @@ insert into plural_form_for_locale (locale_id, plural_form_id) values (802, 1);
 insert into plural_form_for_locale (locale_id, plural_form_id) values (802, 5);
 insert into plural_form_for_locale (locale_id, plural_form_id) values (803, 1);
 insert into plural_form_for_locale (locale_id, plural_form_id) values (803, 5);
- 
+insert into plural_form_for_locale (locale_id, plural_form_id) values (804, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (805, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (806, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (807, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (808, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (809, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (810, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (811, 0);
+
--- a/webapp/src/main/resources/db/migration/V52__Add_Native_Locales.sql
+++ b/webapp/src/main/resources/db/migration/V52__Add_Native_Locales.sql
@@ -1,0 +1,17 @@
+insert into locale (id, bcp47_tag) values (804, 'nv-003');
+insert into locale (id, bcp47_tag) values (805, 'oj-021');
+insert into locale (id, bcp47_tag) values (806, 'iu-021');
+insert into locale (id, bcp47_tag) values (807, 'cr-021');
+insert into locale (id, bcp47_tag) values (808, 'mus-021');
+insert into locale (id, bcp47_tag) values (809, 'dak-021');
+insert into locale (id, bcp47_tag) values (810, 'chr-021');
+insert into locale (id, bcp47_tag) values (811, 'ypk-021');
+
+insert into plural_form_for_locale (locale_id, plural_form_id) values (804, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (805, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (806, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (807, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (808, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (809, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (810, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (811, 0);


patcon:
And FYI, I'm attempting to use [UN M49 Standard Country or Area Codes](https://en.wikipedia.org/wiki/UN_M49) "representing geographical (continental and sub-continental) supranational regions" as opposed to nation-state boundaries that may be disputed by the nations speaking the language (e.g. `US`). This is consistent with `es-419`, which also uses one of these codes.

I tried to choose the best of either `003` (North America, including Mexico and Central America) or `021` (North<em>ern</em> America, which excludes Mexico and below), based on where traditional territories seemed to be. A native speaker may be able to offer more input on how best to handle this.

Other options would be to use `000` for "missing data", or `019` for "Americas", or perhaps something odd like `778` for "Transition countries" (though I think that has a specific meaning). Anyhow, the point is that I would advocate to considering whatever native speakers might want to advocate for.

docs/_docs/refs/mojito-locales.md
```diff
@@ -32,9 +32,12 @@ permalink: /docs/refs/mojito-locales/
 | bn-IN	        | Bengali (India) |
 | bs-BA	        | Bosnian (Bosnia and Herzegovina) |
 | ca-ES	        | Catalan (Spain) |
+| chr-021       | Cherokee (Northern America) |
+| cr-021        | Cree (Northern America) |
 | cs-CZ	        | Czech (Czech Republic) |
```
patcon, line 37:
```suggestion
| ca-ES	        | Catalan (Spain) |
| chr-021	        | Cherokee (Northern America) |
| cr-021	        | Cree (Northern America) |
| cs-CZ	        | Czech (Czech Republic) |
```

docs/_docs/refs/mojito-locales.md
```diff
@@ -32,9 +32,12 @@ permalink: /docs/refs/mojito-locales/
...
 | cy-GB	        | Welsh (United Kingdom) |
 | da-DK	        | Danish (Denmark) |
+| dak-021       | Dakota (Northern America) |
 | de-AT	        | German (Austria) |
```
patcon, line 41:
```suggestion
| dak-021	        | Dakota (Northern America) |
| de-AT	        | German (Austria) |
```

docs/_docs/refs/mojito-locales.md
```diff
@@ -56,7 +59,7 @@ permalink: /docs/refs/mojito-locales/
 | en-US	        | English (United States) |
 | en-ZA	        | English (South Africa) |
 | en-ZW	        | English (Zimbabwe) |
-| en-419	        | Spanish (Latin America) |
+| es-419        | Spanish (Latin America) |
 | es-AR	        | Spanish (Argentina) |
```
patcon, line 63:
```suggestion
| es-419	        | Spanish (Latin America) |
| es-AR	        | Spanish (Argentina) |
```

docs/_docs/refs/mojito-locales.md
```diff
@@ -99,6 +102,7 @@ permalink: /docs/refs/mojito-locales/
 | is-IS	        | Icelandic (Iceland) |
 | it-CH	        | Italian (Switzerland) |
 | it-IT	        | Italian (Italy) |
+| iu-021        | Inuktitut (Northern America) |
 | ja-JP	        | Japanese (Japan) |
```
patcon, line 106:
```suggestion
| iu-021	        | Inuktitut (Northern America) |
| ja-JP	        | Japanese (Japan) |
```

docs/_docs/refs/mojito-locales.md
```diff
@@ -115,12 +119,15 @@ permalink: /docs/refs/mojito-locales/
...
 | mt-MT	        | Maltese (Malta) |
+| mus-021       | Creek (Northern America) |
 | nb-NO	        | Norwegian (Norway) |
 | nl-BE	        | Dutch (Belgium) |
 | nl-NL	        | Dutch (Netherlands) |
 | nn-NO	        | Norwegian (Nynorsk) (Norway) |
 | ns-ZA	        | Northern Sotho (South Africa) |
+| nv-003        | Navajo (North America) |
 | ny-MW	        | Nyanja; Chewa; Chichewa (Malawi) |
+| oj-021        | Ojibwa (Northern America) |
```
patcon, line 130:
```suggestion
| nv-003	        | Navajo (North America) |
| ny-MW	        | Nyanja; Chewa; Chichewa (Malawi) |
| oj-021	        | Ojibwa (Northern America) |
```

docs/_docs/refs/mojito-locales.md
```diff
@@ -115,12 +119,15 @@ permalink: /docs/refs/mojito-locales/
 | ms-BN	        | Malay (Brunei Darussalam) |
 | ms-MY	        | Malay (Malaysia) |
 | mt-MT	        | Maltese (Malta) |
+| mus-021       | Creek (Northern America) |
 | nb-NO	        | Norwegian (Norway) |
```
patcon, line 123:
```suggestion
| mus-021	        | Creek (Northern America) |
| nb-NO	        | Norwegian (Norway) |
```

docs/_docs/refs/mojito-locales.md
```diff
@@ -157,6 +164,7 @@ permalink: /docs/refs/mojito-locales/
 | uz-UZ	        | Uzbek (Uzbekistan) |
 | vi-VN	        | Vietnamese (Viet Nam) |
 | xh-ZA	        | Xhosa (South Africa) |
+| ypk-021       | Yupik (Northern America) |
 | zh-CN	        | Chinese (Simplified) |
```
patcon, line 168:
```suggestion
| ypk-021	        | Yupik (Northern America) |
| zh-CN	        | Chinese (Simplified) |
```

docs/_docs/refs/mojito-locales.md
```diff
@@ -56,7 +59,7 @@ permalink: /docs/refs/mojito-locales/
 | en-US	        | English (United States) |
 | en-ZA	        | English (South Africa) |
 | en-ZW	        | English (Zimbabwe) |
-| en-419	        | Spanish (Latin America) |
+| es-419	        | Spanish (Latin America) |
```
aurambaj, line 62: good catch! can be a first commit just fixing that :)

docs/_docs/refs/mojito-locales.md
```diff
@@ -32,9 +32,12 @@ permalink: /docs/refs/mojito-locales/
 | bn-IN	        | Bengali (India) |
 | bs-BA	        | Bosnian (Bosnia and Herzegovina) |
 | ca-ES	        | Catalan (Spain) |
+| chr-021	        | Cherokee (Northern America) |
+| cr-021	        | Cree (Northern America) |
```
aurambaj, line 36: My current thinking is to not have region/territory unless it is a locale that is available in CLDR. I know currently Mojito always have a region but I'm thinking to move away from that pattern.

The new locale list would be any locale defined in: https://github.com/unicode-cldr/cldr-core/blob/master/availableLocales.json + any languages: https://github.com/unicode-cldr/cldr-localenames-full/blob/master/main/en/languages.json. 

So for the few i checked they'd be only languages

thoughts?

webapp/src/main/resources/db/hsql/data.sql
```diff
@@ -549,4 +557,12 @@ insert into plural_form_for_locale (locale_id, plural_form_id) values (802, 1);
 insert into plural_form_for_locale (locale_id, plural_form_id) values (802, 5);
 insert into plural_form_for_locale (locale_id, plural_form_id) values (803, 1);
 insert into plural_form_for_locale (locale_id, plural_form_id) values (803, 5);
- 
\ No newline at end of file
+insert into plural_form_for_locale (locale_id, plural_form_id) values (804, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (805, 0);
+insert into plural_form_for_locale (locale_id, plural_form_id) values (806, 0);
```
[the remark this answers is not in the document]
patcon: (I just left the "zeros" as placeholders until I had time to figure out what these were.)

I can work on this! Might need some time. Can we add this later, and use a sane default for now, or does merging something incorrect lead to hassle?

docs/_docs/refs/mojito-locales.md
```diff
@@ -157,11 +164,11 @@ permalink: /docs/refs/mojito-locales/
 | uz-UZ	        | Uzbek (Uzbekistan) |
 | vi-VN	        | Vietnamese (Viet Nam) |
 | xh-ZA	        | Xhosa (South Africa) |
+| ypk-021	        | Yupik (Northern America) |
```
aurambaj, line 167: yupik seems to be a group of language, https://en.wikipedia.org/wiki/Yupik_languages. `    "ypk": "ems ess esu ynk",` from this list only "esu" has a display name hence was generated from the script I wrote. All other entries you had are covered. 

Wondering if you really need this entry of if "esu" would work in your case.

docs/_docs/refs/mojito-locales.md
patcon: I think that's fair. I'd love to support someone on getting the most appropriate language codes into CLDR itself through formal registration process, rather than doing something opinionated here :)

docs/_docs/refs/mojito-locales.md
```diff
@@ -115,12 +119,15 @@ permalink: /docs/refs/mojito-locales/
 | ms-BN	        | Malay (Brunei Darussalam) |
 | ms-MY	        | Malay (Malaysia) |
 | mt-MT	        | Maltese (Malta) |
+| mus-021	        | Creek (Northern America) |
```
arecvlohe, line 122: Muscogee (Creek). Creek is the historical name, Muscogee (Creek) is the nation

docs/_docs/refs/mojito-locales.md
```diff
@@ -115,12 +119,15 @@ permalink: /docs/refs/mojito-locales/
...
 | nb-NO	        | Norwegian (Norway) |
 | nl-BE	        | Dutch (Belgium) |
 | nl-NL	        | Dutch (Netherlands) |
 | nn-NO	        | Norwegian (Nynorsk) (Norway) |
 | ns-ZA	        | Northern Sotho (South Africa) |
+| nv-003	        | Navajo (North America) |
```
arecvlohe, line 128: Navajo is a name given by the Spanish. Better to go with Diné

docs/_docs/refs/mojito-locales.md
```diff
@@ -115,12 +119,15 @@ permalink: /docs/refs/mojito-locales/
...
 | ny-MW	        | Nyanja; Chewa; Chichewa (Malawi) |
+| oj-021	        | Ojibwa (Northern America) |
```
arecvlohe, line 130: Ojibway is a French term. Anishinaabe is what they call themselves

docs/_docs/refs/mojito-locales.md
arecvlohe: I think Muscogee works fine here

patcon: Honestly, this is the small tedious stuff that I'd be happy to wade through the bureacracy on -- "Creek" is what's in CLDR dataset that's maintained internationally, so it'll keep showing up if it's wrong, since much of the field of localization seems programmatic or at least highly inclined to follow the standards:
https://github.com/unicode-cldr/cldr-localenames-full/blob/993632df2f5d6a2d33cbbf40d922474c2482eaca/main/en-001/languages.json#L390

If you can confirm this is universal that "Muscogee" is the more appropriate term, then I could wrangle with the bureacracy to have that reflected in the standards. We could review the standards and collect a list of stuff like this to push through en masse

aurambaj: @arecvlohe this is great info, thanks for sharing

docs/_docs/refs/mojito-locales.md
patcon: YES. This is great insight. To clarify, this downstream tool isn't the place to fix this, but I'd like to support on this.

Term originates here: https://github.com/unicode-cldr/cldr-localenames-full/blob/993632df2f5d6a2d33cbbf40d922474c2482eaca/main/en-001/languages.json#L426

Current CLDR is [v36 (release notes)](http://cldr.unicode.org/index/downloads/cldr-36), and `nv` was added in March 2018 for [v33 (release notes)](http://cldr.unicode.org/index/downloads/cldr-33) (or some time in the year prior to its official version release)

I can try to dig up the mailing lists to see the conversation when this was added. It may be that this conversation happened already. But even so, receptivity to this sort of feedback may have changed.

docs/_docs/refs/mojito-locales.md
patcon: ditto above

Quick research says this is much more entrenched than other terms. Seems to have originated in 2009 from days when changes happened via internet engineering taskforce (IETF) RFC:
- https://tools.ietf.org/html/draft-ietf-ltru-4645bis-10
- http://cldr.unicode.org/development/development-process/design-proposals/languages-to-show-for-translation

docs/_docs/refs/mojito-locales.md
patcon: Sorry, requires login to UNICODE ticket tracker: https://unicode-org.atlassian.net/browse/CLDR-13814?jql=text%20~%20%22navajo%22

- `CLDR-13814` [Addition of core data and new locale: Navajo (nv)](https://unicode-org.atlassian.net/browse/CLDR-13814?jql=text%20~%20%22navajo%22)
  - lots of missing information in the registration request -- looks like they could use some support
  - Was submitted by Google on May 24, but they just withdrew on June 8th, as they were not able to "get vetters" in time.

I'm not sure I understand, since it seems it's in there already, but perhaps it's not an official locale yet, as it's lacking some deeper level of detail.

patcon:
No hey, not the weeds at all. At some level, this is important. 

This drove me down a rabbithole into the CLDR issue queue. Seems they're about to cut their yearly release. [The process](http://cldr.unicode.org/index/process) seems to involve changes having "vetters", of which "guest vetters" have a single vote count. From what I gather, a proposal for changes/additions requires 8 votes.

https://unicode-org.atlassian.net/

So any changes that you feel should be merged into the standard can perhaps get into the queue with enough non-insiders registering for their system and vouching for any changes.

docs/_docs/refs/mojito-locales.md
patcon: Oh hey, good news: seems someone else chimed in on this in Aug 2009, and it's in the upcoming release: https://unicode-org.atlassian.net/browse/CLDR-13193?jql=text%20~%20%22muskogee%22

docs/_docs/refs/mojito-locales.md
patcon: Nothing related in tracker for:
- ojibwa https://unicode-org.atlassian.net/browse/ICU-3455?jql=text%20~%20%22ojibwa%22
- Anishinaabe https://unicode-org.atlassian.net/browse/CLDR-3797?jql=text%20~%20%22Anishinaabe%22

docs/_docs/refs/mojito-locales.md
patcon: I can imagine they might lean on decisions like this:
https://www.indianz.com/News/2017/04/19/navajo-nation-council-rejects-bill-to-ch.asp

but if there are alternative perspectives that have been legitimized through community channels, it could perhaps be raised... maybe there's precedent for some other approach (e.g., multiple entries)

docs/_docs/refs/mojito-locales.md
arecvlohe: Oh nice! I guess I don't have to do anything then.

docs/_docs/refs/mojito-locales.md
arecvlohe: Thanks for all the information. Let's talk about this in a chat so I can get the proper perspective on it. What I would like to do is send this out through out social media and gather feedback that way. It would be nice to if someone can point me to something more official.

docs/_docs/refs/mojito-locales.md
arecvlohe: Interesting. I will have to read up more on this. Bureaucracy but without engaging Native peoples it seems. Is that the norm?

arecvlohe:
> No hey, not the weeds at all. At some level, this is important. As I understand, the changes are larger than this specific project, which is just parrotting the terms used from the international language standards.
> 
> This drove me down a rabbithole into the CLDR issue queue. Seems they're about to cut their yearly release. [The process](http://cldr.unicode.org/index/process) seems to involve changes having "vetters", of which "guest vetters" have a single vote count. From what I gather, a proposal for changes/additions requires 8 votes.
> 
> https://unicode-org.atlassian.net/
> 
> So any changes that you feel should be merged into the standard can perhaps get into the queue with enough non-insiders registering for their system and vouching for any changes.

We can take this offline but I am not understanding out to be an outside vetter. Let me know on Slack and we can take it from there.

PR #597, opened 2020-06-22
Add new locales using the CLDR dataset

aurambaj:
Mojito's locale list is now composed of all available locales in CLDR + all the languages that have an entry with a display name in CLDR.

The main different compared to previous data set is that now, languages only are also available. Before that, all locales had a region (beside "en" that was the root locale).
For example, you now have "fr" and "fr-FR".

This also update the frontend to be able to render all locale names (including with scripts).


aurambaj:
@patcon this cover most of your list, only "yupik" is missing which seems to be a group of locale.

patcon:
Looks amazing, Jean! But given the number of operations, might be worth consolidating into `BULK INSERT`? On my mind because hosted apps like heroku often only have so many seconds (90 sec for heroku) to bind a port, and startup is already 60-80 second on first go :)

PR #600, opened 2020-06-24
Strip characters from metrics tag name and values

comunacho:
Our `PrefixHierarchicalNameMapper` might require users to cleanup tag names and values on its metrics, stripping out characters that can't be parsed/processed by the backends collecting metrics. This change provides a straight-forward approach to strip unwanted characters from tag names and values through a simple configuration parameter: `l10n.management.metrics.export.statsd.prefix-hierarchical-name-mapper.strip-characters`

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapper.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapper.java
@@ -1,23 +1,34 @@
 package com.box.l10n.mojito.monitoring;
 
-import com.google.common.base.Preconditions;
+
 import com.google.common.base.Strings;
+
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import javax.annotation.Nullable;
+
 import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.config.NamingConvention;
 import io.micrometer.core.instrument.util.HierarchicalNameMapper;
 
-import java.util.stream.Collectors;
+import static com.google.common.base.Preconditions.checkNotNull;
 
 public class PrefixHierarchicalNameMapper implements HierarchicalNameMapper {
 
     String namePrefix;
     String tagKeyPrefix;
+    String stripCharacters;
     NamingConvention nameConventionOverride;
 
-    public PrefixHierarchicalNameMapper(String namePrefix, String tagKeyPrefix, NamingConvention nameConventionOverride) {
-        Preconditions.checkNotNull(namePrefix, tagKeyPrefix, nameConventionOverride);
-        this.namePrefix = namePrefix;
-        this.tagKeyPrefix = tagKeyPrefix;
+    public PrefixHierarchicalNameMapper(String namePrefix,
+                                        String tagKeyPrefix,
+                                        String stripCharacters,
+                                        @Nullable NamingConvention nameConventionOverride) {
+        this.namePrefix = checkNotNull(namePrefix);
+        this.tagKeyPrefix = checkNotNull(tagKeyPrefix);
+        this.stripCharacters = stripCharacters;
         this.nameConventionOverride = nameConventionOverride;
     }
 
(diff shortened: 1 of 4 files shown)


webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapper.java
```diff
@@ -1,45 +1,74 @@
...
 
-    public PrefixHierarchicalNameMapper(String namePrefix, String tagKeyPrefix, NamingConvention nameConventionOverride) {
-        Preconditions.checkNotNull(namePrefix, tagKeyPrefix, nameConventionOverride);
-        this.namePrefix = namePrefix;
-        this.tagKeyPrefix = tagKeyPrefix;
+    public PrefixHierarchicalNameMapper(String namePrefix,
+                                        String tagKeyPrefix,
+                                        String stripCharacters,
+                                        @Nullable NamingConvention nameConventionOverride) {
+        this.namePrefix = checkNotNull(namePrefix);
```
aurambaj, line 29: thanks for fixing that

webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapper.java
```diff
@@ -1,45 +1,74 @@
...
+        this.tagKeyPrefix = checkNotNull(tagKeyPrefix);
+        this.stripCharacters = checkNotNull(stripCharacters);
         this.nameConventionOverride = nameConventionOverride;
     }
 
+    public PrefixHierarchicalNameMapper(String namePrefix,
+                                        String tagKeyPrefix,
```
aurambaj, line 36: why do we need multiple ctor?

webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapper.java
```diff
@@ -1,45 +1,74 @@
...
+    }
+
+    private String formatTag(Tag tag) {
+        return "." + tagKeyPrefix + stripCharacters(tag.getKey()) + "." + stripCharacters(tag.getValue());
+    }
 
-        return hierarchicalName;
+    private String stripCharacters(String input){
+        return Optional.ofNullable(Strings.emptyToNull(stripCharacters))
+                       .map(pattern -> input.replaceAll("[" + pattern + "]", ""))
```
aurambaj, line 71: is pattern character list? small UT for this function?

webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapperStatsdConfigurationProperties.java
```diff
@@ -12,6 +12,7 @@
     String namePrefix = "";
     String tagKeyPrefix = "";
     NamingConvention namingConventionOverride = null;
+    String stripCharacters = "";
```
aurambaj, line 15: should it be null by default to convey the "nothing to be done", can you have a quick javadoc for format to help with adding the property?

webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapper.java
```diff
@@ -1,45 +1,74 @@
...
     /**
      * Pretty much a copy of {@link HierarchicalNameMapper.DEFAULT} with prefixes added.
      */
     @Override
     public String toHierarchicalName(Meter.Id id, NamingConvention namingConvention) {
 
-        if (nameConventionOverride != null) {
-            namingConvention = nameConventionOverride;
-        }
+        namingConvention = Optional.ofNullable(nameConventionOverride).orElse(namingConvention);
```
aurambaj, line 52: is there a real benefit of this notation in this case?

webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapper.java
comunacho: We don't need them, but it makes easier to test 😄

webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapper.java
comunacho: Yeah, it's a character list.

I've added some tests here: https://github.com/box/mojito/pull/600/files#diff-a085e48a4ec35529d79a15b1d6000457R70-R99 they all cover this functionality (along with the tests we already had), but I'll add a couple tests cases for it.

webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapperStatsdConfigurationProperties.java
comunacho: Yeah, it makes sense. I'll add the javadoc.

webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapper.java
aurambaj: yeah I saw that later reading the test. I tend to prefer more verbose test for a cleaner class (if the ctor is not used in the actual code).

There is a cool feature in intellij to add parameters to functions, it just updates everything at once with default value. It kind of alleviate the need of default values in cto

webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapper.java
comunacho: IMHO, it improves readability quite a bit

aurambaj: yeah, readability is a matter of taste and I have the opposite opinion here for example. So it seems to be an unnecessary change for this PR.

webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapper.java
comunacho: Updated

webapp/src/test/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapperTest.java
```diff
@@ -14,53 +13,121 @@
...
+        mapper = new PrefixHierarchicalNameMapper("mojito", "_t_", stripCharacters, NamingConvention.snakeCase);
+        Tags tags = Tags.of("tag#1", "value/{1}/.")
+                        .and("tag#2", "value$2/{test}/value")
+                        .and("tag#3!", "value!3/@test#")
+                        .and("tag#4", "");
+        Meter.Id id = new Meter.Id("metricName", tags, null, null, Meter.Type.COUNTER);
+        String hierarchicalName = mapper.toHierarchicalName(id, NamingConvention.camelCase);
+        logger.debug(hierarchicalName);
+
+        assertEquals("mojito.metricName._t_tag1.value/1/.._t_tag2.value2/test/value._t_tag3.value3/test._t_tag4.", hierarchicalName);
```
aurambaj, line 81: is `..` fine in `ue/1/.._t_`? or could it cause parsing issue too?

webapp/src/main/java/com/box/l10n/mojito/monitoring/PrefixHierarchicalNameMapperStatsdConfigurationProperties.java
```diff
@@ -12,6 +12,7 @@
     String namePrefix = "";
     String tagKeyPrefix = "";
     NamingConvention namingConventionOverride = null;
+    String stripCharacters = null;
```
aurambaj, line 15: i though having some javadoc here so that it shows when typing the properties in intellij for example but minor

PR #607, opened 2020-07-14
Quartz queue monitoring - Job counting fix

comunacho:
The job queue monitoring code we have for Quartz its flawed since it uses a variable reference as the value to be reported in the gauge, but we need a function reference instead to return the exact value at reporting time. Hence, we create a cache with the results obtained each time from the DB and we gather the value for the gauge using that cache instead.

I've added test cases to our integration tests to ensure subsequent runs report queue sizes correctly.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
@@ -1,5 +1,6 @@
 package com.box.l10n.mojito.monitoring;
 
+import io.micrometer.core.instrument.Tags;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.context.annotation.Profile;
@@ -8,11 +9,16 @@ import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Component;
 
 import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import javax.sql.DataSource;
 
 import io.micrometer.core.instrument.MeterRegistry;
-import io.micrometer.core.instrument.Tags;
 
 @Profile("!disablescheduling")
 @Component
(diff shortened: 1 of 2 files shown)


webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
```diff
@@ -21,28 +25,40 @@
...
         this.jdbcTemplate = new JdbcTemplate(dataSource);
+        this.queueSizes = new ConcurrentHashMap<>();
     }
 
     @Scheduled(fixedRateString = "${l10n.management.metrics.quartz.sql-queue-monitoring.execution-rate}")
     public void reportPendingJobs() {
-        fetchResults().forEach(this::reportResults);
+        // First we remove all of our previously stored results (notice line 52). If a job clears from the queue
+        // we would still report it as having pending jobs as it'll retrieve its size from our queueSizes map.
+        queueSizes.clear();
```
aurambaj, line 41: I guess that makes sense for ephemeral jobs but isn't that clearing the gauge for some jobs that shouldn't be cleared?

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
```diff
@@ -21,28 +25,40 @@
...
     }
 
-    private void reportResults(PendingJob pendingJob){
-        meterRegistry.gauge("quartz.pending.jobs",
-                            Tags.of("jobClass", pendingJob.jobClass, "jobGroup", pendingJob.jobGroup),
-                            pendingJob.count);
+    private void registerResults(PendingJob pendingJob) {
+
+        // For each pending job we store the value we got from the db regardless of its current value in the map
+        queueSizes.put(pendingJob, new AtomicLong(pendingJob.count));
```
aurambaj, line 48: why don't we persist the gauge and then update their value? https://micrometer.io/docs/concepts#_manually_incrementingdecrementing_a_gauge

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
comunacho: Sure, look at line 48 where we reinsert values into the map all over again. And I've covered that same test case you mention in the tests. 

Also, if we were to remove just those that aren't returning results during a particular measurement, we'll need to maintain a record of previously obtained results to compare. Given this is returning 20 rows from the table approximately, is just easier to clear the map.

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
comunacho: We'd still need to maintain a list of previously generated values. At the end of the day is the same issue we're solving here with a map/cache along with a supplier being passed to the gauge builder.

aurambaj: I'm not sure to understand why we need previously generate values? we just need to assign the new values to the gauges, no?

comunacho: Let me use an example:

- At t0, we call `fetchResults()` and we get two results, job1 having 3 pending tasks, job2 having 1 pending task. We report those values to the registry.

- At t1, we get one result, job1 having 2 pending tasks. Pending tasks for job2 are cleared, therefore our `List<PendingJob>` won't have any items from job2 within its results, so we won't update its count to 0 and rather we'll keep it as 1 (the value reported at t0). But since we do `queueSize.clear()` and the we use `() -> queueSizes.getOrDefault(pendingJob, new AtomicLong(0))` as the supplier that will generate values for that gauge, we ensure the value goes back to 0 for that specific gauge (the one we use for job2).

With the direct `Number` reference we lose track of gauges that yielded results previously and in the current measurement, do not return any result.

aurambaj: I'm thinking of pseudo code similar to this:

```
Map gauges = ....
results = fetchResults()
gaugesWithoutResult = ...
gaugeWithResult = ...
resultWithoutGauge = ....
for (gauge :  gaugesWIthoutResult) gauge.set(0)
for (gauge :  gaugesWIthResult) gauge.set(result.getValue())
for (result :  resultWithoutGauge) createGauge()
```

gaugesWithoutResult and gaugeWithResult could also be one and just have a getValue() that either return 0 if there is no entry in `results` or the actual value from `results`

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
```diff
@@ -21,28 +25,49 @@
...
+        return meterRegistry.gauge("quartz.pending.jobs",
+                Tags.of("jobClass", pendingJob.jobClass, "jobGroup", pendingJob.jobGroup),
+                new AtomicLong(pendingJob.count));
     }
 
-    List<PendingJob> fetchResults(){
+    private void updateQueueSizes(List<PendingJob> pendingJobs) {
+        queueSizes.forEach((key, val) -> {
+            Long size = pendingJobs.stream()
+                    .filter(pj -> pj.key.equals(key))
```
aurambaj, line 58: use a Map instead of a list for `List<PendingJob>` to have constant search time instead of iterating on the list?

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
```diff
@@ -21,28 +25,49 @@
...
+                    .findFirst()
+                    .map(pj -> pj.count)
+                    .orElse(0L);
+            // If the list of yielded results doesn't contains the pendingjob, then we update its value to be zero
+            queueSizes.get(key).set(size);
+        });
+    }
+
+    List<PendingJob> fetchResults() {
```
aurambaj, line 67: could return a map here, see previous comment for the search

webapp/src/main/java/com/box/l10n/mojito/monitoring/QuartzPendingJobsReportingTask.java
```diff
@@ -52,18 +77,19 @@ static String extractClassName(String input) {
...
         public String jobGroup;
-        public Long   count;
+        public String key;
+        public Long count;
 
         public PendingJob(String jobClass, String jobGroup, Long count) {
             this.jobClass = jobClass;
             this.jobGroup = jobGroup;
             this.count = count;
+            this.key = jobClass + "-" + jobGroup;
```
aurambaj, line 92: should just be a getter actually? there is no need for a field, is there?

PR #609, opened 2020-07-23
Android Strings File - Reader & Writer

comunacho:
Replacing `AndroidStringsXmlReader` and `AndroidStringsXmlWriter` with new classes that serialize and deserialize Android Strings files to a POJO schema. 

The old implementation mixed `TextUnitDTO` and `Document` concerns, making it very cumbersome to use. I created a hierarchy to generate `AndroidStringDocument` from XML files and backward, which can then be used and mapped independently.

Differently from the result yielded by `AndroidStringsXmlReader`, the `AndroidStringDocument` doesn't provide the `pluralForm`, `pluralFormOther`, etc. methods, as they are `TextUnitDTO` concerns. An upcoming change will provide the mapping from `AndroidStringDocument` to `TextUnitDTO` and backward.


comunacho:
Build is 🟢  https://travis-ci.org/github/box/mojito/builds/711183160

PR #610, opened 2020-07-23
Android Strings file reader & writer

comunacho:
Replacing `AndroidStringsXmlReader` and `AndroidStringsXmlWriter` with new classes that serialize and deserialize Android Strings files to a POJO schema. 

The old implementation mixed `TextUnitDTO` and `Document` concerns, making it very cumbersome to use. I created a hierarchy to generate `AndroidStringDocument` from XML files and backward, which can then be used and mapped independently.

Differently from the result yielded by `AndroidStringsXmlReader`, the `AndroidStringDocument` doesn't provide the `pluralForm`, `pluralFormOther`, etc. methods, as they are `TextUnitDTO` concerns. An upcoming change will provide the mapping from `AndroidStringDocument` to `TextUnitDTO` and backward.


webapp/src/main/java/com/box/l10n/mojito/android/strings/AndroidStringDocument.java
```diff
@@ -0,0 +1,65 @@
...
+import org.w3c.dom.Node;
+
+import java.util.LinkedList;
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+public class AndroidStringDocument {
+
+    private List<AndroidSingular> singulars = new LinkedList<>();
```
aurambaj, line 12: will this prevent customized order, i mean intertwine singular and plurals?

webapp/src/main/java/com/box/l10n/mojito/android/strings/AndroidStringDocumentWriter.java
```diff
@@ -0,0 +1,145 @@
...
+        this.source = requireNonNull(source);
+        buildDomSource();
+    }
+
+    public void buildDomSource() throws ParserConfigurationException {
+        document = documentBuilder().newDocument();
+        root = document.createElement(ROOT_ELEMENT_NAME);
+        document.setXmlStandalone(true);
+        document.appendChild(root);
+        source.getSingulars().forEach(this::addSingular);
```
aurambaj, line 48: I think that confirm my previous feeling. For our specific use case we split files singular vs plural but maybe here it make sense to have an implementation a little bit more flexible that respect the order in which the strings are added

webapp/src/main/java/com/box/l10n/mojito/android/strings/AndroidStringDocumentReader.java
```diff
@@ -0,0 +1,49 @@
...
+        Queue<Node> commentNodes = new LinkedList<>();
+        AndroidStringDocument document = new AndroidStringDocument();
+        Node node;
+
+        for (int i = 0; i < source.getDocumentElement().getChildNodes().getLength(); i++) {
+
+            node = source.getDocumentElement().getChildNodes().item(i);
+
+            if (Node.COMMENT_NODE == node.getNodeType()) {
+
```
aurambaj, line 38: knit: remove extra empty lines?

webapp/src/main/java/com/box/l10n/mojito/android/strings/AndroidStringDocumentUtils.java
```diff
@@ -0,0 +1,20 @@
+package com.box.l10n.mojito.android.strings;
+
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.TransformerFactory;
+
+public abstract class AndroidStringDocumentUtils {
```
aurambaj, line 8: why abstract class? doesn't seem to be extended anywhere. if you want to prevent instantiation, just a private constructor will do + final class?

webapp/src/main/java/com/box/l10n/mojito/android/strings/AndroidStringDocument.java
comunacho: Updated

webapp/src/main/java/com/box/l10n/mojito/android/strings/AndroidStringDocumentUtils.java
comunacho: Updated

webapp/src/main/java/com/box/l10n/mojito/android/strings/AndroidStringDocumentWriter.java
comunacho: Updated

comunacho:
@aurambaj Updated the code per your comments/suggestions. I do agree we should move this to the commons module when we start using it.

aurambaj:
🎉

Issue #612, opened 2020-07-25
JCommander trims argument values from the command line


comunacho:
The current version of JCommander that we're using has a bug that trims arguments values from the command line: https://github.com/cbeust/jcommander/issues/417

The most recent version of it (1.78) has a fix for this bug: https://github.com/cbeust/jcommander/blob/master/src/main/java/com/beust/jcommander/JCommander.java#L912

comunacho:
@aurambaj This can be assigned to me

aurambaj:
Did you have any issue because of that? 

Anyway, upgrading to new version is always good if it is quick.

PR #614, opened 2020-07-27
ThirdPartySyncCommand - Adding new parameters

comunacho:
To implement the `PULL`, `PUSH` and `PUSH_TRANSLATIONS` commands in `ThirdPartyService`, we need to update `ThirdPartySyncCommand` and several dependencies to utilize these new parameters: `--skip-text-units-with-pattern` and `--skip-assets-path-pattern`.

Please note the TODO added to `ThirdPartySyncCommandTest` and documented in #612

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/ThirdPartySyncActionsConverter.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/ThirdPartySyncActionsConverter.java
@@ -1,15 +1,15 @@
 package com.box.l10n.mojito.cli.command;
 
-import com.box.l10n.mojito.rest.client.ThirdPartySync;
+import com.box.l10n.mojito.rest.ThirdPartySyncAction;
 
 /**
  *
  * @author sdemyanenko
  */
-public class ThirdPartySyncActionsConverter extends EnumConverter<ThirdPartySync.Action> {
+public class ThirdPartySyncActionsConverter extends EnumConverter<ThirdPartySyncAction> {
 
     @Override
-    protected Class<ThirdPartySync.Action> getGenericClass() {
-        return ThirdPartySync.Action.class;
+    protected Class<ThirdPartySyncAction> getGenericClass() {
+        return ThirdPartySyncAction.class;
     }
 }
--- a/cli/src/test/java/com/box/l10n/mojito/cli/command/ThirdPartySyncCommandTest.java
+++ b/cli/src/test/java/com/box/l10n/mojito/cli/command/ThirdPartySyncCommandTest.java
@@ -1,32 +1,105 @@
 package com.box.l10n.mojito.cli.command;
 
 import com.box.l10n.mojito.cli.CLITestBase;
+import com.box.l10n.mojito.cli.utils.PollableTaskJobMatcher;
+import com.box.l10n.mojito.cli.utils.TestingJobListener;
 import com.box.l10n.mojito.entity.Repository;
-import com.box.l10n.mojito.rest.client.ThirdPartySync;
+import com.box.l10n.mojito.json.ObjectMapper;
+import com.box.l10n.mojito.rest.thirdparty.ThirdPartySyncAction;
+import com.box.l10n.mojito.service.thirdparty.ThirdPartySyncJob;
+import com.box.l10n.mojito.service.thirdparty.ThirdPartySyncJobInput;
+import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.quartz.JobKey;
+import org.quartz.Matcher;
+import org.quartz.Scheduler;
+import org.quartz.SchedulerException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 
 import java.util.Arrays;
+import java.util.List;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static com.box.l10n.mojito.rest.thirdparty.ThirdPartySyncAction.MAP_TEXTUNIT;
+import static com.box.l10n.mojito.rest.thirdparty.ThirdPartySyncAction.PUSH_SCREENSHOT;
+import static org.assertj.core.api.Assertions.assertThat;
 
 public class ThirdPartySyncCommandTest extends CLITestBase {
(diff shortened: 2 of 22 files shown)


webapp/pom.xml
```diff
@@ -240,7 +241,6 @@
             <groupId>com.box.l10n.mojito</groupId>
             <artifactId>mojito-restclient</artifactId>
             <version>0.111-SNAPSHOT</version>
-            <scope>test</scope>
```
comunacho, line 243: Removing the test scope to use `ThirdPartySync` in `ThirdPartyWS` and `ThirdPartyService`.

cli/src/test/java/com/box/l10n/mojito/cli/command/ThirdPartySyncCommandTest.java
```diff
@@ -1,32 +1,95 @@
...
+    Scheduler scheduler;
+
+    Matcher<JobKey> jobMatcher;
+    TestingJobListener testingJobListener;
+
+    @Before
+    public void setUp() throws SchedulerException {
+        testingJobListener = new TestingJobListener(objectMapper);
+        jobMatcher = new PollableTaskJobMatcher<>(ThirdPartySyncJob.class);
+        scheduler.getListenerManager().addJobListener(testingJobListener, jobMatcher);
```
comunacho, line 42: We add a `JobListener` that applies solely for jobs of the `ThirdPartySyncJob` class

cli/src/test/java/com/box/l10n/mojito/cli/command/ThirdPartySyncCommandTest.java
```diff
@@ -1,32 +1,95 @@
...
+    }
+
+    @After
+    public void tearDown() throws SchedulerException {
+        scheduler.getListenerManager().removeJobListener(testingJobListener.getName());
+        scheduler.getListenerManager().removeJobListenerMatcher(testingJobListener.getName(), jobMatcher);
```
comunacho, line 48: At the end of the test we remove the `TestingJobListener` that matches `ThirdPartySyncJob` quartz jobs.

comunacho:
@aurambaj Created this new PR to merge into `spring2x` branch.

cli/src/main/java/com/box/l10n/mojito/cli/command/ThirdPartySyncActionsConverter.java
```diff
@@ -1,15 +1,15 @@
 package com.box.l10n.mojito.cli.command;
 
-import com.box.l10n.mojito.rest.client.ThirdPartySync;
+import com.box.l10n.mojito.rest.ThirdPartySyncAction;
```
aurambaj, line 3: oh sorry, maybe I wasn't super clear last time we chatted about it. I don't want the rest client and the webservice layer to be linked closely. The main reason is that in most cases there is not a 1:1 mapping between what's done in the server and what makes sense in the client (hibernate entities, slightly different object). I know that in many case the POJO are or file like duplicates between the client and server but I'd rather keep that way. Let's not have the rest client as dependency of webapp, it also becomes messy and we may end up having some code going through the client instead of directly through the service which doesn't make sense. It may become confusing.

comunacho: Just updated the code

cli/src/test/java/com/box/l10n/mojito/cli/command/ThirdPartySyncCommandTest.java
```diff
@@ -1,32 +1,96 @@
...
+                "-sa", skipAssetPattern,
+                "-o", options.get(0), options.get(1));
+
+        waitForCondition("Ensure ThirdPartySyncJob gets executed",
+                () -> testingJobListener.getExecuted().size() > 0);
+
+        ThirdPartySyncJobInput jobInput = testingJobListener.getFirstInputMapAs(ThirdPartySyncJobInput.class);
 
-        String outputString = outputCapture.toString();
-        assertTrue(outputString.contains(Arrays.asList(ThirdPartySync.Action.MAP_TEXTUNIT, ThirdPartySync.Action.PUSH_SCREENSHOT).toString()));
```
aurambaj, line 29: why not keeping that?

cli/src/test/java/com/box/l10n/mojito/cli/command/ThirdPartySyncCommandTest.java
```diff
@@ -1,32 +1,96 @@
...
+        getL10nJCommander().run("thirdparty-sync",
+                "-r", repository.getName(),
+                "-p", projectId,
+                "-a", MAP_TEXTUNIT.name(), PUSH_SCREENSHOT.name(),
+                "-ps", pluralSeparator,
+                "-st", skipTextUnitPattern,
+                "-sa", skipAssetPattern,
+                "-o", options.get(0), options.get(1));
+
+        waitForCondition("Ensure ThirdPartySyncJob gets executed",
```
aurambaj, line 81: I guess that listener is ok for now though I think changing the application context is not too great if we ever start running test in parallel, etc. 

So my preference would be avoiding the listener and just wait for the state (or different part of the state) to get to what we expect which would indicate a successful migration.

On top of my head I don't remember other test intercepting part of the request to perform input validation. In general those are more like end to end test. The parameters are not check directly but instead the results are checked inferring that the proper arguments were passed

comunacho: I've prioritized on testing that the parameters given to the CLI were passed around downstream all the way to the Quartz job for this test. I do agree that here is better to test final states or side effects, but since non of the parameters is being used in `ThirdPartyTMSInMemory` method we have nowadays and we're not triggering side effects, my preference for this PR was to have the argument passing fully tested (since that's what we're changing now).

Since we'll still implement new methods in `ThirdPartyTMS` in the upcoming PRs, I can add end to end testing through the `ThirdPartyTMSInMemory` class once we have these new method added to the interface signature as well as to the interface implementors (`ThirdPartyTMSSmartling` and `ThirdPartyTMSInMemory`)

cli/src/test/java/com/box/l10n/mojito/cli/command/ThirdPartySyncCommandTest.java
comunacho: Updated

PR #615, opened 2020-07-27
Updating JCommander to 1.78

### LINKED ISSUE #612: JCommander trims argument values from the command line

comunacho:
Fixes #612 

There's a known issue with `jCommander.setAcceptUnknownOptions(true)`, described [here](https://github.com/cbeust/jcommander/issues/377). Disabling that option and updating tests that were using unknown parameters helped to get a green build.

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/L10nJCommander.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/L10nJCommander.java
@@ -194,7 +194,7 @@ public class L10nJCommander {
      */
     public void usage(String commandName) {
         StringBuilder stringBuilder = new StringBuilder();
-        jCommander.usage(commandName, stringBuilder, "");
+        jCommander.getUsageFormatter().usage(commandName, stringBuilder, "");
         consoleWriter.a(stringBuilder).println();
     }
 
@@ -206,7 +206,8 @@ public class L10nJCommander {
         logger.debug("Create JCommander instance");
         jCommander = new JCommander();
 
-        jCommander.setAcceptUnknownOptions(true);
+        // TODO: Cant enable this until https://github.com/cbeust/jcommander/issues/377 is fixed
+        // jCommander.setAcceptUnknownOptions(true);
 
         logger.debug("Initialize the JCommander instance");
         jCommander.setProgramName(PROGRAM_NAME);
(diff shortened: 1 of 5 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/L10nJCommander.java
```diff
@@ -206,7 +206,8 @@ public void createJCommanderForRun() {
         logger.debug("Create JCommander instance");
         jCommander = new JCommander();
 
-        jCommander.setAcceptUnknownOptions(true);
+        // TODO: Cant enable this until https://github.com/cbeust/jcommander/issues/377 is fixed
+        // jCommander.setAcceptUnknownOptions(true);
```
aurambaj, line 210: I think that was used to pass extra configuration to spring ... not sure if it is actually used today

cli/src/test/java/com/box/l10n/mojito/cli/command/ImportLocalizedAssetCommandTest.java
```diff
@@ -345,7 +345,6 @@ public void importXcodeXliff() throws Exception {
 
         getL10nJCommander().run("push", "-r", repository.getName(),
                 "-s", getInputResourcesTestDir("source").getAbsolutePath(),
-                "-t", getInputResourcesTestDir("translations").getAbsolutePath(),
```
comunacho, line 348: The Push command doesn't have any `-t` parameter

cli/src/test/java/com/box/l10n/mojito/cli/command/DropImportCommandTest.java
```diff
@@ -119,7 +119,7 @@ public Long answer(InvocationOnMock invocation) throws Throwable {
 
         localizeDropFiles(dropRepository.findById(dropId).orElse(null));
 
-        l10nJCommander.run(new String[]{"drop-import", "-r", repository.getName(), "--number-drop-fetched", "1000"});
```
aurambaj, line 122: 😨

cli/src/main/java/com/box/l10n/mojito/cli/command/L10nJCommander.java
aurambaj: @jeeyi Is this used on your side?

PR #616, opened 2020-07-28
Adding MDC parameters to QuartzPollableJob and ThirdPartyService

comunacho:
This change adds MDC parameters to be shown in the log events appended by the changed class loggers. This will allow easy discoverability when debugging. This is an example of the log event produced by the `ThirdPartyService`:

```json
{
  "@timestamp": "2020-07-28T21:18:41.781Z",
  "@version": "1",
  "message": "thirdparty TMS: thirdPartyTMSMock",
  "logger_name": "com.box.l10n.mojito.service.thirdparty.ThirdPartyService",
  "thread_name": "main",
  "level": "DEBUG",
  "level_value": 10000,
  "thirdPartyProjectId": "someProjectIdForTest",
  "repositoryId": "123"
}
```

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyService.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyService.java
@@ -26,6 +26,7 @@ import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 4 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyService.java
```diff
@@ -126,6 +131,9 @@ void syncMojitoWithThirdPartyTMS(Long repositoryId,
         if (actions.contains(ThirdPartySyncAction.PUSH_SCREENSHOT)) {
             uploadScreenshotsAndCreateMappings(repository, thirdPartyProjectId);
         }
+
+        MDC.remove(REPOSITORY_ID);
+        MDC.remove(THIRD_PARTY_PROJECT_ID);
```
aurambaj, line 136: i'm not a fan of having to remove manually here. an Aspect would be great for that use case

aurambaj: or the `putCloseable`?

PR #619, opened 2020-07-29
New search criterions for TextUnitSearcher

comunacho:
Adding new search criterions on the `TextUnitSearcherParameters` class to allow filtering through the `tm_text_unit.name` and the `asset.path` columns. Both of them will compare using a `lowercase(:column_name:) NOT LIKE 'pattern'` to filter down either text units or assets matching the pattern being passed to `TextUnitSearcher`.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/tm/search/TextUnitSearcher.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/tm/search/TextUnitSearcher.java
@@ -1,14 +1,15 @@
 package com.box.l10n.mojito.service.tm.search;
 
-import com.box.l10n.mojito.nativecriteria.NativeContainsExp;
-import com.box.l10n.mojito.nativecriteria.NativeILikeExp;
-import com.box.l10n.mojito.nativecriteria.NativeColumnEqExp;
 import com.box.l10n.mojito.entity.TMTextUnitVariant;
 import com.box.l10n.mojito.nativecriteria.JpaQueryProvider;
+import com.box.l10n.mojito.nativecriteria.NativeColumnEqExp;
+import com.box.l10n.mojito.nativecriteria.NativeContainsExp;
 import com.box.l10n.mojito.nativecriteria.NativeDateGteExp;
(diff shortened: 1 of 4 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/tm/search/TextUnitSearcher.java
```diff
@@ -284,6 +287,14 @@ NativeCriteria getCriteriaForSearch(TextUnitSearcherParameters searchParameters)
             conjunction.add(new NativeEqExpFix("atu.branch_id", searchParameters.getBranchId()));
         }
 
+        if (!Strings.isNullOrEmpty(searchParameters.getSkipTextUnitWithPattern())) {
```
aurambaj, line 290: why special case for the empty string? you may want every text unit that have a non empty name. It the param is present/not null then just add the filter?

webapp/src/main/java/com/box/l10n/mojito/service/tm/search/TextUnitSearcher.java
```diff
@@ -284,6 +287,14 @@ NativeCriteria getCriteriaForSearch(TextUnitSearcherParameters searchParameters)
...
+            conjunction.add(new NativeNotILikeExp("tu.name", searchParameters.getSkipTextUnitWithPattern()));
+        }
+
+        if (!Strings.isNullOrEmpty(searchParameters.getSkipAssetPathWithPattern())) {
```
aurambaj, line 294: same?

webapp/src/main/java/com/box/l10n/mojito/service/tm/search/TextUnitSearcher.java
comunacho: Updated

webapp/src/main/java/com/box/l10n/mojito/service/tm/search/TextUnitSearcher.java
comunacho: Updated

aurambaj:
looks good

PR #625, opened 2020-08-05
Push command for thirdparty sync

comunacho:
Implementing the `PUSH` action for the `ThirdPartySyncCommand`. A couple of TODOs were added here which I'll be addressing in the upcoming PRs of this series of `ThirdPartyService` related changes.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMS.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMS.java
@@ -36,4 +36,22 @@ interface ThirdPartyTMS {
      * @param thirdPartyImageToTextUnits the list of mappings
      */
     void createImageToTextUnitMappings(String projectId, List<ThirdPartyImageToTextUnit> thirdPartyImageToTextUnits);
+
+    /**
+     * Push source strings to the third party TMS
+     *
+     * @param repository                The Mojito repository
+     * @param projectId                 The third party project Id
+     * @param pluralSeparator           The plural separator to be used in file generation
+     * @param skipTextUnitsWithPattern  The pattern to skip when searching text units
+     * @param skipAssetsWithPathPattern The path pattern to skip when searching assets
+     * @param options                   The options being passed to the third party API
+     * @return The list of files generated in the operation
+     */
+    void push(Repository repository,
+              String projectId,
+              String pluralSeparator,
+              String skipTextUnitsWithPattern,
+              String skipAssetsWithPathPattern,
+              List<String> options);
 }
--- a/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
@@ -1,6 +1,16 @@
 package com.box.l10n.mojito.service.thirdparty;
 
+import com.box.l10n.mojito.android.strings.AndroidStringDocumentMapper;
+import com.box.l10n.mojito.android.strings.AndroidStringDocumentWriter;
 import com.box.l10n.mojito.entity.Repository;
+import com.box.l10n.mojito.service.thirdparty.smartling.SmartlingFile;
+import com.box.l10n.mojito.service.thirdparty.smartling.SmartlingFileUtils;
+import com.box.l10n.mojito.service.thirdparty.smartling.SmartlingOptions;
+import com.box.l10n.mojito.service.tm.search.SearchType;
+import com.box.l10n.mojito.service.tm.search.StatusFilter;
+import com.box.l10n.mojito.service.tm.search.TextUnitDTO;
+import com.box.l10n.mojito.service.tm.search.TextUnitSearcher;
+import com.box.l10n.mojito.service.tm.search.TextUnitSearcherParameters;
 import com.box.l10n.mojito.smartling.AssetPathAndTextUnitNameKeys;
 import com.box.l10n.mojito.smartling.SmartlingClient;
 import com.box.l10n.mojito.smartling.request.Binding;
@@ -8,16 +18,25 @@ import com.box.l10n.mojito.smartling.request.Bindings;
 import com.box.l10n.mojito.smartling.response.ContextUpload;
 import com.box.l10n.mojito.smartling.response.File;
 import com.box.l10n.mojito.smartling.response.StringInfo;
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Streams;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Component;
 
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.TransformerException;
 import java.util.List;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+import static com.box.l10n.mojito.service.thirdparty.smartling.SmartlingFileUtils.isPluralFile;
 
 @ConditionalOnProperty(value = "l10n.ThirdPartyTMS.impl", havingValue = "ThirdPartyTMSSmartling")
 @Component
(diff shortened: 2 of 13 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/tm/TextUnitBatchOperator.java
```diff
@@ -0,0 +1,138 @@
...
+        TextUnitListIterator pluralsIterator = new TextUnitListIterator(searcher, batchSize, repositoryId,
+                localesTags, skipTextUnitsWithPattern, skipAssetsWithPathPattern, false, false, "%");
+
+        batch = 0;
+        while (pluralsIterator.hasNext()) {
+            consumerPlurals.accept(pluralsIterator.next(), batch++);
+        }
+    }
+
+    private static class TextUnitListIterator implements Iterator<List<TextUnitDTO>> {
```
aurambaj, line 62: have you seen this class: `PageFetcherOffsetAndLimitSplitIterator`, it is used by the [Smartling impl](https://github.com/box/mojito/blob/0036896dd6710bf17fd38b03fa5f8fe98d6129b9/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java#L55) already, [drill down in the client](https://github.com/box/mojito/blob/41f4005831a77219fa0f519db3e7ae8f72017ece/webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java#L85) . it is useful for streaming data like that. 

I suspect having a stream will allow to remove `TextUnitBatchOperator`?

webapp/src/test/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartlingITest.java
```diff
@@ -0,0 +1,118 @@
...
+        thirdPartyService.mapMojitoAndThirdPartyTextUnits(repository, testConfig.projectId);
+
+        logger.debug("Second mapping");
+        thirdPartyService.mapMojitoAndThirdPartyTextUnits(repository, testConfig.projectId);
+
+        thirdPartyService.uploadScreenshotsAndCreateMappings(repository, testConfig.projectId);
+    }
+
+    @Test
+    public void testGetFilePattern() {
```
aurambaj, line 111: that should be in the UT, no?

webapp/src/test/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartlingITest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import java.util.regex.Pattern;
+
+/**
+ * Integration test that needs Smartling account and verfication on the screenshot upload done manually in Smartling
+ */
+public class ThirdPartyTMSSmartlingITest extends ServiceTestBase {
```
aurambaj, line 30: yes!

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMS.java
```diff
@@ -36,4 +36,22 @@
      * @param thirdPartyImageToTextUnits the list of mappings
      */
     void createImageToTextUnitMappings(String projectId, List<ThirdPartyImageToTextUnit> thirdPartyImageToTextUnits);
+
+    /**
+     * Push source files to the third party TMS
```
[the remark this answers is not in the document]
comunacho: Updated it to be `void` and changed the internal implementation to maintain a list of files.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -91,7 +122,65 @@ public void createImageToTextUnitMappings(String projectId, List<ThirdPartyImage
...
+            }
+
+            consumer.accept(file);
+
+        } catch (ParserConfigurationException | TransformerException e) {
+            logger.error("An error ocurred when processing a push batch", e);
+        }
+    }
+
+    static String getOutputSourceFile(int batchNumber, String repositoryName, String prefix) {
```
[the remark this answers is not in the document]
comunacho: Not at all, I've moved this to a separate class

webapp/src/test/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartlingITest.java
comunacho: Updated

webapp/src/main/java/com/box/l10n/mojito/service/tm/TextUnitBatchOperator.java
comunacho: I did took a quick peek at it, but considering we need a stateful iteration (we need to keep a reference to the `batchNumber` variable on both "consumers") and that we could lose the inherent streaming capabilities by forcing statefulness into a stateless stream, I moved to a classic Iterator to keep it simple.

aurambaj: also thinking more about it, fetching the text unit DTOs by batch might not make a lot of sense here. Before it was done like that because of the http calls but that's gone now. So just batching when creating the file would simpler (and eventually long term this should rewritten in a more scalable way)

comunacho: Updated

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -1,35 +1,81 @@
...
+    private final SmartlingClient smartlingClient;
+    private final AssetPathAndTextUnitNameKeys assetPathAndTextUnitNameKeys;
+    private final TextUnitSearcher textUnitSearcher; // TODO - Add a cached version of textunitsearcher
+    private final Integer batchSize;
+
+    private List<SmartlingFile> lastPushResult;
 
     @Autowired
-    AssetPathAndTextUnitNameKeys assetPathAndTextUnitNameKeys;
+    public ThirdPartyTMSSmartling(@Autowired SmartlingClient smartlingClient,
```
aurambaj, line 64: isn't `@Autowired` only needed on the ctor?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -91,11 +137,106 @@ public void createImageToTextUnitMappings(String projectId, List<ThirdPartyImage
...
+                     String skipTextUnitsWithPattern,
+                     String skipAssetsWithPathPattern,
+                     List<String> optionList) {
+
+        lastPushResult = new ArrayList<>();
+        List<String> localeTags = ImmutableList.of("en");
+        SmartlingOptions options = SmartlingOptions.parseList(optionList);
+        AndroidStringDocumentMapper mapper = new AndroidStringDocumentMapper(pluralSeparator, null);
+
+        AtomicInteger batchCount = new AtomicInteger(0);
```
aurambaj, line 153: to associate the batch number:  https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Streams.html#mapWithIndex-java.util.stream.Stream-com.google.common.collect.Streams.FunctionWithIndex- ?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -91,11 +137,106 @@ public void createImageToTextUnitMappings(String projectId, List<ThirdPartyImage
...
+        file.setFileName(SmartlingUtils.getOutputSourceFile(batchNumber, repository.getName(), filePrefix));
+
+        try {
+
+            logger.debug("Save source file to: {}", file.getFileName());
+            AndroidStringDocumentWriter writer = new AndroidStringDocumentWriter(mapper.readFromSourceTextUnits(result));
+            file.setFileContent(writer.toText());
+
+        } catch (ParserConfigurationException | TransformerException e) {
+            logger.error("An error ocurred when processing a push batch", e);
```
aurambaj, line 185: re-throw an exception here?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -91,11 +137,106 @@ public void createImageToTextUnitMappings(String projectId, List<ThirdPartyImage
...
+        }
+
+        if (!options.isDryRun()) {
```
aurambaj, line 188: didn't you want to save the file somewhere in case of a dryrun?

comunacho: That's the plan, I'll work on that separately for all three actions at once.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingUtils.java
```diff
@@ -0,0 +1,26 @@
+package com.box.l10n.mojito.service.thirdparty.smartling;
+
+import java.util.regex.Pattern;
+
+public final class SmartlingUtils {
```
aurambaj, line 5: s/SmartlingUtils/SmartlingFiles/ ? or something along those lines

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
comunacho: Ah good catch, I forgot to remove the annotation from the constructor parameters.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
comunacho: 💯

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
aurambaj: btw what are those exeception, just standard XML issues? looks like we won't ever do much about it... should it moved to unchecked exception in the AndroidStringDocumentWriter

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
aurambaj: ok

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -91,11 +136,104 @@ public void createImageToTextUnitMappings(String projectId, List<ThirdPartyImage
...
+        TextUnitSearcherParameters result = new TextUnitSearcherParameters();
+        result.setRepositoryIds(repositoryId);
+        result.setLocaleTags(localesTags);
+        result.setRootLocaleExcluded(false);
+        result.setDoNotTranslateFilter(false);
+        result.setSearchType(SearchType.ILIKE);
+        result.setStatusFilter(StatusFilter.TRANSLATED);
+        result.setPluralFormsFiltered(pluralFormsFiltered);
+        result.setPluralFormsExcluded(pluralFormsExcluded);
+        result.setSkipTextUnitWithPattern(skipTextUnitsWithPattern);
```
aurambaj, line 225: just realizing that the text unit searcher could have been kept more generic by adding a search type: "not like" and search on name or path.

if so, could you clean that up in a following PR, ideally we should try to not have to much boilerplate in the searcher, it has already grown quite a bit.

PR #630, opened 2020-08-06
Pull command for ThirdPartySync

comunacho:
Implementing the `PULL` action for the `ThirdPartySyncCommand`

### DIFF
--- a/common/src/main/java/com/box/l10n/mojito/LocaleMappingHelper.java
+++ b/common/src/main/java/com/box/l10n/mojito/LocaleMappingHelper.java
@@ -1,0 +1,46 @@
+package com.box.l10n.mojito;
+
+import com.google.common.base.Splitter;
+import com.google.common.collect.HashBiMap;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+
+@Component
+public class LocaleMappingHelper {
+
+    /**
+     * Gets the locale mapping given the locale mapping param
+     *
+     * @param localeMapppingParam locale mapping param coming from the CLI
+     * @return A map containing the locale mapping
+     */
+    public Map<String, String> getLocaleMapping(String localeMapppingParam) {
+
+        Map<String, String> localeMappings = null;
+        if (localeMapppingParam != null) {
+            localeMappings = Splitter.on(",").withKeyValueSeparator(":").split(localeMapppingParam);
+        }
+
+        return localeMappings;
+    }
+
+    /**
+     * Gets the inverse locale mapping given the locale mapping param
+     *
+     * @param localeMapppingParam locale mapping param coming from the CLI
+     * @return A map containing the inverse locale mapping
+     */
+    public Map<String, String> getInverseLocaleMapping(String localeMapppingParam) {
+
+        Map<String, String> inverseLocaleMapping = null;
+
+        if (localeMapppingParam != null) {
+            inverseLocaleMapping = HashBiMap.create(getLocaleMapping(localeMapppingParam)).inverse();
+
+        }
+
+        return inverseLocaleMapping;
+    }
+
+}
--- a/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
@@ -1,11 +1,15 @@
 package com.box.l10n.mojito.service.thirdparty;
 
 import com.box.l10n.mojito.android.strings.AndroidStringDocumentMapper;
+import com.box.l10n.mojito.android.strings.AndroidStringDocumentReader;
 import com.box.l10n.mojito.android.strings.AndroidStringDocumentWriter;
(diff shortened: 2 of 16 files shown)


common/src/main/java/com/box/l10n/mojito/LocaleMappingHelper.java
```diff
@@ -0,0 +1,48 @@
+package com.box.l10n.mojito;
+
+import com.google.common.base.Splitter;
+import com.google.common.collect.HashBiMap;
+
+import java.util.Map;
+
+public final class LocaleMappingHelper {
```
aurambaj, line 8: knit: is this refactoring needed in this commit? my guess is that it should just change the webapp. 

Also I see you're moving things into static helper (vs spring component). Overall in the project we've been mainly using components (even for small utils). I don't have a bulletproof definition on when a component should be used but for consistency sake I think we should keep using spring component. I'm happy to have a chat if you have a strong argument for one or the other.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -190,6 +206,70 @@ SmartlingFile processPushBatch(List<TextUnitDTO> result,
...
+                     String pluralSeparator,
+                     Map<String, String> localeMapping,
+                     String skipTextUnitsWithPattern,
+                     String skipAssetsWithPathPattern,
+                     List<String> optionList) {
+
+        SmartlingOptions options = SmartlingOptions.parseList(optionList);
+        AndroidStringDocumentMapper mapper = new AndroidStringDocumentMapper(pluralSeparator, null);
+
+        Long singulars = countSingulars(repository.getId(), EN_LOCALE, skipTextUnitsWithPattern, skipAssetsWithPathPattern);
```
aurambaj, line 221: knit: singularCount?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -232,6 +328,10 @@ private TextUnitSearcherParameters baseParams(Long repositoryId,
         return result;
     }
 
+    long batchesFor(long total) {
```
aurambaj, line 331: not clear at first sight what that is about. Rename? (or document?), test?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingPluralFix.java
```diff
@@ -0,0 +1,34 @@
...
+
+import static com.box.l10n.mojito.android.strings.AndroidPluralQuantity.OTHER;
+
+public final class SmartlingPluralFix {
+
+    private SmartlingPluralFix() {
+        throw new AssertionError("Do not instantiate");
+    }
+
+    public static List<TextUnitDTO> fixTextUnits(List<TextUnitDTO> textUnits) {
```
aurambaj, line 18: on top of my head there were some locales explicitly mentioned in the code for which the fix should be done. I don't see it here, any logic behind that?

common/src/main/java/com/box/l10n/mojito/LocaleMappingHelper.java
comunacho: We need it as we don't have access to CLI code in the webapp, so it was just easier to have these two methods in the common module.

I've updated the class for it to be a spring component as well.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
comunacho: Updated

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
comunacho: Updated with tests.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingPluralFix.java
comunacho: The list of locales to be plural fixed will be passed through the command line, with the `smartling-plural-fix` option. E.g: `mojito thirdparty-sync -r repo -p projectId -o "smartling-pluralfix=pt-BR,pt-PT"`, so if we receive locales in the command line, we apply the fix for them, otherwise, nothing is done.

common/src/main/java/com/box/l10n/mojito/LocaleMappingHelper.java
aurambaj: oh yeah, I tried to check if the file was moved into common and couldn't see it ... I need new 👓,  my bad.

PR #631, opened 2020-08-10
Push translations command for thirdparty sync

comunacho:
Implementing the `PUSH_TRANSLATIONS` action for the `ThirdPartySyncCommand`

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
@@ -15,6 +15,7 @@ import com.box.l10n.mojito.service.tm.search.StatusFilter;
 import com.box.l10n.mojito.service.tm.search.TextUnitDTO;
 import com.box.l10n.mojito.service.tm.search.TextUnitSearcher;
 import com.box.l10n.mojito.service.tm.search.TextUnitSearcherParameters;
+import com.box.l10n.mojito.service.tm.search.UsedFilter;
 import com.box.l10n.mojito.smartling.AssetPathAndTextUnitNameKeys;
 import com.box.l10n.mojito.smartling.SmartlingClient;
 import com.box.l10n.mojito.smartling.request.Binding;
@@ -25,7 +26,6 @@ import com.box.l10n.mojito.smartling.response.StringInfo;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
-import com.google.common.collect.Streams;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -38,14 +38,19 @@ import javax.xml.transform.TransformerException;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.LongStream;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
+import static com.box.l10n.mojito.android.strings.AndroidPluralQuantity.MANY;
 import static com.box.l10n.mojito.service.thirdparty.smartling.SmartlingFileUtils.getOutputSourceFile;
+import static com.box.l10n.mojito.service.thirdparty.smartling.SmartlingFileUtils.getOutputTargetFile;
 import static com.box.l10n.mojito.service.thirdparty.smartling.SmartlingFileUtils.isPluralFile;
+import static com.google.common.collect.Streams.mapWithIndex;
 
 @ConditionalOnProperty(value = "l10n.ThirdPartyTMS.impl", havingValue = "ThirdPartyTMSSmartling")
 @Component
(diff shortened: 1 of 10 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -270,52 +273,145 @@ private void processPullBatch(Long batchNumber,
...
+                                 String skipTextUnitsWithPattern,
+                                 String skipAssetsWithPathPattern,
+                                 List<String> optionList) {
+
+        SmartlingOptions options = SmartlingOptions.parseList(optionList);
+        AndroidStringDocumentMapper mapper = new AndroidStringDocumentMapper(pluralSeparator, null);
+
+        repository.getRepositoryLocales()
+                  .stream()
+                  .filter(RepositoryLocale::isToBeFullyTranslated)
```
aurambaj, line 290: was this in the original script? If so i'm not sure why that's the case: importing locales that are not meant to be fully translated still makes sense (that flag is mostly used to say that's ok to have some string not translated for that locale but it doesn't mean the translation are not good)

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -60,15 +66,16 @@
...
 
     private final SmartlingClient smartlingClient;
     private final AssetPathAndTextUnitNameKeys assetPathAndTextUnitNameKeys;
     private final TextUnitSearcher textUnitSearcher; // TODO - Add a cached version of textunitsearcher
     private final TextUnitBatchImporterService textUnitBatchImporterService;
     private final Integer batchSize;
 
-    private List<SmartlingFile> lastPushResult;
+    private List<SmartlingFile> lastPushResult = new ArrayList<>();
+    private List<SmartlingFile> lastPushTranslationResult = new ArrayList<>();
```
aurambaj, line 78: actually why do we maintain state in the spring component? how is this working with multiple concurrent calls?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
comunacho: Updated.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
comunacho: I was planning to have this resolved as part of the upcoming file upload PR. But I'll share my ideas here first to know your thoughts:

- One option is to remove the Component annotations from the `ThirdPartyTMS` implementors and create a new Component, say `ThirdPartyTMSFactory` to build instances of the `ThirdPartyTMS` implementors (through the factory) in the `ThirdPartyService` each time we call `syncMojitoWithThirdPartyTMS`, moving the conditionals we already have to this factory class (This would be my preferred approach)

- The second option is to add a `requestId` parameter to each of these methods and store the results based off that `requestId` in a Map. Obviously, the problem there is that we'll be storing a lot of data for single use so it's not a scalable solution.

Let me know what you think.

aurambaj: if stated can't be moved easily into the functions you can use @Configurable instead of @Component + creating new instance with `new` (similar to option 1)

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -60,15 +66,16 @@
     file in smartling.
     */
     private static final int DEFAULT_BATCH_SIZE = 5004;
-    private static final List<String> EN_LOCALE = ImmutableList.of("en");
+    private static final String LOCALE_EN = "en";
 
     private final SmartlingClient smartlingClient;
     private final AssetPathAndTextUnitNameKeys assetPathAndTextUnitNameKeys;
     private final TextUnitSearcher textUnitSearcher; // TODO - Add a cached version of textunitsearcher
```
aurambaj, line 73: did you identify perf issue with this? I don't think we use a cache anywhere else for the searcher, and I don't remember it being an issue.

comunacho: Oh yeah, I'll remove this because we won't need it.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingOptions.java
```diff
@@ -41,12 +45,14 @@ public static SmartlingOptions parseList(List<String> options) {
...
         String placeholderFormat = map.getOrDefault(PLACEHOLDER_FORMAT, null);
         String customPlaceholderFormat = map.getOrDefault(PLACEHOLDER_FORMAT_CUSTOM, null);
+        String requestId = map.getOrDefault(REQUEST_ID, null);
 
         return new SmartlingOptions(
                 pluralFixStr.isEmpty() ? Collections.emptySet() : ImmutableSet.copyOf(pluralFixStr.split(",")),
                 placeholderFormat,
                 customPlaceholderFormat,
-                "true".equalsIgnoreCase(dryRunStr));
+                "true".equalsIgnoreCase(dryRunStr),
```
aurambaj, line 54: `parseBoolean()`?

PR #632, opened 2020-08-11
ThirdPartySync - Upload to S3 for dry runs

comunacho:
Last piece of the `ThirdPartySync` migration for Smartling. Allowing the `SmartlingResultProcessor` to upload files when the `dry-run` option is set to true in the command line. Files will go into S3 and we'll log the generated S3 URL to download it if required via the AWS CLI through `aws s3 cp <file_s3_url> .`


webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
```diff
@@ -0,0 +1,105 @@
...
+
+@ConditionalOnProperty(value = "l10n.ThirdPartyTMS.impl", havingValue = "ThirdPartyTMSSmartling")
+@Component
+public class SmartlingResultProcessor {
+
+    static Logger logger = LoggerFactory.getLogger(SmartlingResultProcessor.class);
+
+    private static final String NAME_FORMAT = "%s_%s_%s.zip";
+
+    @Autowired
```
aurambaj, line 32: `@Autowired(required=false)`

but to make it more generic you should use `StructuredBlobStorage`, it will use whatever blob storage is configured and specify a prefix (that how we avoid collision at a system level)

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
```diff
@@ -0,0 +1,105 @@
...
+
+    private String processAction(List<SmartlingFile> files,
+                                 SmartlingOptions options,
+                                 String action) {
+        String result = null;
+
+        if (!files.isEmpty() && options.isDryRun() && s3BlobStorage != null){
+            try {
+                result = uploadFiles(files, options.getRequestId(), action);
+                logger.info("{} result for request id {} uploaded to: {}", action, options.getRequestId(), result);
```
aurambaj, line 53: debug?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
```diff
@@ -0,0 +1,105 @@
...
+            } catch (IOException | SdkClientException e) {
+                logger.error(String.format("An error ocurred when uploading a %s result zip file", action), e);
```
aurambaj, line 55: should the exception be re-thrown? what happens if it fails?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+
+        return result;
+    }
+
+    String uploadFiles(List<SmartlingFile> files, String requestId, String type) throws IOException {
+        checkArgument(!files.isEmpty(), "Smartling file list cant be empty");
```
aurambaj, line 63: why does it matter? what happens if we sync an empty project?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
```diff
@@ -0,0 +1,105 @@
...
+
+        String name = String.format(NAME_FORMAT, requestId, type, ZonedDateTime.now().toEpochSecond());
+        Path zipFile = zipFiles(files, name);
+        byte[] content = Files.readAllBytes(zipFile);
+        s3BlobStorage.put(name, content, Retention.MIN_1_DAY);
+        return s3BlobStorage.getS3Url(name);
+    }
+
+    Path zipFiles(List<SmartlingFile> files, String name) throws IOException {
+        checkArgument(!files.isEmpty(), "Smartling file list cant be empty");
```
aurambaj, line 73: should it be an empty zip, or a file that says no text unit?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
```diff
@@ -0,0 +1,105 @@
...
+public class SmartlingResultProcessor {
+
+    static Logger logger = LoggerFactory.getLogger(SmartlingResultProcessor.class);
+
+    private static final String NAME_FORMAT = "%s_%s_%s.zip";
+
+    @Autowired
+    S3BlobStorage s3BlobStorage;
+
+    public String processPush(List<SmartlingFile> files,
```
aurambaj, line 35: probably won't need to return s3url in those functions when using `StructuredBlobStorage`

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
comunacho: I've used `S3BlobStorage` considering this is meant to be uploaded to S3, not through the generic interface.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
comunacho: Doesn't seems like a good option. We'll need to turn on/off debug logging just for this purpose, and this one line is required to get the s3 URL to be downloaded.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
comunacho: That's why I just logged the exception. Considering this a side-effect, if there's a failure when uploading to S3 doesn't mean we should error out the entire sync process.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
comunacho: Updated

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
comunacho: Updated

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
comunacho: This is specifically meant to be uploaded to S3, not through the generic interface.

aurambaj: ah to make the download easy? 

So I guess the dry run option won't be useful unless using S3. Probably fine since this is mostly for testing I guess.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
aurambaj: we're trying (might not be always consistent) to use debug for repeatable messages and info should be more for one off, like framework initialization. 

Not sure to understand how that system would work? tailing the log would require to have access to the host. it is impractical, no?

Can the URL be guess or displayed in the CLI output?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
aurambaj: well you're doing a dry run, so it shouldn't actually sync and just produce files for testing. Looks like that won't be possible so it should fail hard. 

This won't be done for normal sync will it?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
comunacho: Ultimately that's the goal, to use S3 as the storage for testing purposes and nothing else.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
comunacho: That's a good point. Updated.

And no, we won't be doing for normal sync, just when doing a dry run.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
comunacho: We can obtain this log event by tailing the log file and grepping the `requestId` string, or through Kibana when enabled. And we can guess the URL but eventually, that URL shouldn't be guessed and rather logged so we can get it easily when running tests with a dry run.

aurambaj: Most likely there is some kind of lag in the log propagation and having a predictable URL seem like better option to me. Up to you, since this is mostly for testing

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
aurambaj: ok, i was thinking about people wanting to use that without s3 but yeah not needed at this point since it is mostly to test the implementation

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingResultProcessor.java
aurambaj: can you also pass the error message in the runtimeexception along with the cause?

comunacho: Done

PR #633, opened 2020-08-12
Maven frontend plugin update

comunacho:
Enabling `npx npm-force-resolutions` to be run from maven in case the `preinstall` phase for `npm` fails.


aurambaj:
can you update the commit message with the error and why you're doing that change? I'm pretty we won't remember after some time

PR #637, opened 2020-08-17
Adding missing attributes in for ThirdPartySync PULL

comunacho:
The `AndroidStringDocumentMapper` instance we are using in the `ThirdPartySync` PULL action is missing some attributes.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
@@ -223,24 +223,25 @@ public class ThirdPartyTMSSmartling implements ThirdPartyTMS {
                      List<String> optionList) {
 
         SmartlingOptions options = SmartlingOptions.parseList(optionList);
-        AndroidStringDocumentMapper mapper = new AndroidStringDocumentMapper(pluralSeparator, null);
 
         Long singulars = singularCount(repository.getId(), LOCALE_EN, skipTextUnitsWithPattern, skipAssetsWithPathPattern);
         LongStream.range(0, batchesFor(singulars))
-                  .forEach(num -> processPullBatch(num, mapper, repository, projectId, options, localeMapping, Prefix.SINGULAR));
+                  .forEach(num -> processPullBatch(num, pluralSeparator, repository, projectId, options, localeMapping, Prefix.SINGULAR));
 
         Long plurals = pluralCount(repository.getId(), LOCALE_EN, skipTextUnitsWithPattern, skipAssetsWithPathPattern);
         LongStream.range(0, batchesFor(plurals))
-                  .forEach(num -> processPullBatch(num, mapper, repository, projectId, options, localeMapping, Prefix.PLURAL));
+                  .forEach(num -> processPullBatch(num, pluralSeparator, repository, projectId, options, localeMapping, Prefix.PLURAL));
     }
 
     private void processPullBatch(Long batchNumber,
-                                  AndroidStringDocumentMapper mapper,
+                                  String pluralSeparator,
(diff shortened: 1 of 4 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingPluralFix.java
```diff
@@ -25,6 +26,19 @@ private SmartlingPluralFix() {
                     tu.setPluralForm("many");
                     tu.setPluralFormOther(textUnit.getPluralFormOther());
                     tu.setTarget(textUnit.getTarget());
+
+                    if (!Strings.isNullOrEmpty(textUnit.getRepositoryName())){
```
aurambaj, line 30: why do you need to test for null or empty for those 3?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -223,25 +223,26 @@ public void pull(Repository repository,
                      List<String> optionList) {
 
         SmartlingOptions options = SmartlingOptions.parseList(optionList);
-        AndroidStringDocumentMapper mapper = new AndroidStringDocumentMapper(pluralSeparator, null);
```
aurambaj, line 226: can matching ctor be removed?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingPluralFix.java
comunacho: I just follow this pattern with strings that can be empty or null, but also there's no point in doing the assignation if is either empty or null, right?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
comunacho: We can't, since its being used for other use cases.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/smartling/SmartlingPluralFix.java
aurambaj: everywhere else for pojo mapping we don't do that, so let's keep it consistent with other places. Side notes those (repository name, locale, path)shouldn't be empty because it doesn't make sense though code wise there are little checks.

comunacho: Updated

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
aurambaj: not sure in what case we don't have the repository name and locale name, even i'm confused why it is fine not having it. in fact relooking at the class I'm not sure why that class is not a stateless component?

comunacho: When we do `List<TextUnitDTO>` to `AndroidStringDocument` we don't need the repository or targetLocale, we need it the other way around. But the class isn't stateful, each call to `AndroidStringDocumentMapper.mapToTextUnits` is self-contained.

aurambaj: right, so it looks like a component would make more sense, and pass the required parameters to each function would make it more clearer what is needed.

PR #643, opened 2020-08-24
Handle duplicated string names when mapping TextUnitDTO to AndroidStringDocument

comunacho:
If string names are duplicated when generating an `AndroidStringDocument` from a List of `TextUnitDTO`, the mapper fails, but we should account the `assetPath` to generate the key where we cache the results during conversion.

Also, this fixes an NPE that could occur when no `pluralSeparator` is passed to the `ThirdPartyService` methods.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/android/strings/AndroidStringDocumentMapper.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/android/strings/AndroidStringDocumentMapper.java
@@ -48,7 +48,7 @@ public class AndroidStringDocumentMapper {
             if (isSingularTextUnit(textUnit)) {
                 document.addSingular(textUnitToAndroidSingular(textUnit, useSource));
             } else {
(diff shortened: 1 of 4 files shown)


webapp/src/main/java/com/box/l10n/mojito/android/strings/AndroidStringDocumentMapper.java
```diff
@@ -165,6 +165,10 @@ boolean isSingularTextUnit(TextUnitDTO textUnit) {
         return Strings.isNullOrEmpty(textUnit.getPluralForm());
     }
 
+    String textUnitKey(TextUnitDTO textUnit){
+        return textUnit.getAssetPath() + "_" + textUnit.getPluralFormOther();
```
aurambaj, line 169: don't we have an asset delimiter like `"#@#` ? for singular?

comunacho: Updated

PR #645, opened 2020-09-01
Increasing Smartling auth token expiration delay

comunacho:
The delay we have currently (1 sec) might not be able to catch all scenarios where we would try to use an already expired token (E.g: Smartling grants a token at t0. Then, at t1 (t0 + 2 seconds), we receive that token and save it as valid from t1, but given it was granted at t0, once we're about to reach the grant expiry we could run into authorization errors.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClientConfiguration.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClientConfiguration.java
@@ -12,14 +12,12 @@ import org.springframework.http.client.ClientHttpResponse;
 import org.springframework.security.oauth2.client.DefaultOAuth2ClientContext;
 import org.springframework.security.oauth2.client.OAuth2RestTemplate;
 import org.springframework.security.oauth2.client.resource.OAuth2ProtectedResourceDetails;
-import org.springframework.security.oauth2.client.token.AccessTokenProviderChain;
 import org.springframework.security.oauth2.config.annotation.web.configuration.EnableOAuth2Client;
 import org.springframework.web.client.DefaultResponseErrorHandler;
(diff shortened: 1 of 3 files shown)


webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClientConfiguration.java
```diff
@@ -62,14 +60,10 @@ public OAuth2RestTemplate smartlingRestTemplate() {
...
-                new SmartlingAuthorizationCodeAccessTokenProvider())
-        );
-        oAuth2RestTemplate.setAccessTokenProvider(accessTokenProviderChain);
-
-        DefaultUriTemplateHandler defaultUriTemplateHandler = new DefaultUriTemplateHandler();
-        defaultUriTemplateHandler.setBaseUrl(baseUri);
+        oAuth2RestTemplate.setAccessTokenProvider(new SmartlingAuthorizationCodeAccessTokenProvider());
+        oAuth2RestTemplate.setRetryBadAccessTokens(true);
 
+        DefaultUriBuilderFactory defaultUriTemplateHandler = new DefaultUriBuilderFactory(baseUri);
```
comunacho, line 66: These are drop-in replacements for classes that got deprecated during the Spring boot 2 migration

webapp/src/test/java/com/box/l10n/mojito/smartling/SmartlingClientTest.java
```diff
@@ -88,6 +89,48 @@ public void testGetSourceStringsStream() {
         });
     }
 
+    @Ignore
```
comunacho, line 92: Marking these tests as ignored to avoid delaying the test suite, but all tests were run using this code

PR #647, opened 2020-09-15
Update Translations

boxmoji:


aurambaj:
@ehoogerbeets This PR comes from a  job that automatically publishes translations. I doesn't look like Mojito is still actively translated so can this job be disabled on your side? then this PR won't be re-open.

ehoogerbeets:
old and crufty

PR #680, opened 2021-04-30
Address unit test failure for: BranchDeleteCommandTest (fixes #679)

### LINKED ISSUE #679: BranchDeleteCommandTest unit test failure on clone from master
After cloning and running `mvn test` on Ubuntu 18.04.5, I was getting the following unit test failure:

>2021-04-29 01:21:03.731  WARN 4962 --- [o-auto-1-exec-4] .w.s.m.s.DefaultHandlerExceptionResolver : Resolved [org.springframework.web.method.annotation.MethodArgumentTypeMismatchException: Failed to convert value of type 'java.lang.String' to required type 'org.joda.time.DateTime'; nested exception is org.springframework.core.convert.ConversionFailedException: Failed to convert from type [java.lang.String] to type [@org.springframework.web.bind.annotation.RequestParam org.joda.time.DateTime] for value '2021-04-22T01:21:03.727 01:00'; nested exception is java.lang.IllegalArgumentException: Invalid format: "2021-04-22T01:21:03.727 01:00" is malformed at " 01:00"]
>
>400 : [{"timestamp":1619655663738,"status":400,"error":"Bad Request","message":"","path":"/api/repositories/50/branches"}]

The call in question targetted the following URI (BranchDeleteCommandTest.java#delete:76->checkBranches() ... ): 
>http://localhost:35869/api/repositories/1/branches?deleted=false&createdBefore=2021-04-22T21:24:56.579+01:00

Upon further investigation, it seems that when building the URI for the API call, the AuthenticatedRestTemplate.java isn't encoding in any way the '+' character. Spring boot decodes the '+' character as a space ' ' - which causes an exception when trying to convert the string to a DateTime.
"2021-04-22T21:24:56.579+01:00" is a valid ISO 8601 time format, while "2021-04-22T21:24:56.579 01:00" is not, causing the ConversionFailedException exception.

I'm assuming this UT failure only surfaced for me due to certain mix of locale, time zone (IST) and date display configuration that is set up on my box. 

Relevant discussions:
https://github.com/spring-projects/spring-framework/issues/21399
https://github.com/spring-projects/spring-framework/issues/14464

Environment

```
:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.5 LTS
Release:	18.04
Codename:	bionic
```

```
:~$ java -version
openjdk version "1.8.0_282"
OpenJDK Runtime Environment (build 1.8.0_282-8u282-b08-0ubuntu1~18.04-b08)
OpenJDK 64-Bit Server VM (build 25.282-b08, mixed mode)
```

```
:~$ timedatectl
[...]
                       Time zone: Europe/Dublin (IST, +0100)
       System clock synchronized: yes
systemd-timesyncd.service active: yes
                 RTC in local TZ: no
[...]
````

### PR DESCRIPTION
**Pull request by geoarion**
Addresses: #679 by ensuring the '+' character is encoded to '%2B' in the Rest template only for DateTime objects, so that Spring Boot decodes on the server as '+' instead of space ' '.

The encoding changes are only done for dates and the order of the parameters is preserved.

### DIFF
--- a/restclient/src/main/java/com/box/l10n/mojito/rest/client/RepositoryClient.java
+++ b/restclient/src/main/java/com/box/l10n/mojito/rest/client/RepositoryClient.java
@@ -209,7 +209,7 @@ public class RepositoryClient extends BaseClient {
     public List<Branch> getBranches(Long repositoryId, String branchName, String branchNameRegex,
                                     Boolean deleted, Boolean translated, boolean includeNullBranch,
                                     DateTime createdBefore) {
-        Map<String, String> filterParams = new HashMap<>();
+        Map<String, Object> filterParams = new HashMap<>();
 
         if (branchName != null) {
             filterParams.put("name", branchName);
@@ -224,7 +224,7 @@ public class RepositoryClient extends BaseClient {
         }
 
         if (createdBefore != null) {
-            filterParams.put("createdBefore", createdBefore.toString());
+            filterParams.put("createdBefore", createdBefore);
         }
 
         List<Branch> branches = authenticatedRestTemplate.getForObjectAsListWithQueryStringParams(
--- a/restclient/src/main/java/com/box/l10n/mojito/rest/resttemplate/AuthenticatedRestTemplate.java
+++ b/restclient/src/main/java/com/box/l10n/mojito/rest/resttemplate/AuthenticatedRestTemplate.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
+import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -29,7 +30,10 @@ import org.springframework.web.util.UriComponentsBuilder;
 
 import javax.annotation.PostConstruct;
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
 import java.net.URI;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -191,7 +195,29 @@ public class AuthenticatedRestTemplate {
             uriBuilder.queryParam(entry.getKey(), entry.getValue());
         }
 
-        return uriBuilder.build(false).toUri();
+        // Encode illegal characters in the URI
+        URI uri = uriBuilder.build(false).toUri();
+
+        // Encode DateTime objects in place fully, to ensure the '+' timezone separator doesn't get decoded as a space on the API side
+        uriBuilder = UriComponentsBuilder.fromUri(uri);
+
+        for (Map.Entry<String, ?> entry : queryStringParams.entrySet()) {
+            if (entry.getValue() instanceof DateTime)
+            {
+                String encodedDatetime = null;
+                try {
+                    encodedDatetime = URLEncoder.encode(entry.getValue().toString(), StandardCharsets.UTF_8.toString());
+                } catch (UnsupportedEncodingException ex) {
+                    throw new RuntimeException(ex);
+                }
+
+                uriBuilder.replaceQueryParam(entry.getKey(), encodedDatetime);
+            }
+        }
+
+        // Disable any further encoding on the builder to prevent double encoding
+        uri = uriBuilder.build(true).toUri();
+        return uri;
     }
 
     /**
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: restclient/src/main/java/com/box/l10n/mojito/rest/client/RepositoryClient.java [hunk 1]
```diff
@@ -224,7 +224,7 @@ public void updateRepository(String name, String newName, String description, Bo
         }
 
         if (createdBefore != null) {
-            filterParams.put("createdBefore", createdBefore.toString());
+            filterParams.put("createdBefore", createdBefore);
```
aurambaj, line 227: One solution might be to prevent having the `+` by using some other representation like a timestamp: `filterParams.put("createdBefore", String.valueOf(createdBefore.getMillis()))`;

restclient/src/main/java/com/box/l10n/mojito/rest/resttemplate/AuthenticatedRestTemplate.java
```diff
@@ -191,7 +195,29 @@ protected URI getURIForResourceAndQueryStringParams(String resourcePath, Map<Str
...
 
-        return uriBuilder.build(false).toUri();
+        // Encode illegal characters in the URI
+        URI uri = uriBuilder.build(false).toUri();
+
+        // Encode DateTime objects in place fully, to ensure the '+' timezone separator doesn't get decoded as a space on the API side
+        uriBuilder = UriComponentsBuilder.fromUri(uri);
+
+        for (Map.Entry<String, ?> entry : queryStringParams.entrySet()) {
+            if (entry.getValue() instanceof DateTime)
```
aurambaj, line 205: Could this issue be larger than just dates? reading quickly it is not obvious to me at which level we should handle the encoding and for what types.

aurambaj (commented):
Hi George, thanks for trying out Mojito!

Just left some quick initial comments. I need to read the spring thread you mentioned to better understand... wondering how we never had issues before (maybe something with the migration spring 1.x to 2.x)

Note that `webapp` and `common` modules run tests in `UTC` and it looks like `cli` doesn't. That might be needed for other tests. Doing so might hide this issue which will still need to be fixed for real CLI usage. https://github.com/box/mojito/blob/7155ac2a51c83cf90450737a589c961aef846c30/common/pom.xml#L342

restclient/src/main/java/com/box/l10n/mojito/rest/client/RepositoryClient.java
geoarion: That worked nicely and is much simpler that the initial PR (I previously assumed only dates in ISO 8601 were expected on the API side.)

restclient/src/main/java/com/box/l10n/mojito/rest/resttemplate/AuthenticatedRestTemplate.java
geoarion: Potentially, but I couldn't say with any certainty.
I noticed a comment in GitBlameWithUsageClient re another area where encoding was mentioned as a potential issue, though not sure what the current status of that is:

```
        // GitBlameWithUsageClient.getGitBlameWithUsages relies on [] not being encoded. keeps this for backward compatibility
        // TODO(spring2) - relaxed-query-chars - should we encode here instead?
        // I'm wondering why spring client doesn't encode those query param, it may be the way we call it that is not
        // good enough since it seems those should be encoded by default
        filterParams.put("repositoryIds[]", repositoryId.toString());
        filterParams.put("offset", offset.toString());
        filterParams.put("limit", batchSize.toString());
```

The current way the client is building the URI ensures _only_ illegal characters are escaped, as there's no .Encode() call in getURIForResourceAndQueryStringParams()
'+' and '[' , ']' are  legal characters - though, depending on the context in which they're used, sometimes they need escaping.
If that call would instead use: uriBuilder.build(false).**encode()**.toUri() , the [] would be escaped to %5B%5D .


In that specific instance though, I'm guessing long term ideally the param names would no longer have [] in their name. Today, there's a couple of collection-type params that have [] in their names:
- repositoryIds[]
- repositoryNames[]
- tmTextUnitIds[]
- localeTags[]

As backward compatibility is likely to be an issue there - one could set up the new API methods with the param names without the [] side by side with the old ones, mark the old ones as deprecated and make sure both methods call the same implementation.

Relevant: 
https://stackoverflow.com/questions/47217647/encoding-query-parameters-with-uricomponentsbuilder
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/util/UriComponentsBuilder.html#fromHttpUrl-java.lang.String-

geoarion: Side note: I reverted the previous commit in the same branch to preserve the conversation on this PR (and not rewrite the history), assuming the pull request would be merged with the squash option - in order to keep a clean history.

PR #700, opened 2021-07-22
Use smartling refresh token if it exists

maallen:
The Spring OAuth2RestTemplate getAccessToken() method calls the AccessTokenProvider obtainAccessToken method if the token is null or if the token is expired, see https://github.com/spring-projects/spring-security-oauth/blob/2.3.8.RELEASE/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/OAuth2RestTemplate.java#L221

The current code doesn't attempt to refresh the access token and instead calls the Smartling access token uri for a new token, I've added a check to see if an existing token with a refresh token exists and if so call the refresh access token flow to refresh the expired token.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingAuthorizationCodeAccessTokenProvider.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingAuthorizationCodeAccessTokenProvider.java
@@ -52,10 +52,10 @@ public class SmartlingAuthorizationCodeAccessTokenProvider implements AccessToke
 
         OAuth2AccessToken accessToken = null;
         OAuth2AccessToken existingToken = accessTokenRequest.getExistingToken();
-        if(existingToken != null && existingToken.getRefreshToken() != null){
+        if (existingToken != null && existingToken.getRefreshToken() != null) {
             logger.debug("Token exists with refresh token, refreshing access token");
             accessToken = refreshAccessToken(details, existingToken.getRefreshToken(), accessTokenRequest);
-        }else {
+        } else {
             try {
                 DateTime now = getNowForToken();
                 AuthenticationResponse authenticationResponse = getRestTemplate().postForObject(details.getAccessTokenUri(), request, AuthenticationResponse.class);
@@ -119,7 +119,7 @@ public class SmartlingAuthorizationCodeAccessTokenProvider implements AccessToke
         return DateTime.now().minusSeconds(15);
     }
 
-    protected RestTemplate getRestTemplate(){
+    protected RestTemplate getRestTemplate() {
         return restTemplate;
     }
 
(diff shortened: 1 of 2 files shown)


webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingAuthorizationCodeAccessTokenProvider.java
```diff
@@ -50,18 +50,24 @@ public OAuth2AccessToken obtainAccessToken(OAuth2ProtectedResourceDetails detail
...
-            DateTime now = getNowForToken();
-            AuthenticationResponse authenticationResponse = restTemplate.postForObject(details.getAccessTokenUri(), request, AuthenticationResponse.class);
-            defaultOAuth2AccessToken = getDefaultOAuth2AccessToken(now, authenticationResponse);
-        } catch (Exception e) {
-            String msg = "Can't get Smartling token";
-            logger.debug(msg, e);
-            throw new OAuth2AccessDeniedException(msg, details, e);
+        OAuth2AccessToken accessToken = null;
+        OAuth2AccessToken existingToken = accessTokenRequest.getExistingToken();
+        if(existingToken != null && existingToken.getRefreshToken() != null){
```
aurambaj, line 55: knit: can you re-indent the "if" statement? we don't have auto formatting but the default from Intellij should be a good first step.

webapp/src/test/java/com/box/l10n/mojito/smartling/SmartlingAuthorizationCodeAccessTokenProviderTest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.springframework.security.oauth2.common.DefaultOAuth2AccessToken;
+import org.springframework.security.oauth2.common.OAuth2AccessToken;
+import org.springframework.security.oauth2.common.OAuth2RefreshToken;
+import org.springframework.web.client.RestTemplate;
+
+import java.util.Map;
+
+import static org.mockito.ArgumentMatchers.*;
+
+public class SmartlingAuthorizationCodeAccessTokenProviderTest {
```
aurambaj, line 24: thanks for adding the tests

webapp/src/test/java/com/box/l10n/mojito/smartling/SmartlingAuthorizationCodeAccessTokenProviderTest.java
```diff
@@ -0,0 +1,91 @@
...
+    public void testRefreshTokenMethodNotCalledIfNoExistingToken(){
+
+        OAuth2AccessToken accessToken = smartlingAuthorizationCodeAccessTokenProvider.obtainAccessToken(mockResourceDetails, mockAccessTokenRequest);
+        Mockito.verify(smartlingAuthorizationCodeAccessTokenProvider, Mockito.times(0)).refreshAccessToken(isA(OAuth2ProtectedResourceDetails.class), isA(OAuth2RefreshToken.class), isA(AccessTokenRequest.class));
+        Assert.assertEquals(mockAuthData.getAccessToken(), accessToken.getValue());
+        Assert.assertEquals(mockAuthData.getRefreshToken(), accessToken.getRefreshToken().getValue());
+    }
+
+    @Test
+    public void testRefreshTokenMethodNotCalledIfNoRefreshToken(){
```
aurambaj, line 73: just for my understanding, is there a case where we won't have a refresh token in real life? or is that a theoretical case?

maallen: Theoretical test, I wanted to introduce a safe guard for any potential future refactor that we don't attempt the refresh flow if there's no refresh token available.

PR #701, opened 2021-07-23
Updated file download exception log message to include projectId & locale variable values

maallen:
Added the request projectId and locale variable strings to the exception message for additional information on error flows.


webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -62,7 +62,7 @@ public String getValue() {
 
     static final String ERROR_CANT_GET_FILES = "Can't get files";
     static final String ERROR_CANT_GET_SOURCE_STRINGS = "Can't get source strings";
-    static final String ERROR_CANT_DOWNLOAD_FILE = "Can't download file: %s ";
+    static final String ERROR_CANT_DOWNLOAD_FILE = "Can't download file: %s, projectId: %s, locale %s";
```
aurambaj, line 65: `locale:` ?

maallen: Good catch!

PR #702, opened 2021-07-27
Retry implementation for Smartling client

maallen:
Added project reactor retry logic for Smartling client calls, if a request fails there will be a maximum of 3 retry attempts before the request is marked as a failure and an exception is thrown.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
@@ -19,6 +19,7 @@ import com.box.l10n.mojito.service.tm.search.TextUnitSearcherParameters;
 import com.box.l10n.mojito.service.tm.search.UsedFilter;
 import com.box.l10n.mojito.smartling.AssetPathAndTextUnitNameKeys;
 import com.box.l10n.mojito.smartling.SmartlingClient;
+import com.box.l10n.mojito.smartling.SmartlingClientException;
 import com.box.l10n.mojito.smartling.SmartlingJsonKeys;
 import com.box.l10n.mojito.smartling.request.Binding;
 import com.box.l10n.mojito.smartling.request.Bindings;
@@ -37,10 +38,14 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Component;
 import org.xml.sax.SAXException;
+import reactor.core.publisher.Mono;
+import reactor.util.retry.Retry;
+import reactor.util.retry.RetryBackoffSpec;
 
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.transform.TransformerException;
 import java.io.IOException;
+import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -382,9 +452,17 @@ private SmartlingFile processTranslationBatch(List<TextUnitDTO> batch,
 
         if (!options.isDryRun()) {
             logger.debug("Push Android file to Smartling project: {} and locale: {}", projectId, localeTag);
-            smartlingClient.uploadLocalizedFile(projectId, sourceFilename, "android",
+            Mono.fromCallable(() -> smartlingClient.uploadLocalizedFile(projectId, sourceFilename, "android",
                     getSmartlingLocale(localeMapping, localeTag), file.getFileContent(),
-                    options.getPlaceholderFormat(), options.getCustomPlaceholderFormat());
+                    options.getPlaceholderFormat(), options.getCustomPlaceholderFormat()))
+                    .retryWhen(Retry.backoff(3, Duration.ofSeconds(1))
```
aurambaj, line 458: should we have a function that create the default retry since it is repeated a few times? also retrying right away the first time might be better (eg. for authentication issue ... though that should be fixed now) and have a max backoff of 1 second (`.maxBackoff(Duration.ofSeconds(1))`)?

webapp/src/main/java/com/box/l10n/mojito/smartling/SmartlingClient.java
```diff
@@ -135,6 +136,8 @@ public void deleteFile(String projectId, String fileUri) {
         } catch(HttpClientErrorException e) {
             throw wrapIntoSmartlingException(e, ERROR_CANT_DELETE_FILE, fileUri);
         }
+
+        return Mono.empty();
```
aurambaj, line 140: it seems odd to have return an empty Mono here since it is not really anything with it. I guess it is to help with `Mono.fromCallable`? maybe then using `Mono.fromRunnable` does the same out of the box?

maallen: Yes I added the `Mono.empty()` to allow the method be called by the `Mono.fromCallable`, `Mono.fromRunnable` achieves the same so I'll remove the `Mono.empty()` return type for both `SmartlingCLient.deleteFile` and `SmartlingClient.createBindings`

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -123,9 +132,19 @@ public ThirdPartyTMSSmartling(SmartlingClient smartlingClient,
...
+                    String msg = String.format("Error uploading image to Smartling, project id: %s, name: %s", projectId, name);
+                    logger.error(msg, e);
+                    throw new SmartlingClientException(msg, e);
+                })
+                .blockOptional();
+        contextUpload.orElseThrow(() -> new SmartlingClientException("Error with image upload to Smartling, context upload information is not present."));
+
         ThirdPartyTMSImage thirdPartyTMSImage = new ThirdPartyTMSImage();
-        thirdPartyTMSImage.setId(contextUpload.getContextUid());
+        thirdPartyTMSImage.setId(contextUpload.get().getContextUid());
```
aurambaj, line 147: you can move the `.orElseThrow(` after `blockOptional` and keep `ContextUpload` instead of `Optional<ContextUpload>`

and in fact you probably even do a `map` that instantiate the  `ThirdPartyTMSImage`

aurambaj (approved):

PR #703, opened 2021-07-30
Disable the screenshot upload button and update button text while upload is in progress

maallen:
Disables the screenshot upload button while an upload is in progress to prevent repeated upload requests of the same image, button text is changed to "Uploading Screenshot..." while an upload is in progress.


webapp/src/main/resources/public/js/components/branches/BranchesScreenshotUploadModal.js
```diff
@@ -8,6 +8,7 @@ class BranchesScreenshotUploadModal extends React.Component {
     static propTypes = {
         "show": PropTypes.bool.isRequired,
         "uploadDisabled": PropTypes.bool.isRequired,
```
aurambaj, line 10: Is "uploadDisabled" field used at all? should it be removed?

PR #704, opened 2021-08-04
Added S3 image storage & retrieval implementation

maallen:
Adds functionality to store and retrieve images from Amazon S3 storage. 

The S3 image service is enabled via the *l10n.image-service.storage.type* configuration parameter being set to s3 or s3Fallback, the service also requires that configuration is provided for an S3BlobStorage and an AmazonS3 client. The s3Fallback option will check the Mojito database for any requested images that can't be found in S3, if the image is found in the database then an asynchronous task will be triggered to upload the image to S3.

The existing database blob storage solution will be used if the *l10n.image-service.storage.type* parameter is missing or has a "database" value.

The prefix that images are stored in S3 under can be configured via the *l10n.image-service.storage.s3.prefix* parameter, if this is not provided it will default to "image" as the prefix.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/rest/images/ImageWS.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/rest/images/ImageWS.java
@@ -3,6 +3,7 @@ package com.box.l10n.mojito.rest.images;
 import com.box.l10n.mojito.entity.Image;
 import com.box.l10n.mojito.service.image.ImageService;
 import java.io.IOException;
+import java.util.Optional;
 import javax.servlet.http.HttpServletRequest;
 import org.apache.commons.io.FilenameUtils;
 import org.slf4j.Logger;
@@ -41,20 +42,13 @@ public class ImageWS {
 
         String imageName = getImageNameFromRequest(httpServletRequest);
 
-        Image image = imageService.getImage(imageName);
-
-        ResponseEntity responseEntity = null;
-
-        if (image != null) {
-            responseEntity = ResponseEntity
-                    .ok()
-                    .contentType(getMediaTypeFromImageName(imageName))
-                    .body(image.getContent());
-        } else {
-            responseEntity = ResponseEntity.notFound().build();
-        }
+        Optional<Image> image = imageService.getImage(imageName);
         
-        return responseEntity;
+        return image.map(i -> ResponseEntity
+                        .ok()
+                        .contentType(getMediaTypeFromImageName(imageName))
+                        .body(image.get().getContent()))
+                .orElseGet(() -> ResponseEntity.notFound().build());
     }
 
     @RequestMapping(value = "/api/images/**", method = RequestMethod.PUT)
--- a/webapp/src/main/java/com/box/l10n/mojito/service/image/DatabaseImageService.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/image/DatabaseImageService.java
@@ -1,0 +1,47 @@
+package com.box.l10n.mojito.service.image;
+
+import com.box.l10n.mojito.entity.Image;
+import org.slf4j.Logger;
+
+import java.util.Optional;
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+/**
+ * Service to uploadImage and serve images.
+ *
+ * @author jeanaurambault
+ */
+public class DatabaseImageService implements ImageService {
+
+    /**
+     * logger
+     */
+    static Logger logger = getLogger(ImageService.class);
+
+    ImageRepository imageRepository;
+
+    public DatabaseImageService(ImageRepository imageRepository) {
+        this.imageRepository = imageRepository;
+    }
+
+    public Optional<Image> getImage(String name) {
+        logger.debug("Get image with name: {}", name);
+        return imageRepository.findByName(name);
+    }
+
+    public void uploadImage(String name, byte[] content) {
+
+        logger.debug("Upload image with name: {}", name);
+
+        Image image = imageRepository.findByName(name).orElseGet(() -> {
+                Image img = new Image();
+                img.setName(name);
+                return img;
+        });
+
+        image.setContent(content);
+
+        imageRepository.save(image);
+    }
+}
--- a/webapp/src/main/java/com/box/l10n/mojito/service/image/ImageServiceConfiguration.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/image/ImageServiceConfiguration.java
@@ -1,0 +1,70 @@
+package com.box.l10n.mojito.service.image;
+
+import com.box.l10n.mojito.service.blobstorage.s3.S3BlobStorage;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Primary;
+
+@Configuration
+public class ImageServiceConfiguration {
+
+    @ConditionalOnProperty(value = "l10n.image-service.storage.type", havingValue = "s3Fallback")
+    static class S3FallbackImageServiceConfiguration {
+
+        @Autowired
+        ImageRepository imageRepository;
+
+        @Autowired
+        S3BlobStorage s3BlobStorage;
+
+        @Bean @Qualifier("databaseImageService")
+        public DatabaseImageService databaseImageService() {
+            return new DatabaseImageService(imageRepository);
+        }
+
+        @Bean @Qualifier("s3ImageService")
+        public S3ImageService s3ImageService(S3BlobStorage s3BlobStorage, @Value("${l10n.image-service.storage.s3.prefix:image}") String s3PathPrefix) {
+            return new S3ImageService(s3BlobStorage, s3PathPrefix);
+        }
+
+        @Bean
+        public S3UploadImageAsyncTask s3UploadImageAsyncTask(S3ImageService s3ImageService) {
+            return new S3UploadImageAsyncTask(s3ImageService);
+        }
+
+        @Bean @Primary
+        public ImageService s3ImageFallback(@Qualifier("s3ImageService") S3ImageService s3ImageService,
+                                            @Qualifier("databaseImageService") DatabaseImageService databaseImageService,
+                                            S3UploadImageAsyncTask s3UploadImageAsyncTask) {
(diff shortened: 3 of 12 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/image/S3ImageService.java
```diff
@@ -0,0 +1,69 @@
...
+    static Logger logger = getLogger(S3ImageService.class);
+
+    @Autowired
+    S3BlobStorage s3BlobStorage;
+
+    @Autowired
+    S3UploadImageAsyncTask s3UploadImageAsyncTask;
+
+    @Autowired
+    ImageRepository imageRepository;
```
aurambaj, line 41: should this be delegated to `DatabaseImageService`?

webapp/src/main/java/com/box/l10n/mojito/service/image/S3ImageService.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public Image getImage(String name) {
```
aurambaj, line 44: what about having have another service implementation: `S3FallbackImageService`. This way the `S3ImageService` stays clean.

S3FallbackImageService#get would just be getting from S3 implementation and then from database implementation (probably need to return an Optional<Image> in the interface) and optionally trigger the async task.

webapp/src/main/java/com/box/l10n/mojito/service/image/S3UploadImageAsyncTask.java
```diff
@@ -0,0 +1,35 @@
...
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+/**
+ * Async task to upload an image to S3.
+ *
+ * @author maallen
+ */
+@Component
+@ConditionalOnProperty(value = "l10n.blob-storage.s3.images.enabled", havingValue = "true")
```
aurambaj, line 19: @ConditonialOnBean(S3ImageService)?

webapp/src/main/java/com/box/l10n/mojito/service/image/S3UploadImageAsyncTask.java
```diff
@@ -0,0 +1,35 @@
...
+public class S3UploadImageAsyncTask {
+
+    /**
+     * logger
+     */
+    static Logger logger = getLogger(S3UploadImageAsyncTask.class);
+
+    @Autowired
+    S3BlobStorage s3BlobStorage;
```
aurambaj, line 28: use S3ImageService?

webapp/src/main/java/com/box/l10n/mojito/service/image/S3ImageService.java
maallen: The DatabaseImageService bean won't be present in the container if the S3ImageService is active which is why I used the ImageRepository bean. 

If the DatabaseImageService is active we'd have two active ImageService beans in the container and would then need to qualify which bean type to use in ImageWS & ThirdPartyService in different circumstances, I think keeping the specific implementation hidden from the higher-level beans is cleaner.

aurambaj: Can we have something more similar to `BlobStorageConfiguration`, instead of enable/disable it can have different values ` @ConditionalOnProperty(value = "l10n.blob-storage.type", havingValue = "s3")`. 

So 3 types `database`, `s3`, `s3Fallback` and then some configration like that do deal with multi instance:

```
@ConditionalOnProperty(value = "l10n.blob-storage.type", havingValue = "s3Fallback")`. 
static class S3FallbackImageServiceConfiguration {
    @Bean @Qualifier("database") 
    database()

    @Bean @Qualifier("s3") 
    s3()

     @Bean @Primary
    s3Fallback(@Qualifier("database) db, @Qualifier("s3")
}

@ConditionalOnProperty(value = "l10n.blob-storage.type", havingValue = "s3")`. 
static class S3ImageServiceConfiguration {
    @Bean 
    s3()
}

@ConditionalOnProperty(value = "l10n.blob-storage.type", havingValue = "database")`. 
static class DatabaseImageServiceConfiguration {
    @Bean 
    database()
}
```

maallen: Sure I can add something more flexible

webapp/src/main/java/com/box/l10n/mojito/service/image/S3FallbackImageService.java
```diff
@@ -0,0 +1,54 @@
...
+                                  S3UploadImageAsyncTask s3UploadImageAsyncTask) {
+        this.s3ImageService = s3ImageService;
+        this.databaseImageService = databaseImageService;
+        this.s3UploadImageAsyncTask = s3UploadImageAsyncTask;
+    }
+
+    @Override
+    public Image getImage(String name) {
+        logger.debug("Attempt image retrieval from S3 with name: {}", name);
+        Image image = Optional.ofNullable(s3ImageService.getImage(name)).orElseGet(() ->
```
aurambaj, line 39: make the ImageService return an Optional? you did the change on the repository interface but not the service one, I guess it would simplify here too.

Eventually, we could even use project reactor to get retries and fallback logic (switchIfEmpty) out of the box.

aurambaj (commented):
Looks good! just one minor comment.

webapp/src/main/java/com/box/l10n/mojito/rest/images/ImageWS.java
```diff
@@ -41,15 +42,15 @@ public ResponseEntity getImage(HttpServletRequest httpServletRequest) throws IOE
 
         String imageName = getImageNameFromRequest(httpServletRequest);
 
-        Image image = imageService.getImage(imageName);
+        Optional<Image> image = imageService.getImage(imageName);
 
         ResponseEntity responseEntity = null;
 
-        if (image != null) {
+        if (image.isPresent()) {
```
aurambaj, line 49: With optional you can remove the `if` construct:

```
return image.map(i -> ResponseEntity
                    .ok()
                    .contentType(getMediaTypeFromImageName(imageName))
                    .body(image.get().getContent()))
.orElseGet(ResponseEntity.notFound().build())
```

webapp/src/main/java/com/box/l10n/mojito/service/image/DatabaseImageService.java
```diff
@@ -0,0 +1,50 @@
...
+        return imageRepository.findByName(name);
+    }
+
+    public void uploadImage(String name, byte[] content) {
+
+        logger.debug("Upload image with name: {}", name);
+        Image image;
+        Image prevVersion = imageRepository.findByName(name).orElse(null);
+
+        if (prevVersion != null) {
```
aurambaj, line 39: same here, something like: 

```
image = imageRepository.findByName(name).orElseGet(() -> { Image image = new Image(); image.setName(name)});
image.setContent(content);
imageRepository.save(image)

File: webapp/src/main/java/com/box/l10n/mojito/service/image/ImageServiceConfiguration.java [hunk 8]
```diff
@@ -0,0 +1,70 @@
...
+    @ConditionalOnProperty(value = "l10n.image-service.storage.type", havingValue = "s3Fallback")
+    static class S3FallbackImageServiceConfiguration {
+
+        @Autowired
+        ImageRepository imageRepository;
+
+        @Autowired
+        S3BlobStorage s3BlobStorage;
+
+        @Bean("databaseImageService")
```
aurambaj, line 24: I'm not 100% sure but on top of my mind `@Bean("databaseImageService")` defines the bean name (and not qualifier). If not provided it uses the method names. So, here it is probably redundant since it is the same as the method name. 

I think it is either 
- `@Bean @Qualifier("databaseImageService")` with `@Qualifier("databaseImageService")` in the method definition. 
- Or if using bean names `@Bean("databaseImageService")` (which with this method name could be @Bean only) it matches on the method param names `databaseImageService` and the qualifier is not used. 

If my assumption are correct, to make it easier to read and less brittle to code changes, it is better to use @Qualifier

webapp/src/main/java/com/box/l10n/mojito/service/image/S3FallbackImageService.java
```diff
@@ -0,0 +1,52 @@
...
+                                  S3UploadImageAsyncTask s3UploadImageAsyncTask) {
+        this.s3ImageService = s3ImageService;
+        this.databaseImageService = databaseImageService;
+        this.s3UploadImageAsyncTask = s3UploadImageAsyncTask;
+    }
+
+    @Override
+    public Optional<Image> getImage(String name) {
+        logger.debug("Attempt image retrieval from S3 with name: {}", name);
+        return Optional.ofNullable(s3ImageService.getImage(name).orElseGet(() ->
```
aurambaj, line 39: that can be simplified

```
s3ImageService.getImage(name).orElseGet(() ->
                databaseImageService.getImage(name).map(img -> {
                    logger.debug("Found image {} in database, triggering async upload to S3", img.getName());
                    s3UploadImageAsyncTask.uploadImageToS3(img.getName(), img.getContent());
                    return img;
                })
```

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyService.java
```diff
@@ -329,7 +329,7 @@ void uploadImageWithMappings(String projectId, Screenshot screenshot, Image imag
 
         Optional<Image> image = Optional.ofNullable(screenshot.getSrc())
                 .map(src -> src.replaceAll("^api/images/", ""))
-                .map(name -> imageService.getImage(name));
+                .map(name -> imageService.getImage(name).get());
```
aurambaj, line 332: `map` --> `flapMap()`?

webapp/src/main/java/com/box/l10n/mojito/service/image/S3FallbackImageService.java
maallen: The lamda function supplied to orElseGet requires a type of Image returned so I need to wrap it back into an Optional to meet the interface contract, I've moved the return Option.ofNullable() to a new line so it looks less cluttered.

aurambaj: oh yes.. Java 8 doesn't have `or()`, it is available starting Java 9.

I've seen that pattern too, maybe a slight more simple:

```
s3ImageService.getImage(name).map(Optional::of)
.orElseGet(() -> databaseImageService.getImage(name).map(img -> {
                    logger.debug("Found image {} in database, triggering async upload to S3", img.getName());
                    s3UploadImageAsyncTask.uploadImageToS3(img.getName(), img.getContent());
                    return img;
                })
```

aurambaj (approved):

PR #710, opened 2021-08-17
Update error message if screenshot upload failure is due to unauthorized response

maallen:
Updates the error message observed on screenshot upload failure to indicate that the upload has failed as user authentication has expired.

Updated the putBinaryData function to treat a redirect response as an error as when OAuth is being used for login the initial image upload fails as the response is an OAuth redirect to re-authorize. This isn't treated as an error by the frontend but triggers multiple redirects under the hood which eventually fail with a TOO_MANY_REDIRECTS error, this causes the error to be displayed but doesn't trigger the handleUnAuthenticatedRequests flow.


garionpin (approved):

maallen:
Handling for user auth checks on page focus done in PR: https://github.com/box/mojito/pull/717

PR #711, opened 2021-08-18
Fix for occasional duplicate key errors on session create or update 

maallen:
Updates the CreateSessionAttribute query to perform an UPDATE on the attribute bytes if a duplicate session id is found, this is a workaround for an issue in spring session where concurrent requests adding the same attribute result in a duplicate key violation, see discussion on https://github.com/spring-projects/spring-session/issues/1213 for further information.


garionpin (approved):

webapp/src/main/java/com/box/l10n/mojito/service/security/user/session/CustomCreateSessionAttributeInsertQueryConfiguration.java
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.session.jdbc.config.annotation.web.http.JdbcHttpSessionConfiguration;
+import org.springframework.util.StringUtils;
+
+
+/**
+ * Workaround for issue <a href="https://github.com/spring-projects/spring-session/issues/1213">Concurrent requests adding the same session attribute result in duplicate key constraint violation</a>
+ * to change the Create Session Attribute query to use an MySQL UPDATE command to update the ATTRIBUTE_BYTES
+ * of the existing row if a duplicate key is encountered.
+ *
+ * <b>NOTE:</b>This custom insert query should only be used with a MySQL database.
```
aurambaj, line 17: `only be used with a MySQL database`, can this be made explicit with a `@Conditional`?

maallen: I've pushed an update containing a custom condition (MySQLDatabaseCondition) that checks the prefix of the `spring.datasource.url` property, if the prefix matches "jdbc:mysql:" the condition passes

webapp/src/main/java/com/box/l10n/mojito/service/security/user/session/MySQLDatabaseCondition.java
```diff
@@ -0,0 +1,21 @@
...
+ * Conditional that checks if a MySQL database is in use by checking if the
+ * <b>spring.datasource.url</b> property is prefixed with <b>jdbc:mysql:</b>
+ *
+ * @author maallen
+ */
+public class MySQLDatabaseCondition implements Condition {
+
+    @Override
+    public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+        String datasourceUrl = context.getEnvironment().getProperty("spring.datasource.url");
```
aurambaj, line 18: could `DBUtils` be reused/merged in some ways? not super important but if we can avoid duplication and have a single place with that logic.

maallen: I missed the DBUtils class, I don't think it can be merged into the condition as the condition logic is executed before the DBUtils bean is initialised but I could remove the conditional and autowire it into the CreateSessionAttributeInsertQuery bean. I'll update to execute the check in there and if there's not a mysql db then  we'll log a warning stating the custom query has not been set.

PR #717, opened 2021-08-30
Check for authentication when app gains focus

maallen:
Check if user authentication is still valid when tab regains focus, if not redirect to login page

### DIFF
--- a/webapp/src/main/resources/public/js/app.js
+++ b/webapp/src/main/resources/public/js/app.js
@@ -96,37 +96,37 @@ function startApp(messages) {
     }
 
     ReactDOM.render(
-            <AppConfig appConfig={APP_CONFIG}>
-                <IntlProvider locale={APP_CONFIG.locale} messages={messages}>
-                    <Router history={browserHistory}>
-                        <Route component={Main}>
-                            <Route path="/" component={App}
-                                onEnter={onEnterRoot}>
-                                <Route path="workbench" component={Workbench}
-                                       onEnter={getAllRepositoriesDeffered}
-                                       onLeave={onLeaveWorkbench}/>
-                                <Route path="repositories" component={Repositories}
-                                       onEnter={getAllRepositoriesDeffered}/>
-                                <Route path="project-requests" component={Drops}/>
-                                <Route path="branches" component={BranchesPage}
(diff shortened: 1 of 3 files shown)


webapp/src/main/resources/public/js/app.js
```diff
@@ -140,6 +140,10 @@ function startApp(messages) {
             let pathNameStrippedLeadingSlash = location.pathname.substr(1 + APP_CONFIG.contextPath.length, location.pathname.length);
             let currentLocation = pathNameStrippedLeadingSlash + window.location.search;
 
+            if (currentLocation === "") {
+                currentLocation = "workbench"
```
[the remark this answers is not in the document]
maallen: It handles the edge case where the user logs in and doesn't do anything and the token expires, in this scenario the showPath query parameter to the login page is an empty string. When the user logs back in the browser directs them to baseurl + api/users/session (which is an empty page), this defaults the entry point to workbench if showPath will be empty.

webapp/src/main/resources/public/js/app.js
```diff
@@ -146,30 +146,34 @@ function startApp(messages) {
                 // re-authenticate and the load the API response instead of showing the frontend
                 window.location.href = UrlHelper.getUrlWithContextPath(currentLocation);
             } else {
-                // if log through the login page use showPage to redirect to the requested page before auth failed
-                window.location.href = UrlHelper.getUrlWithContextPath("/login?") + UrlHelper.toQueryString({"showPage": currentLocation});
+                if (currentLocation === "") {
+                    window.location.href = UrlHelper.getUrlWithContextPath("/")
```
aurambaj, line 150: would you mind adding a comment to explain this is added?

webapp/src/main/resources/public/js/app.js
```diff
@@ -146,30 +146,34 @@ function startApp(messages) {
                 // re-authenticate and the load the API response instead of showing the frontend
                 window.location.href = UrlHelper.getUrlWithContextPath(currentLocation);
```
aurambaj, line 147: when i tried that path -dev, it seemed to go back to the repository page? given the comment I was wondering if it will go to the API instead. Have you done more testing around that?

PR #719, opened 2021-09-01
Add Workbench filter to search by tmTextUnitId

garionpin:
Added a new "TM Text Unit ID" option in the Workbench filter dropdown
to be able to search by the IDs of the text units.

### DIFF
--- a/webapp/src/main/resources/properties/en.properties
+++ b/webapp/src/main/resources/properties/en.properties
@@ -145,6 +145,9 @@ search.filter.asset=Asset
 # Search filter option "plural form"
 search.filter.pluralFormOther=Plural of
 
+# Search filter option "tm text unit ID"
+search.filter.tmTextUnitId=Tm Text Unit ID
+
 # Search filter option to search for strings that matche exactly the search critera
 search.filter.exact=Exact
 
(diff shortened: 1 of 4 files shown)


webapp/src/main/resources/properties/de.properties
```diff
@@ -142,6 +142,9 @@ search.filter.asset=Asset
 # Search filter option "plural form"
 search.filter.pluralFormOther=Der Plural von
 
+# Search filter option "tm text unit ID"
+search.filter.tmTextUnitId=Tm Text Unit ID
```
aurambaj, line 146: localized files are generated, no need to edit them

aurambaj (commented):
Looks good! The deep linking/reload for URL was already implemented, right?

webapp/src/main/resources/properties/de.properties
garionpin: I've removed the non En files.

garionpin:
> Looks good! The deep linking/reload for URL was already implemented, right?

Yes, that was already implemented - I did however test the functionality again as part of these changes to make sure everything still works for both scenarios.

PR #721, opened 2021-09-09
Change fallback logic to reduce DB calls

garionpin:
Avoid DB calls if no text units have been added in the latest diff extraction.


aurambaj:
@geoarion It would be nice to have a test that ensure we're not making API call unless for pushing new strings. Might be done by pointing the resttemplate to a wrong url/port in CLITestBase.

Merging that this PR for now.

PR #722, opened 2021-09-14
Added new PrintfLike integrity checkers

maallen:
Added `PrintfLikeIgnorePercentageAfterBracketsIntegrityChecker` & `PrintfLikeVariableTypeIntegrityChecker` integrity checkers.

`PrintfLikeIgnorePercentageAfterBracketsIntegrityChecker` checker ignores '%' symbols as tokens if they are immediately preceded by a closing bracket ('}' or ')').

`PrintfLikeVariableTypeIntegrityChecker` verifies that token types are not modified in the translated string, note that the token name must be enclosed within brackets to be checked.


restclient/src/main/java/com/box/l10n/mojito/rest/entity/IntegrityCheckerType.java
```diff
@@ -16,5 +16,7 @@
     HTML_TAG,
     ELLIPSIS,
     BACKQUOTE,
-    EMPTY_TARGET_NOT_EMPTY_SOURCE;
+    EMPTY_TARGET_NOT_EMPTY_SOURCE,
+    PRINTF_LIKE_IGNORE_PERCENTAGE_AFTER_BRACKETS,
+    PRINTF_LIKE_VARIABLE_TYPE;
```
aurambaj, line 21: group them with the other PRINTF?

webapp/src/test/java/com/box/l10n/mojito/service/assetintegritychecker/integritychecker/PrintLikeVariableTypeIntegrityCheckerTest.java
```diff
@@ -0,0 +1,103 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class PrintLikeVariableTypeIntegrityCheckerTest {
+
+    @Test
+    public void testCheckPassesIfTypeVariablesMatch() throws PrintfLikeIntegrityCheckerException {
+        PrintfLikeVariableTypeIntegrityChecker checker = new PrintfLikeVariableTypeIntegrityChecker();
+        String source = "%(count)s view";
+        String target = "%(count)s view";
```
aurambaj, line 14: btw, do you know what format that is? is it standard?

aurambaj (approved):

webapp/src/test/java/com/box/l10n/mojito/service/assetintegritychecker/integritychecker/PrintLikeVariableTypeIntegrityCheckerTest.java
ehoogerbeets: @aurambaj That style is used in python: https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting

webapp/src/test/java/com/box/l10n/mojito/service/assetintegritychecker/integritychecker/PrintfLikeIgnorePercentageAfterBracketsIntegrityCheckerTest.java
```diff
@@ -0,0 +1,296 @@
...
+    @Test
+    public void testTokenIsNotRecognizedIfPrecededByRightBracket() throws PrintfLikeWhitespaceInTokenInvalidIntegrityCheckerException {
+        // Verify that "% to" is not matched as a token
+        PrintfLikeIgnorePercentageAfterBracketsIntegrityChecker checker = new PrintfLikeIgnorePercentageAfterBracketsIntegrityChecker();
+        String source = "from (from_val)% to (to_val)%";
+        String target = "(from_val)%-ról (to_val)%-ra";
+
+        checker.check(source, target);
+    }
+
```
ehoogerbeets, line 295: Mark, first, thanks for adding this checker!

Please also add unit tests for the following:

- negative tests where % is used as a plain old percent sign. eg. "The level is down to 10% of the original." should be an error because the percent char is not doubled.
- tests where the % is doubled to make an actual percent sign appear in the output. eg. "The level is down to 10%% of the original." should not be an error.
- a test in Turkish. My projects often fail on Turkish because they format percentages where the percent symbol comes first: "The level is down to 10% of the original." -> "Seviye, orijinalin %10'una düştü." should not be an error (but often is with the other integrity checkers)

maallen: Hi Edwin,
Thanks for the comments, I've pushed an update that adds the above tests. I've made the assumption that for the turkish test the percentages should be doubled i.e. "The level is down to 10%% of the original." -> "Seviye, orijinalin %%10'una düştü."

ehoogerbeets: Yes, good point about the Turkish test! Thanks for adding those extra tests

restclient/src/main/java/com/box/l10n/mojito/rest/entity/IntegrityCheckerType.java
```diff
@@ -17,4 +19,5 @@
     ELLIPSIS,
     BACKQUOTE,
     EMPTY_TARGET_NOT_EMPTY_SOURCE;
+    ;
```
aurambaj, line 22: extra `;` ?

aurambaj (approved):

PR #723, opened 2021-09-15
Test to ensure no API call are made for empty diffs

garionpin:
Test to ensure no API call are made unless there are new text units founds in the extraction diff.


aurambaj (approved):
thanks!

PR #724, opened 2021-09-16
Added log on integrity check failure

maallen:
Added a log on integrity check failure to provide the reason for the failure.


webapp/src/main/java/com/box/l10n/mojito/rest/textunit/TextUnitWS.java
```diff
@@ -361,6 +361,7 @@ public TMTextUnitIntegrityCheckResult checkTMTextUnit(@RequestBody TextUnitCheck
             tmTextUnitIntegrityCheckService.checkTMTextUnitIntegrity(textUnitCheckBody.getTmTextUnitId(), textUnitCheckBody.getContent());
             result.setCheckResult(true);
         } catch (IntegrityCheckException e) {
+            logger.info("Integrity check failed for text unit id {}, content {}: {}", textUnitCheckBody.getTmTextUnitId(), textUnitCheckBody.getContent(), e.getMessage());
```
garionpin, line 364: Would there be value from also getting the stack trace of the exception logged in this case (vs. just the message of the exception)? (Same question applies to changes in the other file).

aurambaj (approved):

PR #738, opened 2021-11-17
Added RunChecks cli command with spell & comment/context checker

maallen:
Adds a cli command that runs checks on newly added source strings from a list on checkers. The command soft fails on check failures by default but can be configured to hard fail for specific failures. A phabricator diff id can be added as an optional command parameter, if provided a comment will be added to the relevant diff with any failures found in the soft failure scenario.

The following checkers are included in this PR:

**SPELL_CHECKER** checks for spelling errors within the added source strings, on failure it provides suggestions on correct spellings for the misspelt word.

**CONTEXT_COMMENT_CHECKER** verifies that both a context and comment have been provided with the source string and that they are not identical to each other.

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/Command.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/Command.java
@@ -3,9 +3,12 @@ package com.box.l10n.mojito.cli.command;
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
+import com.ibm.icu.text.MessageFormat;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Base class for Commands, provides basic support for usage display.
@@ -116,4 +119,5 @@ public abstract class Command {
     public void setOriginalArgs(List<String> originalArgs) {
         this.originalArgs = originalArgs;
     }
+
 }
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
@@ -1,0 +1,258 @@
+package com.box.l10n.mojito.cli.command;
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
+import com.box.l10n.mojito.cli.command.checks.AbstractCliChecker;
+import com.box.l10n.mojito.cli.command.checks.CliCheckResult;
+import com.box.l10n.mojito.cli.command.checks.CliCheckerExecutor;
+import com.box.l10n.mojito.cli.command.checks.CliCheckerOptions;
+import com.box.l10n.mojito.cli.command.checks.CliCheckerType;
+import com.box.l10n.mojito.cli.command.extraction.AssetExtractionDiff;
+import com.box.l10n.mojito.cli.command.extraction.ExtractionDiffPaths;
+import com.box.l10n.mojito.cli.command.extraction.ExtractionDiffService;
+import com.box.l10n.mojito.cli.command.extraction.ExtractionPaths;
+import com.box.l10n.mojito.cli.command.extraction.MissingExtractionDirectoryExcpetion;
+import com.box.l10n.mojito.cli.console.ConsoleWriter;
+import com.box.l10n.mojito.phabricator.DifferentialRevision;
+import com.box.l10n.mojito.phabricator.PhabricatorMessageBuilder;
+import com.box.l10n.mojito.regex.PlaceholderRegularExpressions;
+import com.google.common.collect.ImmutableMap;
+import org.fusesource.jansi.Ansi;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Command to execute checks against any new source strings
+ */
+@Component
+@Scope("prototype")
+@Parameters(commandNames = {"extraction-check"}, commandDescription = "Execute checks against new source strings")
+public class ExtractionCheckCommand extends Command {
+
+    static Logger logger = LoggerFactory.getLogger(ExtractionDiffCommand.class);
+
+    @Autowired
+    ExtractionDiffService extractionDiffService;
+
+    @Autowired
+    ConsoleWriter consoleWriter;
+
+    @Autowired(required = false)
+    PhabricatorMessageBuilder phabricatorMessageBuilder;
+
+    @Autowired(required = false)
+    DifferentialRevision differentialRevision;
+
+    @Parameter(names = {"--checker-list", "-cl"}, arity = 1, required = true, description = "List of checks to be run against new source strings")
+    List<String> checkerList;
+
+    @Parameter(names = {"--hard-fail", "-hf"}, arity = 1, required = false, description = "List of checks that will cause a hard failure, use ALL if all checks should be hard failures")
+    List<String> hardFailList = new ArrayList<>();
+
+    @Parameter(names = {"--parameter-regexes", "-pr"}, arity = 1, required = false, description = "Regex types used to identify parameters in source strings")
+    List<String> parameterRegexList = new ArrayList<>();
+
+    @Parameter(names = {"--dictionary-file", "-df"}, arity = 1, required = false, description = "Path to the dictionary file to be used by the spelling check. The file must be in .dic format")
+    String dictionaryFilePath = null;
+
+    @Parameter(names = {"--dictionary-affix-file", "-daff"}, arity = 1, required = false, description = "Path to the dictionary affix file to be used by the spelling check. The file must be in .aff format")
+    String dictionaryAffixFilePath = null;
+
+    @Parameter(names = {"--dictionary-additions-file", "-daf"}, arity = 1, required = false, description = "Path to the dictionary additions file used for the spelling check. The file format is a new spelling per line.")
+    String dictionaryAdditionsFilePath = "";
+
+    @Parameter(names = {"--glossary-file", "-gf"}, arity = 1, required = false, description = "Path to the glossary file used for the glossary check. The file is a json formatted file containing an array of glossary term objects, each object contains the 'term' and a 'severity' which is 'MAJOR' or 'MINOR'.")
+    String glossaryFilePath = "";
+
+    @Parameter(names = {"--name", "-n"}, arity = 1, required = false, description = ExtractionDiffCommand.EXCTRACTION_DIFF_NAME_DESCRIPTION)
+    String extractionDiffName = null;
+
+    @Parameter(names = {"--objectid", "-oid"}, arity = 1, required = false, description = "Phabricator object id, required if adding a comment with check failures to Phabricator.")
+    String phabObjectId = null;
+
+    @Parameter(names = {"--phab-message-template", "-pmt"}, arity = 1, required = false, description = "Optional message template to customize the Phabricator notification message. eg. '{baseMessage}. Check [[https://build.org/1234|build]].' ")
+    String messageTemplate = "{baseMessage}";
+
+    @Parameter(names = {"--current", "-c"}, arity = 1, required = true, description = ExtractionDiffCommand.CURRENT_EXTRACTION_NAME_DESCRIPTION)
+    String currentExtractionName;
+
+    @Parameter(names = {"--base", "-b"}, arity = 1, required = true, description = ExtractionDiffCommand.BASE_EXTRACTION_NAME_DESCRIPTION)
+    String baseExtractionName;
+
+    @Parameter(names = {"--output-directory", "-o"}, arity = 1, required = false, description = ExtractionDiffCommand.OUTPUT_DIRECTORY_DESCRIPTION)
+    String outputDirectoryParam = ExtractionDiffPaths.DEFAULT_OUTPUT_DIRECTORY;
+
+    @Parameter(names = {"--input-directory", "-i"}, arity = 1, required = false, description = ExtractionDiffCommand.INPUT_DIRECTORY_DESCRIPTION)
+    String inputDirectoryParam = ExtractionPaths.DEFAULT_OUTPUT_DIRECTORY;
+
+    @Override
+    protected void execute() throws CommandException {
+
+        checkPhabricatorPreconditions();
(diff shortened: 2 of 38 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/checks/ContextAndCommentCliChecker.java
```diff
@@ -0,0 +1,117 @@
...
+        StringBuilder notificationText = new StringBuilder();
+        runChecks(cliCheckResult, notificationText);
+        if(!cliCheckResult.isSuccessful()) {
+            cliCheckResult.setNotificationText("Context and comment check found failures:" + System.lineSeparator() + notificationText);
+        }
+        return cliCheckResult;
+    }
+
+    private void runChecks(CliCheckResult cliCheckResult, StringBuilder notificationText) {
+        getAddedTextUnits().stream()
```
aurambaj, line 64: Is it possible to have the logic that run the check and then logic that builds the report separated? in theory we'd try to have avoid side effects in stream too

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/ContextAndCommentCliChecker.java
```diff
@@ -0,0 +1,117 @@
...
+            .map(assetExtractorTextUnit -> getContextAndCommentCliCheckerResult(assetExtractorTextUnit, checkTextUnit(assetExtractorTextUnit)))
+            .filter(result -> result.isFailed())
```
aurambaj, line 66: knit: Result::isFailed

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/SpellCliChecker.java
```diff
@@ -0,0 +1,228 @@
...
+ *
+ * @author mallen
+ */
+public class SpellCliChecker extends AbstractCliChecker {
+
+    static Logger logger = LoggerFactory.getLogger(SpellCliChecker.class);
+
+    Hunspell hunspell = Hunspell.forDictionaryInResources("en_US", "dictionaries/");
+
+    class SpellCliCheckerResult {
```
aurambaj, line 37: should this be static?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/SpellCliChecker.java
```diff
@@ -0,0 +1,228 @@
...
+                List<String> suggestions = hunspell.suggest(word);
+                logger.debug("{} is spelt incorrectly. Suggested correct spellings are {}", word, suggestions);
+                failureMap.put(word, suggestions);
+                checked.add(word);
+            }
+        }
+        return failureMap;
+    }
+
+    private List<String> getSourceStringsFromDiff() {
```
aurambaj, line 180: do you foresee this to be used by more checkers? should it be in the abstract class?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/SpellCliChecker.java
```diff
@@ -0,0 +1,228 @@
...
+        List<String> sourceStrings = new ArrayList<>();
+        getAddedTextUnits().stream().forEach(assetExtractorTextUnit -> sourceStrings.add(assetExtractorTextUnit.getSource()));
+        return sourceStrings;
+    }
+
+    private String buildNotificationText(Map<String, Map<String, List<String>>> failureMap) {
```
aurambaj, line 186: Should the notification builder be in its own class?

cli/src/main/java/com/box/l10n/mojito/cli/command/RunChecksCommand.java
```diff
@@ -0,0 +1,278 @@
...
+import java.util.stream.Stream;
+
+import static com.google.common.base.Strings.nullToEmpty;
+
+/**
+ * Command to execute checks against any new source strings
+ */
+@Component
+@Scope("prototype")
+@Parameters(commandNames = {"run-checks"}, commandDescription = "Execute checks against new source strings")
```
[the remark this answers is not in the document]
maallen: I like extraction-check

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/CliCheckerType.java
```diff
@@ -0,0 +1,19 @@
+package com.box.l10n.mojito.cli.command.checks;
+
+public enum CliCheckerType {
+
+    SPELL_CHECKER(SpellCliChecker.class.getName()),
```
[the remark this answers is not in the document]
maallen: It's used in the RunChecksCommand class (to be renamed to ExtractionCheckCommand) to generate the list of checkers to execute and build the hard failure list passed from the command line. Basic use is to map what the user provides on the cli to an implementation checker class

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/SpellCliChecker.java
maallen: Initially that was my intention, every checker would call a separate class to build up the failure notification text. However the way the different checkers display failures isn't uniform, for example the SpellChecker needs to have a way to display multiple errors in a single source string as its possible for there to be multiple misspellings in a single string and then it also needs to provide suggestions for those misspellings.

Alternatively the  Context & Comment checker has a one to one mapping with strings and failures so it doesn't follow the same text format as the spell checker which is why I decided to let each checker manage its own failure text .

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
```diff
@@ -0,0 +1,253 @@
...
+                .build();
+        String message = "{icon} {header}" + System.lineSeparator() + System.lineSeparator() + "{notificationText}";
+        return getFormattedPhabricatorMessage(messageTemplate, "baseMessage", getBaseMessage(messageParamMap, message));
+    }
+
+    private String buildPhabricatorComment(List<CliCheckResult> failedCheck) {
+        StringBuilder sb = new StringBuilder();
+        sb.append("**Failed checks:**");
+        addDoubleNewLine(sb);
+        failedCheck.stream().forEach(check -> {
```
aurambaj, line 175: same, avoid side effect inside stream using `.collect(Collectors.joining(System.lineSeparator))`

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    private CliCheckerOptions generateCheckerOptions() {
+        Set<PlaceholderRegularExpressions> regexSet = generateParameterRegexSet();
+        Set<String> hardFailureSet = generateHardFailureSet();
+        return new CliCheckerOptions(regexSet, hardFailureSet, dictionaryAdditionsFilePath, glossaryFilePath);
+    }
+
+    private Set<String> generateHardFailureSet() {
+        if(hardFailList.stream().anyMatch(checkName -> checkName.equalsIgnoreCase("ALL"))) {
```
aurambaj, line 217: swap the check to avoid NPE, `"ALL".equalsIgnoreCase(checkName)`

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
```diff
@@ -0,0 +1,253 @@
...
+        });
+    }
+
+    private CliCheckerOptions generateCheckerOptions() {
+        Set<PlaceholderRegularExpressions> regexSet = generateParameterRegexSet();
+        Set<String> hardFailureSet = generateHardFailureSet();
+        return new CliCheckerOptions(regexSet, hardFailureSet, dictionaryAdditionsFilePath, glossaryFilePath);
+    }
+
+    private Set<String> generateHardFailureSet() {
```
aurambaj, line 216: Knit: renaming this might be a little more clear, `getClassNamesOfHardFailures` or something along those lines

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
```diff
@@ -0,0 +1,253 @@
...
+
+    private List<AbstractCliChecker> generateChecks(List<AssetExtractionDiff> assetExtractionDiffs) {
+        CliCheckerOptions options = generateCheckerOptions();
+        return checkerList.stream().map(check -> {
+            Optional<CliCheckerType> checkEnum = Enums.getIfPresent(CliCheckerType.class, check);
+            if(checkEnum.isPresent()){
+                AbstractCliChecker checker = AbstractCliChecker.createInstanceForClassName(checkEnum.get().getClassName());
+                checker.setAssetExtractionDiffs(assetExtractionDiffs);
+                checker.setCliCheckerOptions(options);
+                return checker;
```
aurambaj, line 248: `return checkEnum.map(c ->  AbstractCliChecker ...).orElseThrow(`

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractCliChecker.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Pattern;
+
+public abstract class AbstractCliChecker {
+
+    protected CliCheckerOptions cliCheckerOptions;
+
+    protected List<AssetExtractionDiff> assetExtractionDiffs;
```
aurambaj, line 15: Q: what's the rational this being an attribute of this object? Should it be an input of the run methods? You can instantiate a checker of this type and potentially run it on many diffs.

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractCliChecker.java
```diff
@@ -0,0 +1,57 @@
...
+        this.cliCheckerOptions = options;
+    }
+
+    public void setAssetExtractionDiffs(List<AssetExtractionDiff> assetExtractionDiffs) {
+        this.assetExtractionDiffs = assetExtractionDiffs;
+    }
+
+    protected List<AssetExtractorTextUnit> getAddedTextUnits() {
+        List<AssetExtractorTextUnit> addedTextUnits = new ArrayList<>();
+        assetExtractionDiffs.stream().forEach(assetExtractionDiff -> addedTextUnits.addAll(assetExtractionDiff.getAddedTextunits()));
```
aurambaj, line 42: `stream().map(AssetExtractionDiff::getAddedTextUnits).collect()`

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractCliChecker.java
```diff
@@ -0,0 +1,57 @@
...
+        return addedTextUnits;
+    }
+
+    protected List<String> getSourceStringsFromDiff() {
+        List<String> sourceStrings = new ArrayList<>();
+        getAddedTextUnits().stream().forEach(assetExtractorTextUnit -> sourceStrings.add(assetExtractorTextUnit.getSource()));
```
aurambaj, line 48: same

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractCliChecker.java
```diff
@@ -0,0 +1,57 @@
...
+        return sourceStrings;
+    }
+
+    protected List<Pattern> getRegexPatterns() {
+        List<Pattern> patterns = new ArrayList<>();
+        cliCheckerOptions.getParameterRegexSet().stream().forEach(regex -> patterns.add(Pattern.compile(regex.getRegex())));
```
aurambaj, line 54: same

aurambaj (commented):
first pass, need to finish the checkers

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractCliChecker.java
maallen: This is a hangover from when I was running the commands in parallel and needed to implemented the Callable interface, I'll tidy it up and  pass the list of diffs to the run method

cli/src/main/java/com/box/l10n/mojito/cli/command/Command.java
```diff
@@ -116,4 +119,14 @@ public String getName() {
     public void setOriginalArgs(List<String> originalArgs) {
         this.originalArgs = originalArgs;
     }
+
+    protected String getFormattedPhabricatorMessage(String messageTemplate, String messageKey, String message){
```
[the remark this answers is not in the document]
maallen: Agree on the slack message addition, I've refactored the existing code into a PhabricatorMessageBuilder for the time being but will look into expanding it to be similar to the branch notification code.

aurambaj (commented):
what's the licence of the dictionary data?

maallen:
The dictionary is version 2017.03.01 of the libre office dictionaries (https://extensions.libreoffice.org/en/extensions/show/english-dictionaries) which is licensed under BSD. More recent dictionaries are licensed under LGPL

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/SpellCliChecker.java
```diff
@@ -0,0 +1,223 @@
...
+            if(!suggestions.isEmpty()){
+                notificationText.append("- Did you mean ");
+                notificationText.append(suggestions.stream().collect(Collectors.collectingAndThen(Collectors.toList(), joinCommaSeparated(", ", " or "))));
+                notificationText.append("?");
+            }
+            notificationText.append(System.lineSeparator());
+        });
+    }
+
+    private static Function<List<String>, String> joinCommaSeparated(String delimiter, String finalDelimiter) {
```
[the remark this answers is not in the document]
maallen: I tried upticking the icu dependency to 70.1 to utilize this which caused failures in the pull command tests for the android & mac strings types. For android it was adding an additional many plural type to the outputted files, wasn't clear what the difference was for Mac strings but I expect it's for a similar reason. I reverted the change as didn't wish to further complicate this PR by including a lib update and it's potential associated updates

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/ContextAndCommentCliChecker.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        public boolean isFailed() {
+            return failed;
+        }
+    }
+
+    @Override
+    public CliCheckResult run(List<AssetExtractionDiff> assetExtractionDiffs) {
+        CliCheckResult cliCheckResult = new CliCheckResult(isHardFail(), CliCheckerType.CONTEXT_COMMENT_CHECKER.name());
```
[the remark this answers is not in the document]
maallen: I've refactored the enum type to be similar to `FileTypes` enum and added a createCliCheckResult method to the abstract class.

maallen:
I have removed the dictionary files from the built Mojito jar, now no dictionary files are packaged and the user must provide paths to a .dict file & a .aff file on the command line

aurambaj (approved):

PR #742, opened 2021-12-02
Added Empty Placeholder, Glossary, Placeholder comment & Recommend String id cli checkers

maallen:
Added additional cli checker implementations.

**EMPTY_PLACEHOLDER_CHECKER**
Checks that there is no empty placeholders in a source string when the single or double brace placeholder regex is in use.

**GLOSSARY_CHECKER**
Checks source strings for terms specified in a glossary file, if a term is found then it needs to exactly match the glossary term. Glossary terms can be configured with either a MAJOR or MINOR severity level. MINOR failures will not fail the check, a warning will be added to the notification text. MAJOR failures fail the glossary check and an error message will be added to the notification text. The glossary file is in json format and contains an array of objects containing a term and its associated severity level, example below.

`[
  {
    "term": "Some major term",
    "severity": "MAJOR"
  },
  {
    "term": "Some minor term",
    "severity": "MINOR"
  }
]`

**PLACEHOLDER_COMMENT_CHECKER**
Verifies that any placeholders in a source string have a description in the comment explaining the placeholder. The comment must contain the description in the form _**<placeholder_name>:<placeholder_description>**_

**Recommend String Id**
Checks that string context value is prefixed with the top two directory names in the file path, if not a recommendation is made to prefix the message context value with the relevant directory names.

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractPlaceholderDescriptionCheck.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractPlaceholderDescriptionCheck.java
@@ -4,6 +4,7 @@ import org.apache.commons.lang3.StringUtils;
 
 import java.util.Optional;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 public abstract class AbstractPlaceholderDescriptionCheck {
 
@@ -12,10 +13,14 @@ public abstract class AbstractPlaceholderDescriptionCheck {
     public Optional<String> getFailureText(String placeholder) {
         String failureText = null;
         if (StringUtils.isNumeric(placeholder)) {
-            failureText = "Missing description for placeholder number '" + placeholder + "' in comment.";
+            failureText = "Missing description for placeholder number '" + placeholder + "' in comment. Please add a description in the string comment in the form " + placeholder + ":<description>";
         } else if (!placeholder.trim().isEmpty()) {
-            failureText ="Missing description for placeholder with name '" + placeholder + "' in comment.";
+            failureText = "Missing description for placeholder with name '" + placeholder + "' in comment. Please add a description in the string comment in the form " + placeholder + ":<description>";
         }
         return Optional.ofNullable(failureText);
+    }
+
+    protected boolean isPlaceholderDescriptionMissingInComment(String comment, String placeholder) {
+        return StringUtils.isBlank(comment) || !Pattern.compile(placeholder + ":.+").matcher(comment).find();
     }
 }
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/EmptyPlaceholderChecker.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/EmptyPlaceholderChecker.java
@@ -1,0 +1,76 @@
+package com.box.l10n.mojito.cli.command.checks;
+
+import com.box.l10n.mojito.cli.command.extraction.AssetExtractionDiff;
+import com.box.l10n.mojito.regex.PlaceholderRegularExpressions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * {@link AbstractCliChecker} that verifies a source string does not contain an empty placeholder e.g. '{}'.
+ *
+ * The check uses Single or Double braces regex to identify placeholders in a source string and then checks that the
+ * placeholder substring contains regex 'word' values.
+ *
+ * @author mallen
+ */
+public class EmptyPlaceholderChecker extends AbstractCliChecker {
+
+    static Logger logger = LoggerFactory.getLogger(EmptyPlaceholderChecker.class);
+
+    private Pattern wordPattern = Pattern.compile("\\w+");
+
+    @Override
+    public CliCheckResult run(List<AssetExtractionDiff> assetExtractionDiffs) {
+        CliCheckResult cliCheckResult = createCliCheckerResult();
+        Set<String> failures = checkForEmptyPlaceholders(assetExtractionDiffs);
+        if (!failures.isEmpty()) {
+            cliCheckResult.setSuccessful(false);
+            cliCheckResult.setNotificationText(buildNotificationText(failures).toString());
+        }
+
+        return cliCheckResult;
+    }
+
+    private Set<String> checkForEmptyPlaceholders(List<AssetExtractionDiff> assetExtractionDiffs) {
+        return cliCheckerOptions.getParameterRegexSet().stream()
+                .filter(regex -> isEmptyPlaceholderRegex(regex))
+                .flatMap(placeholderRegularExpressions -> getAddedTextUnits(assetExtractionDiffs).stream()
+                    .map(assetExtractorTextUnit -> assetExtractorTextUnit.getSource())
+                    .filter(source -> isSourceStringWithEmptyPlaceholders(placeholderRegularExpressions, source))).collect(Collectors.toSet());
+    }
+
+    private boolean isEmptyPlaceholderRegex(PlaceholderRegularExpressions regex) {
+        return regex.equals(PlaceholderRegularExpressions.SINGLE_BRACE_REGEX) || regex.equals(PlaceholderRegularExpressions.DOUBLE_BRACE_REGEX);
+    }
+
+    private boolean isSourceStringWithEmptyPlaceholders(PlaceholderRegularExpressions placeholderRegularExpressions, String source) {
+        Matcher matcher = Pattern.compile(placeholderRegularExpressions.getRegex()).matcher(source);
+        while (matcher.find()) {
+            String placeholder = source.substring(matcher.start(), matcher.end());
+            logger.debug("Found placeholder '{}' in source string '{}'", placeholder, source);
+            if (!wordPattern.matcher(placeholder).find()) {
+                logger.debug("Found empty placeholder '{}' in source string '{}'", placeholder, source);
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private StringBuilder buildNotificationText(Set<String> failures) {
+        StringBuilder notificationText = new StringBuilder();
+        notificationText.append("Found empty placeholders in the following source strings, please remove or update placeholders to contain a descriptive name:");
+        notificationText.append(System.lineSeparator());
+        notificationText.append(failures.stream()
+                .map(source -> "* '" + source + "'")
+                .collect(Collectors.joining(System.lineSeparator())));
+        notificationText.append(System.lineSeparator());
+        return notificationText;
+    }
+
+}
--- a/cli/src/test/java/com/box/l10n/mojito/cli/command/checks/RecommendStringIdCheckerTest.java
+++ b/cli/src/test/java/com/box/l10n/mojito/cli/command/checks/RecommendStringIdCheckerTest.java
@@ -1,0 +1,225 @@
+package com.box.l10n.mojito.cli.command.checks;
+
+import com.box.l10n.mojito.cli.command.extraction.AssetExtractionDiff;
+import com.box.l10n.mojito.okapi.extractor.AssetExtractorTextUnit;
+import com.google.common.collect.Sets;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.nio.file.FileSystems;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static com.box.l10n.mojito.regex.PlaceholderRegularExpressions.SINGLE_BRACE_REGEX;
+
+public class RecommendStringIdCheckerTest {
+
+    private RecommendStringIdChecker recommendStringIdChecker;
+
+    private List<AssetExtractionDiff> assetExtractionDiffs;
+
+    @Before
+    public void setup() {
+        recommendStringIdChecker = new RecommendStringIdChecker();
+        List<String> modifiedFiles = new ArrayList<>(Arrays.asList("someDir/someOtherDir/evenDeeperDir/test1.txt", "someDir/someOtherDir/test2.txt"));
+        recommendStringIdChecker.setCliCheckerOptions(new CliCheckerOptions(Sets.newHashSet(SINGLE_BRACE_REGEX), Sets.newHashSet(),
+                "", ""));
+        List<AssetExtractorTextUnit> addedTUs = new ArrayList<>();
+        AssetExtractorTextUnit assetExtractorTextUnit = new AssetExtractorTextUnit();
+        assetExtractorTextUnit.setName("A source string with no errors. --- someDir.someSubDir.someStringId");
+        assetExtractorTextUnit.setSource("A source string with no errors.");
+        assetExtractorTextUnit.setUsages(Sets.newHashSet("someDir/someSubDir/someSourceFile.java"));
+        addedTUs.add(assetExtractorTextUnit);
+        assetExtractionDiffs = new ArrayList<>();
+        AssetExtractionDiff assetExtractionDiff = new AssetExtractionDiff();
+        assetExtractionDiff.setAddedTextunits(addedTUs);
+        assetExtractionDiffs.add(assetExtractionDiff);
+
+    }
+
+    @Test
+    public void testSuccess() {
+        CliCheckResult result = recommendStringIdChecker.run(assetExtractionDiffs);
+        Assert.assertTrue(result.isSuccessful());
+        Assert.assertTrue(result.getNotificationText().isEmpty());
+    }
+
+    @Test
+    public void testFailure() {
+        CliCheckResult result = recommendStringIdChecker.run(setTextUnitId("incorrect.prefix.someStringId"));
+        Assert.assertFalse(result.isSuccessful());
+        Assert.assertEquals("Recommended id updates for the following strings:" + System.lineSeparator()
+                + "* Please update id for string 'A source string with no errors.' to be prefixed with 'someDir.someSubDir.'" + System.lineSeparator(), result.getNotificationText());
+    }
+
+    @Test
+    public void testMultipleUsagesProvided() {
+        List<AssetExtractorTextUnit> addedTUs = new ArrayList<>();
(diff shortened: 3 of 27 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private void checkGlossaryTerm(String source, GlossarySearchResult result, List<String> failures, Node current) {
+        String trimmedWhitespace = source.trim().replaceAll(" +", " ");
+        int index = StringUtils.indexOfIgnoreCase(trimmedWhitespace, current.glossaryTerm.getTerm());
+        String glossaryTermInString = trimmedWhitespace.substring(index, index + current.glossaryTerm.getTerm().length());
+        if(!glossaryTermInString.equals(current.glossaryTerm.getTerm())){
+            result.isSuccess = false;
+            if(current.glossaryTerm.getSeverity() == GlossaryTermSeverity.MAJOR) {
+                result.isMajorFailure = true;
+                failures.add(String.format("MAJOR: String '%s' contains glossary term '%s' which must match case exactly.", source, current.glossaryTerm.getTerm()));
```
garionpin, line 66: When doing glossary checks - are case matches the only (or main) scenarios that we need to care about?
E.g. If someone uses "Brand-Name" instead of "BrandName" - is that a concern? Assuming no, but wanted to double-check.

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/EmptyPlaceholderChecker.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * {@link AbstractCliChecker} that verifies a source string does not contain an empty placeholder e.g. '{}'.
+ *
+ * The check uses Single or Double braces regex to identify placeholders in a source string and then checks that the
+ * placeholder substring contains regex 'word' values.
```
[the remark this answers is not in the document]
maallen: I'm not sure that all non whitespace characters are valid placeholder values but it's beyond the scope of this check to verify placeholder validity. If an invalid placeholder is provided I would expect this will cause runtime errors at the source code level anyway which a dev would need to fix.

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
maallen: I've updated the implementation to catch this, using your example the glossary check will match Brand-Name as it's splitting on \W regex so the hypen is removed. This use case originally caused an index out of bounds exception, I've fixed this in the latest commit.

cli/src/test/java/com/box/l10n/mojito/cli/command/checks/GlossaryCheckerTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+    private List<AssetExtractionDiff> assetExtractionDiffs;
+
+    @Before
+    public void setup() {
+        glossaryChecker = new GlossaryChecker();
+        glossaryChecker.setCliCheckerOptions(new CliCheckerOptions(Sets.newHashSet(), Sets.newHashSet(), "", "target/test-classes/com/box/l10n/mojito/cli/glossarychecker/glossary.json"));
+        List<AssetExtractorTextUnit> addedTUs = new ArrayList<>();
+        AssetExtractorTextUnit assetExtractorTextUnit = new AssetExtractorTextUnit();
+        assetExtractorTextUnit.setSource("A source string with Pinterest in it.");
```
garionpin, line 25: Spotted a single "Pinterest" reference remaining (you might have it fixed locally already and not pushed - just wanted to double check).

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
```diff
@@ -0,0 +1,110 @@
...
+                }
+            }
+        }
+        if(allFailed) {
+            addFailureText(source, result, failures, current);
+        }
+//        int index = StringUtils.indexOfIgnoreCase(trimmedWhitespace, glossaryTerm.getTerm());
+//        if(index > -1) {
+//            String glossaryTermInString = trimmedWhitespace.substring(index, index + current.glossaryTerm.getTerm().length());
+//            if(!glossaryTermInString.equals(current.glossaryTerm.getTerm())){
```
garionpin, line 78: Dead/commented-out code

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
```diff
@@ -0,0 +1,110 @@
...
+        } else {
+            if (current.glossaryTerms.size() > 1) {
+                failures.add(String.format("WARN: String '%s' contains glossary terms %s but does not exactly match one of the terms.", source,
+                        current.glossaryTerms.stream().map(term -> "'" + term.getTerm() + "'")
+                        .collect(Collectors.joining(" or "))));
+            } else {
+                failures.add(String.format("WARN: String '%s' contains glossary term '%s' but does not exactly match the glossary term.", source, current.glossaryTerms.get(0).getTerm()));
+            }
+        }
+//        if (current.glossaryTerm.getSeverity() == GlossaryTermSeverity.MAJOR) {
```
garionpin, line 103: Dead/commented-out code

garionpin (approved):

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
```diff
@@ -0,0 +1,94 @@
...
+                    checkGlossaryTerms(source, result, failures, current);
+                }
+            }
+        }
+        result.failures = failures;
+        return result;
+    }
+
+    private void checkGlossaryTerms(String source, GlossarySearchResult result, List<String> failures, Node current) {
+        String trimmedWhitespace = source.trim().replaceAll(" +", " ");
```
aurambaj, line 61: shouldn't this be done once and for all at the very beginning of the process?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
```diff
@@ -0,0 +1,94 @@
...
+            this.add(term);
+        }
+    }
+
+    public void add(GlossaryTerm term) {
+        String[] words = term.getTerm().split("\\W+");
+        Node currentNode = rootNode;
+        for (String word : words) {
+            word = word.toLowerCase();
+            if(!currentNode.children.containsKey(word)) {
```
aurambaj, line 32: knit: indentation (just saying once but there are a lot more after)

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class GlossaryTermTrie {
+
+    private class Node {
+        Map<String, Node> children = new HashMap<>();
+        List<GlossaryTerm> glossaryTerms = new ArrayList<>();
```
aurambaj, line 16: why do we have a list here? we'll have duplicated terms? 

A leaf node is a node with an non-empty list ?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
```diff
@@ -0,0 +1,94 @@
...
+        String[] words = source.split("\\W+");
+        GlossarySearchResult result = new GlossarySearchResult(source);
+        List<String> failures = new ArrayList<>();
+        for(int i = 0; i < words.length; i++) {
+            Node current = rootNode;
+            for (int j = i; j < words.length; j++) {
+                if (!current.children.containsKey(words[j].toLowerCase())){
+                    break;
+                }
+                current = current.children.get(words[j].toLowerCase());
```
aurambaj, line 50: can the lookup be done once and check for = null to break?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
```diff
@@ -0,0 +1,94 @@
...
+                if(!current.glossaryTerms.isEmpty()){
+                    checkGlossaryTerms(source, result, failures, current);
```
aurambaj, line 52: we found a leaf/match, why do we need to re-process the whole source string in `checkGlossaryTerms`?

could we search in parallel for insensitive and senstive match in the trie, and if one succeed but not the other it is a failure then?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
```diff
@@ -0,0 +1,94 @@
...
+            }
+        }
+        result.failures = failures;
+        return result;
+    }
+
+    private void checkGlossaryTerms(String source, GlossarySearchResult result, List<String> failures, Node current) {
+        String trimmedWhitespace = source.trim().replaceAll(" +", " ");
+        boolean allFailed = true;
+        for (GlossaryTerm glossaryTerm : current.glossaryTerms) {
```
aurambaj, line 63: i'm not sure to understand in which case will have multiple glossary term for 1 node and, i'm not sure to understand why we need to search again in the string for the whole match since we've done the search using the trie before.

Is this logic to check that case in the source is the same as declared in the glossary term but we lost the info in the try?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
```diff
@@ -0,0 +1,94 @@
...
+                }
+            }
+        }
+        result.failures = failures;
+        return result;
+    }
+
+    private void checkGlossaryTerms(String source, GlossarySearchResult result, List<String> failures, Node current) {
+        String trimmedWhitespace = source.trim().replaceAll(" +", " ");
+        boolean allFailed = true;
```
aurambaj, line 62: is the meaning here is "has match"? what has failed in this case? failed to find an exact match even though there was a case insensitive match?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
maallen: It's to handle the use case where the glossary contains terms in both cases e.g."Tag" & "tag", in this example both are acceptable so we maintain a list of glossary terms on the node and then check against them. 

The trie stores it's entries in lower case, we then traverse the trie by setting our search words to lower case too, this should guarantee we get a case insensitive match. Once we have the match we then want to check that the term also matches the case, although looking at the code again I don't think I need to find the index of the term in the source string I can just check if the source string contains the glossary term, if not then we know it doesn't match the glossary

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
maallen: Yes failed to find an exact match even though there was a case insensitive match which means we have a failure

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
maallen: We can have terms in the glossary that are duplicated except for case e.g. "Tag" & "tag"

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
```diff
@@ -0,0 +1,81 @@
...
+                if(current.glossaryTerm != null){
+                    checkGlossaryTerm(source, result, failures, current);
+                }
+            }
+        }
+        result.failures = failures;
+        return result;
+    }
+
+    private void checkGlossaryTerm(String source, GlossarySearchResult result, List<String> failures, Node current) {
```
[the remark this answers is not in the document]
maallen: That looks like it checks a glossary file with translations to see if the correct translations has been used in the translated string, in our case we won't have the translations at this point. It might be useful to pull that functionality into an integrity checker on the server though when we do have translations in a separate PR?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
```diff
@@ -0,0 +1,94 @@
...
+    public GlossaryTermTrie(List<GlossaryTerm> terms) {
+        for(GlossaryTerm term: terms) {
+            this.add(term);
+        }
+    }
+
+    public void add(GlossaryTerm term) {
+        String[] words = term.getTerm().split("\\W+");
+        Node currentNode = rootNode;
+        for (String word : words) {
```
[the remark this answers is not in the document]
maallen: The majority of the terms in the glossary are 2 or 3 words long, I used a word based trie to avoid matching on terms that were a substring of a larger string e.g. in a character based trie with a glossary term 'ad' we'll find a match when traversing a source string containing "additional", with a word based trie this doesn't occur as we just search for 'additional' instead of walking the characters in the term.

cli/src/test/java/com/box/l10n/mojito/cli/command/checks/RecommendStringIdCheckerTest.java
```diff
@@ -0,0 +1,169 @@
...
+public class RecommendStringIdCheckerTest {
+
+    private RecommendStringIdChecker recommendStringIdChecker;
+
+    private List<AssetExtractionDiff> assetExtractionDiffs;
+
+    @Before
+    public void setup() {
+        recommendStringIdChecker = new RecommendStringIdChecker();
+        List<String> modifiedFiles = new ArrayList<>(Arrays.asList("target/test-classes/com/box/l10n/mojito/cli/phab-string-id-checker/someDir/someOtherDir/evenDeeperDir/test1.txt",
```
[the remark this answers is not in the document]
maallen: No I was using them for testing with absolute paths (we don't want to recommend a string id that starts with directories outside of the repo structure), the assumption is that the mojito command will be executed from the repo root dir. I've updated the code to check if a path is absolute and if so strip it to the repository root.

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
maallen: The trie uses lower case words as it's key so that we can match words in the source string by setting them to lower case, I don't think we can search in parallel for both cases as we can't predict what case the source string will contain so need to standardise them both to lower case. 

Once we find a match then we need to extract the string as is from the source string and check that it matches the glossary term exactly

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
aurambaj: I was sensing that but then I was a little bit puzzle with the use case then. Since we basically want to enforce a certain case (upper case Tag) then having both is sort of pointless, no? That'd allow tag and Tag and refuse TAg but is that not relevant to check, I mean we could do that for any string then. 

Wondering if that should just be called a "CaseChecker", we're not really checking much from the glossary point of view. I guess more classic glossary check is for example to make sure that the translation uses the translation specified in the glossary but for source only analysis, here we just check casing, right?

Let's discuss more offline.

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
```diff
@@ -132,6 +132,11 @@ protected void execute() throws CommandException {
         consoleWriter.fg(Ansi.Color.GREEN).newLine().a("Checks completed").println(2);
     }
 
+    private boolean isPhabricatorSkipChecks() {
```
aurambaj, line 135: knit: not used?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractPlaceholderDescriptionCheck.java
```diff
@@ -12,10 +13,14 @@
     public Optional<String> getFailureText(String placeholder) {
         String failureText = null;
         if (StringUtils.isNumeric(placeholder)) {
-            failureText = "Missing description for placeholder number '" + placeholder + "' in comment.";
+            failureText = "Missing description for placeholder number '" + placeholder + "' in comment. Please add a description in the string comment in the form " + placeholder + ":<description>";
         } else if (!placeholder.trim().isEmpty()) {
-            failureText ="Missing description for placeholder with name '" + placeholder + "' in comment.";
+            failureText ="Missing description for placeholder with name '" + placeholder + "' in comment. Please add a description in the string comment in the form " + placeholder + ":<description>";
```
aurambaj, line 18: knit: identation

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermTrie.java
```diff
@@ -0,0 +1,92 @@
...
+            word = word.toLowerCase();
+            if (!currentNode.children.containsKey(word)) {
+                currentNode.children.put(word, new Node());
+            }
+            currentNode = currentNode.children.get(word);
+        }
+        currentNode.glossaryTerms.add(term);
+    }
+
+    public GlossaryCaseCheckerSearchResult runGlossaryCheck(String source) {
```
aurambaj, line 34: probably want to mention in the class name (and related classes) and function name that it is a trie for the "case checking" too?

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
maallen: Decided to move this implementation to a different PR and this must've slipped into a later commit on this branch

aurambaj (approved):

PR #743, opened 2021-12-02
Fix DropImportCommand flaky test

garionpin:
Fix DropImportCommand flaky test by ensuring the wait condition is
applied to exportable locales only instead of all locales, by reusing
the shouldCreateDrop method.


aurambaj:
🤞

PR #745, opened 2021-12-08
Address multiple statistics mapping to the same text unit

garionpin:
Address multiple statistics mapping to the same text unit by ensuring
we only retrieve TextUnits by ID once and that we merge the information
from those multiple statistics data points into one consolidated data
point for the DB TextUnit.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/tm/TMTextUnitStatisticService.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/tm/TMTextUnitStatisticService.java
@@ -21,6 +21,7 @@ import com.google.common.collect.Iterables;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tags;
 import org.apache.commons.lang3.StringUtils;
+import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/tm/TMTextUnitStatisticService.java
```diff
@@ -198,7 +200,22 @@ private void uploadStatistics(Map <ImportTextUnitStatisticsBody, List<TMTextUnit
...
+                .collect(Collectors.collectingAndThen(
+                        Collectors.toMap(TMTextUnitStatistic::getTMTextUnit, Function.identity(), (stat1, stat2) -> {
+                            stat1.setLastDayUsageCount(stat1.getLastDayUsageCount() + stat2.getLastDayUsageCount());
+                            stat1.setLastPeriodUsageCount(stat1.getLastPeriodUsageCount() + stat2.getLastPeriodUsageCount());
+                            DateTime stat1date = stat1.getLastSeenDate();
+                            DateTime stat2date = stat2.getLastSeenDate();
+                            DateTime maxDate = stat1date.compareTo(stat2date) >= 0 ? stat1date : stat2date;
+                            stat1.setLastSeenDate(maxDate);
+
+                            logger.warn("Merged statistics information for TextUnit with id: {} , name: {}",
```
aurambaj, line 212: warn? is this something that is not expected to happen regularly?

aurambaj (approved):

webapp/src/main/java/com/box/l10n/mojito/service/tm/TMTextUnitStatisticService.java
garionpin: _Ideally_ we wouldn't want this to happen regularly - as the hope would be that the statistics data being sent in maps 1:1 or 1:N to units - without having two _different_ statistics mapping back to the same TextUnit.
Emitting a warning as this implies the data could be a bit less accurate for those units - and would allow us to trace it back if ever needed.

PR #750, opened 2021-12-09
Added Phabricator override for extraction-check command

maallen:
Allows extraction-checks to be disabled by adding '**SKIP_I18N_CHECKS**' to the relevant Phabricator revision test plan string.


garionpin (approved):

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
```diff
@@ -41,6 +42,10 @@
 
     static Logger logger = LoggerFactory.getLogger(ExtractionDiffCommand.class);
 
+    protected static final String PHAB_OVERRIDE_FLAG = "SKIP_I18N_CHECKS";
+    protected static final String PHAB_OVERRIDE_MESSAGE = PhabricatorIcon.WARNING + " **WARNING**" + System.lineSeparator() + System.lineSeparator() + "I18N source string checks have been disabled by override, " +
```
aurambaj, line 46: Can we have a generic way of handling notifications? Ideally, `ExtractionCheckCommand` should be generic and not have specific logic tied to a specific notification system. Those new parameters are becoming very specific. It could be gathered in PhabricatorDiffInfoCommand and then just have a "skip" flag here when the command is made more generic.

Since we already had some Phabricator logic in there maybe we should rename the command to make it a "Phabricator" command, similar to "PhabricatorExtractionDiffNotificationCommand"

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
```diff
@@ -41,6 +42,10 @@
 
     static Logger logger = LoggerFactory.getLogger(ExtractionDiffCommand.class);
 
+    protected static final String PHAB_OVERRIDE_FLAG = "SKIP_I18N_CHECKS";
```
aurambaj, line 45: knit: should the constant renamed to something like `SKIP_I18N_CHECKS_FLAG`? I'm not super clear about the "override" meaning. and then maybe `isOverrideActive()` is `areChecksSkipped()`?

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
maallen: I've renamed the ExtractionCheck command to PhabricatorExtractionCheck, cli call to phab-extraction-check. I'll update the notifications to be more generic in a future PR.

aurambaj (approved):

PR #754, opened 2022-01-06
Cleaned up checker parameters and options and added label ignore functionality to RecommendStringIdChecker

maallen:
Added functionality in RecommendStringIdChecker to ignore a 'label' string using a regex in context message string ids.

Removed specific checker options as command parameters and instead provide them as a list of options to the command which get converted to an ImmutableMap before being passed to the commands in the CliCheckerOptions object.


cli/src/main/java/com/box/l10n/mojito/cli/command/PhabricatorExtractionCheckCommand.java
```diff
@@ -100,8 +91,8 @@
     @Parameter(names = {"--input-directory", "-i"}, arity = 1, required = false, description = ExtractionDiffCommand.INPUT_DIRECTORY_DESCRIPTION)
     String inputDirectoryParam = ExtractionPaths.DEFAULT_OUTPUT_DIRECTORY;
 
-    @Parameter(names = {"--context-comment-reject-pattern", "-ccrp"}, arity = 1, required = false, description = "Regex pattern that is used to identify context or comment strings that should be rejected.")
-    String contextCommentRejectPattern = null;
+    @Parameter(names = {"--checker-options", "-co"}, variableArity = true, required = false, listConverter = CheckerOptionsConverter.class, description = "Map of options to be used in individual checkers. Options must be in the form 'name':'value'.")
+    List<Map<String, String>> checkerOptions = new ArrayList<>();
```
garionpin, line 95: Given that we have a list option:value pairs - would a simple Map<String, String> work here to contain all of the values - or do we need multiple separate maps?
E.g.: could CheckerOptionsConverter return a MapEntry instead of a whole Map with only one value?
There might not be a cleaner to do that - just checking. (I saw someone else enquiring about similar functionality at some point: https://github.com/cbeust/jcommander/issues/218 )

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/CliCheckerOptions.java
```diff
@@ -40,23 +33,31 @@ public CliCheckerOptions(Set<PlaceholderRegularExpressions> parameterRegexSet, S
...
     }
 
     public String getGlossaryFilePath() {
-        return glossaryFilePath;
+        return optionsMap.get(GLOSSARY_FILE_PATH_KEY);
     }
 
     public String getDictionaryFilePath() {
-        return dictionaryFilePath;
+        return optionsMap.get(DICTIONARY_FILE_PATH_KEY);
```
garionpin, line 44: Given that all of these values are now stored into a dict and there's wrapper methods to get all of the right values - would a strongly typed class with those properties be a better fit here in terms of validation?

That would make CheckerOptionsConverter a bit more involved - but it would also allow us to check that no bogus values are provided (e.g.: if someone misspells a key name) or no important keys are missing.

cli/src/main/java/com/box/l10n/mojito/cli/command/PhabricatorExtractionCheckCommand.java
maallen: I think I should be able to clean it up to a custom map entry object and then map the stream of those objects into a Map (I need them in a map so that I can use the putAll method of ImmutableMap)

garionpin (approved):

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/CliCheckerTypeConverter.java
```diff
@@ -0,0 +1,16 @@
+package com.box.l10n.mojito.cli.command.checks;
+
+import com.beust.jcommander.IStringConverter;
+import com.box.l10n.mojito.cli.command.CommandException;
+
+public class CliCheckerTypeConverter implements IStringConverter<CliCheckerType> {
```
aurambaj, line 6: Can `EnumConverter` be re-used?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/PlaceholderRegularExpressionConverter.java
```diff
@@ -0,0 +1,17 @@
+package com.box.l10n.mojito.cli.command.checks;
+
+import com.beust.jcommander.IStringConverter;
+import com.box.l10n.mojito.cli.command.CommandException;
+import com.box.l10n.mojito.regex.PlaceholderRegularExpressions;
+
+public class PlaceholderRegularExpressionConverter implements IStringConverter<PlaceholderRegularExpressions> {
```
aurambaj, line 7: Can EnumConverter be re-used?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/RecommendStringIdChecker.java
```diff
@@ -70,9 +68,8 @@ private String removeLineNumberFromUsage(String usage) {
 
     private RecommendStringIdCheckResult generateRecommendation(AssetExtractorTextUnit textUnit, RecommendStringIdCheckResult recommendation) {
         List<String> possiblePrefixes = new ArrayList<>();
-        String msgctxt = textUnit.getName().contains("---") ? textUnit.getName().split("---")[1].trim() : "";
+        String msgctxt = removeIgnoredLabelFromContext(textUnit.getName().contains("---") ? textUnit.getName().split("---")[1] : "");
```
aurambaj, line 71: use constant for `---` ?

aurambaj (approved):

PR #756, opened 2022-01-17
Update PO plural rules used for Hebrew

maallen:
Updated POPluralRules so that plural forms for Hebrew match CLDR rules specified [here](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html)

### DIFF
--- a/common/src/main/java/com/box/l10n/mojito/po/PoPluralRule.java
+++ b/common/src/main/java/com/box/l10n/mojito/po/PoPluralRule.java
@@ -69,6 +69,10 @@ public enum PoPluralRule {
             "nplurals=4; plural=n%100==1 ? 0 : n%100==2 ? 1 : n%100==3 || n%100==4 ? 2 : 3;",
             PoFormToCLDRForm.THREE_FORMS_ONE_FEW_MANY,
             CldrFormsToCopyOnImport.FEW_TO_OTHER),
+    FOUR_FORMS_FRACTIONAL_DIGITS_OTHER(
+            "nplurals=4; plural=(n == 1 && n % 1 == 0) ? 0 : (n == 2 && n % 1 == 0) ? 1: (n % 10 == 0 && n % 1 == 0 && n > 10) ? 2 : 3;",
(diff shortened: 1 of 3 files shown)


garionpin (approved):

common/src/main/java/com/box/l10n/mojito/po/PoPluralRule.java
```diff
@@ -97,6 +101,8 @@
         mappingForNonDefault.put("sk", PoPluralRule.THREE_FORMS_SPECIAL_FOR_1_2_3_4);
         mappingForNonDefault.put("pl", PoPluralRule.THREE_FORMS_SPECIAL_FOR_1_AND_ENDING_234);
         mappingForNonDefault.put("sl", PoPluralRule.THREE_FORMS_SPECIAL_ENDING_020310_AND_1199); // incompatible with CLDR?
+        mappingForNonDefault.put("he", PoPluralRule.FOUR_FORMS_FRACTIONAL_DIGITS_OTHER);
+        mappingForNonDefault.put("iw", PoPluralRule.FOUR_FORMS_FRACTIONAL_DIGITS_OTHER);
```
maallen, line 105: Added "iw" as locales containing regions e.g. 'he-IL', get converted to 'iw-IL' code which then uses the default plural rules unless it's handled in this map

common/src/main/java/com/box/l10n/mojito/po/PoPluralRule.java
```diff
@@ -191,6 +197,12 @@ private CldrFormsToCopyOnImport(ImmutableMultimap<String, String> formMap) {
                 "1", "two",
                 "2", "few"
         )),
+        FOUR_FORMS_ONE_TWO_MANY_OTHER(ImmutableBiMap.of(
+                "0", "one",
```
aurambaj, line 201: Can you put a note that this diverge from what's in https://www.gnu.org/savannah-checkouts/gnu/gettext/manual/html_node/Plural-forms.html (see doc at the very top). 

I think there are other languages that have diverged from CLDR but I kept it like in the GNU doc, I "think" for backward compatibility.

aurambaj (approved):

PR #770, opened 2022-02-24
Added marker tag for quoting style in check notification text

maallen:
Added quote marker for extraction checks which gets replaced by the notification service specific quote marker by the relevant notify service sender.

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
@@ -40,6 +40,8 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static com.box.l10n.mojito.cli.command.extractioncheck.ExtractionCheckNotificationSender.QUOTE_MARKER;
+
 /**
  * Command to execute checks against any new source strings
  */
@@ -207,7 +209,7 @@ public class ExtractionCheckCommand extends Command {
     private void checkForHardFail(List<CliCheckResult> results) {
         if (getCheckerHardFailures(results).count() > 0) {
             String cliFailureString = "The following checks had hard failures:" + System.lineSeparator() +
-                    getCheckerHardFailures(results).map(failure -> failure.getCheckName() + " failures: " + System.lineSeparator() + failure.getNotificationText() + System.lineSeparator())
+                    getCheckerHardFailures(results).map(failure -> failure.getCheckName() + " failures: " + System.lineSeparator() + failure.getNotificationText().replaceAll(QUOTE_MARKER, "'") + System.lineSeparator())
                             .collect(Collectors.joining(System.lineSeparator()));
 
             logger.debug(cliFailureString);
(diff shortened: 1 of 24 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermCaseCheckerTrie.java
```diff
@@ -8,6 +8,8 @@
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import static com.box.l10n.mojito.cli.command.extractioncheck.ExtractionCheckNotificationSender.QUOTE_MARKER;
```
aurambaj, line 11: Wondering if we should call that "monospaced" (or inline code) in that case since it is the back tick. Trying to find the common ground 
- https://slack.com/help/articles/202288908-Format-your-messages
- https://secure.phabricator.com/book/phabricator/article/remarkup/

"Quote" or "Block code" is `>` in both, so `QUOTE_MARKER` would be kept it for that.

Wondering if in that case we could have even use the back-tick directly without substitution? I guess that does not work for  some other like bolding that don't share the exact same code hence doing that? Eye balling isn't that happening for the list?

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
```diff
@@ -239,7 +241,8 @@ private void outputFailuresToCommandLine(List<CliCheckResult> failedCheckNames)
         consoleWriter.fg(Ansi.Color.YELLOW).newLine().a("Failed checks: ").println();
         failedCheckNames.stream().forEach(check -> {
             consoleWriter.fg(Ansi.Color.YELLOW).newLine().a(check.getCheckName()).println();
-            consoleWriter.fg(Ansi.Color.YELLOW).newLine().a(check.getNotificationText().replaceAll("\\*", "\t*")).println();
+            consoleWriter.fg(Ansi.Color.YELLOW).newLine().a(check.getNotificationText().replaceAll("\\*", "\t*")
```
aurambaj, line 244: or should we use a marker here to then? looks like you're changing that as well. 

also introduce a function for the replacement? Should we consider some small helper class that do the mapping, marker to console, phab, slack?

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
```diff
@@ -207,7 +209,7 @@ private void sendFailureNotifications(List<CliCheckResult> failures, boolean har
     private void checkForHardFail(List<CliCheckResult> results) {
         if (getCheckerHardFailures(results).count() > 0) {
             String cliFailureString = "The following checks had hard failures:" + System.lineSeparator() +
-                    getCheckerHardFailures(results).map(failure -> failure.getCheckName() + " failures: " + System.lineSeparator() + failure.getNotificationText() + System.lineSeparator())
+                    getCheckerHardFailures(results).map(failure -> failure.getCheckName() + " failures: " + System.lineSeparator() + failure.getNotificationText().replaceAll(QUOTE_MARKER, "'") + System.lineSeparator())
```
aurambaj, line 212: keeping back-tick might be more readable actually. At least in Phabricator, I found the quote misleading when the English string had one, which prompted to ask for using the code block. 

Using function too?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/GlossaryTermCaseCheckerTrie.java
maallen: The individual notification service senders are applying the bolding in the message at present but the failure text from the checkers themselves has no bolding. 

We could use the backtick without substitution, if we do use a new notifier service in future that doesn't use the backtick that sender could do the replacement.

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionCheckCommand.java
maallen: I'm replacing the static marker I've added with a quotation here but I think the backtick may make more sense as a wrapper on the console as the source string may contain a quote as you mention in following comment so no need for the replacement

PR #771, opened 2022-02-24
Pull option to format untranslated strings with a custom pattern

garionpin:
Option for the pull command to mark untranslated strings with a user
specified pattern, when untranslated strings are meant to be included
in the pulled file.

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
@@ -87,6 +87,9 @@ public class PullCommand extends Command {
     @Parameter(names = {"--async-ws"}, required = false, description = "Use async WS, use for processing big files (should become default eventually)")
     Boolean asyncWS = false;
 
+    @Parameter(names = {"--untranslated-format"}, required = false, hidden = true, description = "Format for untranslated strings. The expected format for the pattern is: \"prefix{source}suffix\". In order to have the source string displayed, the string: {source} needs to be present in the pattern.")
+    String useParentUntranslatedPatternParam = null;
+
     @Autowired
     AssetClient assetClient;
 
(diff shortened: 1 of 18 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/tm/TranslatorWithInheritance.java
```diff
@@ -89,7 +92,13 @@ public String getTranslationFromTextUnitDTO(TextUnitDTO textUnitDTO, String sour
             translation = textUnitDTO.getTarget();
         } else if (InheritanceMode.USE_PARENT.equals(inheritanceMode)) {
             logger.debug("No TextUnitDTO, fallback to source");
+
             translation = source;
+
+            if(!Strings.isNullOrEmpty(untranslatedStringsFormat))
```
aurambaj, line 98: wondering if we should have the parameter/command name being more tight to "USE_PARENT" like use `useParentUntranslatedPattern`?

If we have a more independent param (which might be fine), that logic should be done regardless of the inheritance mode, I guess. It would make little sense with the `REMOVE_UNTRANSLATED` though and we don't have any other  mode for now. Or we could also have a special inheritance mode for it?

Leaning toward just having that param as a config of USE_PARENT like your current code but renaming.

Knit: I'd change the if statement to have the else close do `translation = source` 

Thoughts?

cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
```diff
@@ -87,6 +87,9 @@
     @Parameter(names = {"--async-ws"}, required = false, description = "Use async WS, use for processing big files (should become default eventually)")
     Boolean asyncWS = false;
 
+    @Parameter(names = {"--untranslated-format"}, required = false, hidden = true, description = "Format for untranslated strings. The expected format for the pattern is: \"prefix{source}suffix\". In order to have the source string displayed, the string: {source} needs to be present in the pattern.")
```
aurambaj, line 90: `--use-parent-untranslated-format` ? or you kept it short on purpose?

aurambaj (approved):

cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
garionpin: Naming fix: https://github.com/box/mojito/pull/772

PR #775, opened 2022-03-03
Added Smartling glossary sync functionality

maallen:
Adds functionality to sync a Smartling glossary to a Mojito repository. The Smartling glossary name is used as the asset path, when executing a PULL_SOURCE action the text units are pulled from Smartling and any existing text units in Mojito are deleted before importing the updates.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartlingGlossary.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartlingGlossary.java
@@ -1,0 +1,300 @@
+package com.box.l10n.mojito.service.thirdparty;
+
+import com.box.l10n.mojito.entity.Asset;
+import com.box.l10n.mojito.entity.AssetTextUnit;
+import com.box.l10n.mojito.entity.Repository;
+import com.box.l10n.mojito.entity.RepositoryLocale;
+import com.box.l10n.mojito.service.asset.AssetRepository;
+import com.box.l10n.mojito.service.asset.VirtualAsset;
+import com.box.l10n.mojito.service.asset.VirtualAssetBadRequestException;
+import com.box.l10n.mojito.service.asset.VirtualAssetRequiredException;
+import com.box.l10n.mojito.service.asset.VirtualAssetService;
+import com.box.l10n.mojito.service.asset.VirtualAssetTextUnit;
+import com.box.l10n.mojito.service.asset.VirutalAssetMissingTextUnitException;
+import com.box.l10n.mojito.service.assetTextUnit.AssetTextUnitRepository;
+import com.box.l10n.mojito.service.thirdparty.smartling.SmartlingTBXReader;
+import com.box.l10n.mojito.smartling.SmartlingClient;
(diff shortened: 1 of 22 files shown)


garionpin (approved):

webapp/src/main/resources/public/js/utils/LinkHelper.js
```diff
@@ -14,7 +14,7 @@ class LinkHelper {
         let url = TemplateHelper.renderTemplate(urlTemplate, paramsWithEncoded);
 
         if (url) {
-            linkOrLabel = <a href={url}>{label}</a>;
+            linkOrLabel = <a href={url} target="_blank">{label}</a>;
```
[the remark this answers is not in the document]
aurambaj: knit: not sure about this change, i'd just keep the default behavior and expect people to use shift+click if they want the new tab. In any case, this could be a different PR.

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartlingGlossary.java
```diff
@@ -0,0 +1,300 @@
...
+
+    private VirtualAsset getVirtualAsset(Repository repository, String glossaryName) {
+        VirtualAsset virtualAsset = new VirtualAsset();
+        virtualAsset.setRepositoryId(repository.getId());
+        virtualAsset.setPath(glossaryName);
+        virtualAsset.setDeleted(false);
+        try {
+            virtualAsset = virtualAssetService.createOrUpdateVirtualAsset(virtualAsset);
+        } catch (VirtualAssetBadRequestException e) {
+            throw new ThirdPartyTMSGlossarySyncException(e.getMessage(), e);
```
aurambaj, line 157: what would be the error here?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartlingGlossary.java
```diff
@@ -0,0 +1,300 @@
...
+    @Autowired
+    AssetRepository assetRepository;
+
+    @Autowired
+    AssetTextUnitRepository assetTextUnitRepository;
+
+    @Value("${l10n.smartling.accountId}")
+    String accountId;
+
+    public void pullSourceTextUnitsFromGlossary(Repository repository, String glossaryUID, Map<String, String> localeMapping) {
```
aurambaj, line 70: knit: `pullSourceTextUnits`?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartlingGlossary.java
maallen: Errors are around invalid repository id e.g. "A repository id must be provided", "A valid repository must be provided when creating an asset"

webapp/src/main/resources/public/js/utils/LinkHelper.js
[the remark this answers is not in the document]
maallen: Removed the target here and left it as it was previously

aurambaj (approved):

PR #778, opened 2022-03-07
Control character checker

maallen:
Adds a new extraction check that scans a source string for the presence of control characters. If found the check fails and the error outputted to the user indicates the index in the string that caused the failure.

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/ControlCharacterChecker.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/ControlCharacterChecker.java
@@ -1,0 +1,65 @@
+package com.box.l10n.mojito.cli.command.checks;
+
+import com.box.l10n.mojito.cli.command.extraction.AssetExtractionDiff;
+import com.google.common.base.CharMatcher;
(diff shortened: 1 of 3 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/checks/ControlCharacterChecker.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+        return result;
+    }
+
+    private ControlCharacterCheckerResult getControlCharacterCheckerResult(String source) {
+        ControlCharacterCheckerResult result = new ControlCharacterCheckerResult();
+        char[] characters = source.toCharArray();
+        List<Integer> indexMatches = new ArrayList<>();
+        for (int i = 0; i < characters.length; i++) {
+            if (Character.isISOControl(characters[i])) {
```
garionpin, line 38: Certain control characters should likely be excluded from this check, e.g.: CR, LF (and I remember seeing another one excluded in one of Jean's PRs - but can't remember which one).

garionpin (approved):

PR #784, opened 2022-03-23
Add configurable retry for SmartlingClient and re-enable exponential …

garionpin:
…backoff.

SmartlingClient retry is now configurable and the retry policy has
been re-enabled, minDuration set to 1.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
@@ -36,13 +36,10 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Component;
 import org.xml.sax.SAXException;
 import reactor.core.publisher.Mono;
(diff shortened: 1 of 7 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -54,9 +51,7 @@
 import java.util.stream.StreamSupport;
 
 import static com.box.l10n.mojito.android.strings.AndroidPluralQuantity.MANY;
-import static com.box.l10n.mojito.service.thirdparty.smartling.SmartlingFileUtils.getOutputSourceFile;
-import static com.box.l10n.mojito.service.thirdparty.smartling.SmartlingFileUtils.getOutputTargetFile;
-import static com.box.l10n.mojito.service.thirdparty.smartling.SmartlingFileUtils.isPluralFile;
+import static com.box.l10n.mojito.service.thirdparty.smartling.SmartlingFileUtils.*;
```
maallen, line 54: We should remove the wildcard import here

maallen (approved):

PR #787, opened 2022-04-05
Add Machine Translation API with Microsoft API engine implementation

garionpin:
New MT API added with NoOp and Microsoft API implementation.
Has source leveraging enabled by default with Spring caching enabled.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/rest/machinetranslation/BatchTranslationRequestDTO.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/rest/machinetranslation/BatchTranslationRequestDTO.java
@@ -1,0 +1,74 @@
+package com.box.l10n.mojito.rest.machinetranslation;
+
+import java.util.List;
+
+/**
+ * DTO to batch translate a set of source strings against a set of target languages.
+ *
+ * @author garion
+ */
+public class BatchTranslationRequestDTO {
+    List<String> textSources;
+    String sourceBcp47Tag;
+    List<String> targetBcp47Tags;
+    boolean skipFunctionalProtection;
+    boolean skipLeveraging;
+    List<Long> repositoryIds;
+    List<String> repositoryNames;
+
+    public List<String> getTextSources() {
+        return textSources;
+    }
+
+    public void setTextSources(List<String> textSources) {
+        this.textSources = textSources;
+    }
+
+    public String getSourceBcp47Tag() {
+        return sourceBcp47Tag;
+    }
+
+    public void setSourceBcp47Tag(String sourceBcp47Tag) {
+        this.sourceBcp47Tag = sourceBcp47Tag;
+    }
+
+    public List<String> getTargetBcp47Tags() {
+        return targetBcp47Tags;
+    }
+
+    public void setTargetBcp47Tags(List<String> targetBcp47Tags) {
+        this.targetBcp47Tags = targetBcp47Tags;
+    }
+
+    public boolean isSkipFunctionalProtection() {
+        return skipFunctionalProtection;
+    }
+
+    public void setSkipFunctionalProtection(boolean skipFunctionalProtection) {
+        this.skipFunctionalProtection = skipFunctionalProtection;
+    }
+
+    public boolean isSkipLeveraging() {
+        return skipLeveraging;
+    }
+
+    public void setSkipLeveraging(boolean skipLeveraging) {
+        this.skipLeveraging = skipLeveraging;
+    }
+
+    public List<Long> getRepositoryIds() {
+        return repositoryIds;
+    }
+
+    public void setRepositoryIds(List<Long> repositoryIds) {
+        this.repositoryIds = repositoryIds;
+    }
+
+    public List<String> getRepositoryNames() {
+        return repositoryNames;
+    }
+
+    public void setRepositoryNames(List<String> repositoryNames) {
+        this.repositoryNames = repositoryNames;
+    }
+}
--- a/webapp/src/main/java/com/box/l10n/mojito/rest/machinetranslation/TranslationRequestDTO.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/rest/machinetranslation/TranslationRequestDTO.java
@@ -1,0 +1,74 @@
+package com.box.l10n.mojito.rest.machinetranslation;
+
+import java.util.List;
+
+/**
+ * DTO to translate one source string for one target language.
+ *
+ * @author garion
+ */
+public class TranslationRequestDTO {
+    String textSource;
+    String sourceBcp47Tag;
+    String targetBcp47Tag;
+    boolean skipFunctionalProtection;
+    boolean skipLeveraging;
+    List<Long> repositoryIds;
+    List<String> repositoryNames;
+
+    public String getTextSource() {
+        return textSource;
+    }
+
+    public void setTextSource(String textSource) {
+        this.textSource = textSource;
+    }
+
+    public String getSourceBcp47Tag() {
+        return sourceBcp47Tag;
+    }
+
+    public void setSourceBcp47Tag(String sourceBcp47Tag) {
+        this.sourceBcp47Tag = sourceBcp47Tag;
+    }
+
+    public String getTargetBcp47Tag() {
+        return targetBcp47Tag;
+    }
+
+    public void setTargetBcp47Tag(String targetBcp47Tag) {
+        this.targetBcp47Tag = targetBcp47Tag;
+    }
+
+    public boolean isSkipFunctionalProtection() {
+        return skipFunctionalProtection;
+    }
+
+    public void setSkipFunctionalProtection(boolean skipFunctionalProtection) {
+        this.skipFunctionalProtection = skipFunctionalProtection;
+    }
+
+    public boolean isSkipLeveraging() {
+        return skipLeveraging;
+    }
+
+    public void setSkipLeveraging(boolean skipLeveraging) {
+        this.skipLeveraging = skipLeveraging;
+    }
+
+    public List<Long> getRepositoryIds() {
+        return repositoryIds;
+    }
+
+    public void setRepositoryIds(List<Long> repositoryIds) {
+        this.repositoryIds = repositoryIds;
+    }
+
+    public List<String> getRepositoryNames() {
+        return repositoryNames;
+    }
+
+    public void setRepositoryNames(List<String> repositoryNames) {
+        this.repositoryNames = repositoryNames;
+    }
+}
--- a/webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/MachineTranslationService.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/MachineTranslationService.java
@@ -1,0 +1,249 @@
+package com.box.l10n.mojito.service.machinetranslation;
+
+import com.box.l10n.mojito.entity.TMTextUnit;
+import com.box.l10n.mojito.service.leveraging.LeveragerByContent;
+import com.box.l10n.mojito.service.leveraging.LeveragerByContentAndRepository;
+import com.box.l10n.mojito.service.tm.search.TextUnitDTO;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import io.micrometer.core.annotation.Timed;
+import org.apache.logging.log4j.util.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import java.util.AbstractMap;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BinaryOperator;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Service to apply leveraging and machine translation on a set of provided strings.
+ *
+ * @author garion
+ */
+@Component
+public class MachineTranslationService {
+    static Logger logger = LoggerFactory.getLogger(MachineTranslationService.class);
+
+    public static final String DEFAULT_LOCALE = "en";
+
+    final MachineTranslationEngine machineTranslationEngine;
+
+    final TranslationMerger translationMerger;
+
+    public MachineTranslationService(
+            MachineTranslationEngine machineTranslationEngine, TranslationMerger translationMerger) {
+        this.machineTranslationEngine = machineTranslationEngine;
+        this.translationMerger = translationMerger;
+    }
+
+    public TranslationsResponseDTO getTranslations(
+            List<String> textSources,
+            String sourceBcp47Tag,
+            List<String> targetBcp47Tags,
+            boolean skipFunctionalProtection,
+            boolean skipLeveraging,
+            List<Long> repositoryIds,
+            List<String> repositoryNames
+    ) {
+        ImmutableMap<String, ImmutableList<TranslationDTO>> translationsBySourceText = ImmutableMap.of();
+
+        // De-duplicate input strings
+        List<String> untranslatedStrings = textSources.stream().distinct().collect(Collectors.toList());
+
+        if (!skipLeveraging) {
+            ImmutableMap<String, ImmutableList<TranslationDTO>> leveragedTranslationsBySourceText =
+                    getLeveragedTranslationsBySourceText(textSources, targetBcp47Tags, repositoryIds, repositoryNames);
+            translationsBySourceText = leveragedTranslationsBySourceText;
+
+            // Only MT strings for which we didn't leverage translationsBySourceText for all the expected target languages.
+            // Note: this would benefit from further optimization at some point to take into account partially leveraged
+            // text units - though that would require additional MT API calls for certain MT engine implementations.
+            if (!leveragedTranslationsBySourceText.isEmpty()) {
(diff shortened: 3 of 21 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/MachineTranslationService.java
```diff
@@ -0,0 +1,165 @@
...
+                    }
+                }
+            } else {
+                logger.debug("No TMTextUnit Match found for content: {}", sourceText);
+            }
+        }
+        return translationsResponseDTO;
+    }
+
+    @SuppressWarnings("SpringCacheableComponentsInspection")
```
aurambaj, line 157: why is that needed?

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/MachineTranslationService.java
```diff
@@ -0,0 +1,165 @@
...
+    @Cacheable("machineTranslation")
+    List<TextUnitDTO> getLeveragedTextUnitDTOS(String sourceText) {
+        TMTextUnit searchTmTextUnit = new TMTextUnit();
+        searchTmTextUnit.setContent(sourceText);
+
+        return leveragerByContent.getLeveragingMatches(searchTmTextUnit, null, null);
```
aurambaj, line 163: is that fine in term of performance?

garionpin: Performance isn't great as that API returns more results than desired generally and the query isn't solely optimized for this purpose, which is why I've included a skipLeveraging parameter - though it is usable if it's not on a critical path - and especially fast on subsequent calls with the in-memory caching being enabled on getLeveragedTextUnitDTOS().

Ideally we'd do this operation in bulk for multiple sources and only return the latest/best match per language, as per the comment from line 122.

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/microsoft/MicrosoftMTEngine.java
```diff
@@ -0,0 +1,190 @@
...
+            throw e;
+        }
+    }
+
+    private TranslationsResponse getTranslationsResponseDTO(
+            List<String> textSources, MicrosoftTextTranslationDTO[] response, List<String> targetLocaleTags) {
+        List<TextUnitTranslationGroup> textUnitTranslationGroups = new ArrayList<>();
+
+        // Translation objects in the output for each source string are returned in the same order as provided by the client.
+        for (int i = 0; i < textSources.size(); i++) {
```
aurambaj, line 148: nit: can probably do soemthing like Streams.zip(textSources.stream(), response.stream(), (ts, mt) -> ... ).collect(toList())

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/microsoft/MicrosoftMTEngine.java
```diff
@@ -0,0 +1,190 @@
...
+        // Within a set of translations for a specific source string, the languages are ordered in the same way as
+        // the order of target languages provided to the API.
+        for (int j = 0; j < targetLocaleTags.size(); j++) {
+            MicrosoftLanguageTranslationDTO microsoftLanguageTranslationDTO = microsoftTextTranslationDTO.getTranslations().get(j);
+            String requestTargetLocale = targetLocaleTags.get(j);
+
+            Translation translationDTO = new Translation();
+            translationDTO.setTranslationSource(getSource());
+            translationDTO.setLocaleTag(requestTargetLocale);
+            translationDTO.setText(microsoftLanguageTranslationDTO.getText());
```
aurambaj, line 172: nit: I see potential stream everywhere now 😭

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/microsoft/MicrosoftMTEngine.java
```diff
@@ -0,0 +1,190 @@
...
+    public static final String OCP_APIM_SUBSCRIPTION_REGION = "Ocp-Apim-Subscription-Region";
+    public static final String X_CLIENT_TRACE_ID = "X-ClientTraceId";
+
+    static Logger logger = LoggerFactory.getLogger(MicrosoftMTEngine.class);
+
+    @Autowired
+    private RestTemplate restTemplate;
+
+    @Autowired
+    private MicrosoftMTEngineConfiguration mtEngineConfiguration;
```
aurambaj, line 47: Auto-wiring the config works and I know the code use that pattern in some place but in that case it might be preferable to keep your service agnostic of spring. 

MicrosoftMTEngine (like S3BlobStorage) seems to be a good candidate to work without spring and be plain Java. 

You'd pass required config via the constructor instead (see S3BlobStorageConfigurationConfiguration). 

Also you end with a central place for setting up the MT engine like (see BlobStorageConfiguration). It is easier to see what the app is doing overall and you get a bit more flexibility (see DatabaseBlobStorageConfiguration it create the blobStorage but also setup job for clean up, etc).

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/MachineTranslationService.java
garionpin: Due to the caching the results of a non-public method. The error is emitted as that would not generally work - except in our case we use AspectJ - which does ensure that functionality is well covered. (Might make that method public at some point though as there might be some value in it being reused like that).

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/MachineTranslationService.java
garionpin: I'll add repositoryId as an additional parameter to be able to narrow the search - if not specified, we'll search in the entire DB. 

Anything that narrows the scope of the search is welcome as far as I'm concerned.

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/MachineTranslationService.java
```diff
@@ -0,0 +1,165 @@
...
+                .getTextUnitTranslations().get(0)
+                .getTranslations().get(0);
+    }
+
+    @SuppressWarnings("SpringCacheableComponentsInspection")
+    @Cacheable("machineTranslation")
+    @Timed("MachineTranslationService.getMachineTranslatedResponse")
+    TranslationsResponse getMachineTranslatedResponse(String sourceLocaleTag, List<String> targetLocaleTags, List<String> textSources, Boolean skipFunctionalProtection) {
+        if (textSources.isEmpty()) {
+            return new TranslationsResponse();
```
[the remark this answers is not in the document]
garionpin: It's to ensure we do not try to ask our specific MT engine implementations to translate empty/null strings - which in some cases results in errors - this ensured a well formatted response in those cases.

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/microsoft/MicrosoftMTEngine.java
garionpin: We can't use streams here (as with some of the other scenarios) as we need to iterate through two data sources at the same time.
The API response from the MT engine gives us back results in a specific order but with their own language tags - and we want to reply back to our users with the original language tags that they gave us for consistency.

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/microsoft/MicrosoftMTEngine.java
garionpin: 'zip(java.util.stream.Stream<A>, java.util.stream.Stream<B>, java.util.function.BiFunction<? super A,? super B,R>)' is marked unstable with @Beta 

- are we still Ok to use it given that warning?

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/NoOpEngine.java
```diff
@@ -0,0 +1,34 @@
...
+public class NoOpEngine implements MachineTranslationEngine{
+    static Logger logger = LoggerFactory.getLogger(NoOpEngine.class);
+
+    @Override
+    public TranslationSource getSource() {
+        return TranslationSource.NOOP;
+    }
+
+    @Override
+    public TranslationsResponse translate(List<String> textSources, String sourceLocaleTag, List<String> targetLocaleTags, String sourceMimeType, String customModel) {
```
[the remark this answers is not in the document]
garionpin: That's precisely the reason. 

To get the best performance - we should be able to provide a list of sources x a list of locales. If some engines don't support that, the specific engine implementations can then break it down in different types of batches (e.g.: if they support multiple sources for one locale or one source for multiple locales).

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/microsoft/MicrosoftMTEngine.java
aurambaj: I'd probably fine. looks like the beta was added later, like this old doc doesn't have it: https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Streams.html

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/microsoft/MicrosoftMTEngine.java
garionpin: Done

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/microsoft/MicrosoftMTEngineConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Configuration settings for the Microsoft Machine Translation client.
+ * Only used when "l10n.mt.impl" has the "MicrosoftMTEngine" value and the
+ * subscription key is provided in the "l10n.microsoft-mt.subscriptionKey" property.
+ *
+ * @author garion
+ */
+@Configuration
```
aurambaj, line 14: Not sure about that one, don't you want `@Component @ConfigurationProperties` and no `@ConditionalOnProperty`?

garionpin: Changed it to the advised setup now.
 
We do use @Configuration together with @ConfigurationProperties in other places: ActuatorHealthLegacyConfig, FlyWayConfig, QuartzPropertiesConfig, DropScheduleConfig, SlaCheckerEmailConfig, SmartlingClientConfiguration & a set of test classes.

The reason why I added @ConditionalOnProperty is to fail early if the engine isn't configured correctly, i.e.: the subscription key is missing - otherwise we'd only fail at runtime, when we actually do the http call.

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/MachineTranslationService.java
```diff
@@ -0,0 +1,252 @@
...
+        // TODO(garion): Implement functional protection / placeholder processing
+        if (skipFunctionalProtection != null && skipFunctionalProtection) {
+            logger.debug("function / placeholder protection");
+        }
+
+        return machineTranslationEngine.getTranslationsBySourceText(
+                textSources,
+                sourceBcp47Tag,
+                targetBcp47Tags,
+                "",
```
aurambaj, line 130: empty strings? should that be null, Optional or some enum?

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/MachineTranslationService.java
```diff
@@ -0,0 +1,252 @@
...
+                    return new AbstractMap.SimpleEntry<>(sourceText, translations);
+                })
+                .filter(entry -> !entry.getValue().isEmpty())
+                .collect(ImmutableMap.toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
+    }
+
+    private ImmutableMap<String, TextUnitDTO> getMostRecentTranslationsByLanguage(ArrayList<Long> repositoryIds, ArrayList<String> repositoryNames, String sourceText) {
+        List<TextUnitDTO> textUnitDTOsForLeveraging = getLeveragedTextUnitDTOS(sourceText, repositoryIds, repositoryNames);
+
+        Comparator<TextUnitDTO> tmTextUnitCreatedDateComparator = Comparator.comparing(TextUnitDTO::getTmTextUnitCreatedDate);
```
aurambaj, line 183: nit: fine inlining?

webapp/src/main/java/com/box/l10n/mojito/service/leveraging/LeveragerByContent.java
```diff
@@ -40,6 +42,23 @@ public List<TextUnitDTO> getLeveragingMatches(TMTextUnit tmTextUnit, Long source
         return textUnitSearcher.search(textUnitSearcherParameters);
     }
 
+    public List<TextUnitDTO> getLeveragingMatchesForRepository(TMTextUnit tmTextUnit, Long sourceTmId, ArrayList<Long> repositoryIds, ArrayList<String> repositoryNames) {
```
aurambaj, line 45: nit: this should be in another class since it is not used by this leverager. probably fine in the caller class?

aurambaj (commented):
looks great! a few question and mostly cosmetic comments

webapp/src/main/java/com/box/l10n/mojito/rest/machinetranslation/BatchTranslationRequestDTO.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * @author garion
+ */
+public class BatchTranslationRequestDTO {
+    List<String> textSources;
+    String sourceBcp47Tag;
+    ArrayList<String> targetBcp47Tags;
+    boolean skipFunctionalProtection;
+    boolean skipLeveraging;
+    ArrayList<Long> repositoryIds;
```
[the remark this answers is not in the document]
garionpin: Left over from when I had those as params (modeled after TextUnitWS).
Fixed now.

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/MachineTranslationService.java
```diff
@@ -0,0 +1,252 @@
...
+        ImmutableMap<String, ImmutableList<TranslationDTO>> machineTranslationsBySourceText = getMachineTranslationBySourceText(
+                getSourceBcp47TagOrDefault(sourceBcp47Tag),
+                targetBcp47Tags,
+                untranslatedStrings, skipFunctionalProtection);
+
+        translationsBySourceText = translationMerger.getMergedTranslationsBySourceText(translationsBySourceText, machineTranslationsBySourceText);
+        return getSortedTranslationsResponse(textSources, targetBcp47Tags, translationsBySourceText);
+    }
+
+    public TranslationDTO getSingleTranslation(
```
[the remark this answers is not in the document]
garionpin: That is correct - TranslationDTO is meant for the simpler getSingleTranslation() API while TranslationResponseDTO is the fully featured batch response - that also includes the source strings and all of the data in a properly sorted order.

I'd expect most applications, at least initially, to use the simpler variant.

webapp/src/main/java/com/box/l10n/mojito/rest/machinetranslation/TranslationRequestDTO.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.ArrayList;
+
+/**
+ * DTO to translate one source string for one target language.
+ *
+ * @author garion
+ */
+public class TranslationRequestDTO {
+    String textSource;
```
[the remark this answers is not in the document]
garionpin: There are a couple of differences in the parameters expected as you mentioned, but also, the return type for the single translation API method (which we expect initially to account for the bulk of the integrations) is significantly easier to consume than the batch one.

I don't see much benefit in having both a sync and an async variant for exactly the same method and input object and response.

Given the choice - I'd expect the consumers to always choose the simpler, sync, variant to the detriment of the async one. The main reason to have async for the batch is better management of resources on the server side - which I wouldn't expect the clients to care greatly about.

A natural way of achieving good performance and scaling in this case is to keep the small/fast API calls synchronous, while forcing the expensive computations/operations to an async model. 
Doing large batches synchronously isn't great for the server, while doing async (+ the PollableTask overhead) on a call that returns very fast isn't justified either. 
Not exposing those suboptimal choices ensures that the client can't make the wrong choice.

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/MachineTranslationService.java
garionpin: That should be null for now. Once a source placeholder escaping mechanism is introduced, that will be replaced with a relevant enum.

aurambaj (approved):

PR #789, opened 2022-04-12
Add a database-backed cache

garionpin:
Add a JCache compatible cache that stores the entries in the Mojito
database.
Entries can have a TTL associated with them.
Multiple cache types can be stored in the database, in the same data
structure.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/CachingConfig.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/CachingConfig.java
@@ -1,36 +1,91 @@
 package com.box.l10n.mojito;
 
+import com.box.l10n.mojito.service.cache.DatabaseCache;
+import com.box.l10n.mojito.service.cache.DatabaseCacheConfiguration;
+import com.box.l10n.mojito.service.cache.TieredCache;
+import com.box.l10n.mojito.service.machinetranslation.MTServiceCacheConfiguration;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import org.checkerframework.checker.nullness.qual.NonNull;
 import org.springframework.cache.Cache;
 import org.springframework.cache.CacheManager;
 import org.springframework.cache.annotation.CachingConfigurerSupport;
 import org.springframework.cache.annotation.EnableCaching;
+import org.springframework.cache.caffeine.CaffeineCache;
 import org.springframework.cache.concurrent.ConcurrentMapCache;
 import org.springframework.cache.interceptor.KeyGenerator;
 import org.springframework.cache.support.SimpleCacheManager;
 import org.springframework.context.annotation.AdviceMode;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+
 import java.util.Arrays;
+import java.util.concurrent.TimeUnit;
 
 /**
- *
  * @author jaurambault
  */
 @Configuration
 @EnableCaching(mode = AdviceMode.ASPECTJ)
 public class CachingConfig extends CachingConfigurerSupport {
 
+    final MTServiceCacheConfiguration mtServiceCacheConfiguration;
+
+    public CachingConfig(MTServiceCacheConfiguration mtServiceCacheConfiguration) {
+        this.mtServiceCacheConfiguration = mtServiceCacheConfiguration;
+    }
+
     @Bean
     @Override
     public CacheManager cacheManager() {
-        Cache defaultCache = new ConcurrentMapCache("default");
-        Cache localesCache = new ConcurrentMapCache("locales");
-        Cache pluralForm = new ConcurrentMapCache("pluralForms");
+        Cache defaultCache = new ConcurrentMapCache(CacheType.Names.DEFAULT);
+        Cache localesCache = new ConcurrentMapCache(CacheType.Names.LOCALES);
+        Cache pluralForm = new ConcurrentMapCache(CacheType.Names.PLURAL_FORMS);
+
+        TieredCache machineTranslationTieredCache = getMachineTranslationCache();
 
         SimpleCacheManager manager = new SimpleCacheManager();
-        manager.setCaches(Arrays.asList(defaultCache, localesCache, pluralForm));
+        manager.setCaches(Arrays.asList(
+                defaultCache,
+                localesCache,
+                pluralForm,
+                machineTranslationTieredCache));
 
         return manager;
+    }
+
+    private TieredCache getMachineTranslationCache() {
+        com.github.benmanes.caffeine.cache.@NonNull Cache<Object, Object> machineTranslationCaffeineCache =
+                Caffeine.newBuilder()
+                        .expireAfterWrite(
+                                mtServiceCacheConfiguration.getInMemory().getTtl().toMillis(),
+                                TimeUnit.MILLISECONDS)
+                        .maximumSize(mtServiceCacheConfiguration.getInMemory().getMaximumSize())
+                        .build();
+
+        CaffeineCache machineTranslationMemoryCache = new CaffeineCache(
+                "machineTranslationInMemory",
+                machineTranslationCaffeineCache
+        );
+
+        DatabaseCache machineTranslationDatabaseCache = null;
+
+        if (mtServiceCacheConfiguration.getDatabase().isEnabled()) {
+            DatabaseCacheConfiguration mtDbCacheConfiguration = new DatabaseCacheConfiguration();
+            mtDbCacheConfiguration.setTtl(mtServiceCacheConfiguration.getDatabase().getTtl());
+            mtDbCacheConfiguration.setEvictEntryOnDeserializationFailure(
+                    mtServiceCacheConfiguration.getDatabase().isEvictEntryOnDeserializationFailure());
+
+            machineTranslationDatabaseCache = new DatabaseCache(
+                    "machineTranslationInDb",
+                    mtDbCacheConfiguration
+            );
+        }
+
+        return new TieredCache(
+                CacheType.Names.MACHINE_TRANSLATION,
+                machineTranslationMemoryCache,
+                machineTranslationDatabaseCache
+        );
     }
 
     @Bean
--- a/webapp/src/main/java/com/box/l10n/mojito/entity/ApplicationCache.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/entity/ApplicationCache.java
@@ -1,0 +1,84 @@
+package com.box.l10n.mojito.entity;
+
+import org.hibernate.annotations.BatchSize;
+import org.hibernate.annotations.Type;
+import org.joda.time.DateTime;
+
+import javax.persistence.Column;
+import javax.persistence.EmbeddedId;
+import javax.persistence.Entity;
+import javax.persistence.ForeignKey;
+import javax.persistence.Index;
+import javax.persistence.JoinColumn;
+import javax.persistence.Lob;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+
+/**
+ * Entity that contains the cache entry details for a database-backed application cache.
+ * Each entity is identified by a specific {@Link ApplicationCacheType} and MD5 hash of the key.
+ * A TTL / expiry date can also be configured on the entries, the application being responsible for the entry eviction.
+ *
+ * @author garion
+ */
+@Entity
+@Table(name = "application_cache",
+        indexes = {
+                @Index(name = "UK__APPLICATION_CACHE__CACHE_TYPE_ID__KEY_MD5", columnList = "cache_type_id, key_md5", unique = true),
+                @Index(name = "I__APPLICATION_CACHE__EXPIRY_DATE", columnList = "expiry_date")
+        })
+@BatchSize(size = 1000)
+public class ApplicationCache extends BaseEntity {
+    @ManyToOne(optional = false)
+    @JoinColumn(name = "cache_type_id", foreignKey = @ForeignKey(name = "FK__APPLICATION_CACHE__CACHE_TYPE_ID"))
+    private ApplicationCacheType applicationCacheType;
+
+    @Column(name = "key_md5", length = 32)
+    private String keyMD5;
+
+    @Column(name = "value", length = Integer.MAX_VALUE)
+    @Lob
+    private byte[] value;
+
+    @Column(name = "created_date")
+    @Type(type = "org.jadira.usertype.dateandtime.joda.PersistentDateTime")
+    private DateTime createdDate;
+
+    @Column(name = "expiry_date")
+    @Type(type = "org.jadira.usertype.dateandtime.joda.PersistentDateTime")
+    private DateTime expiryDate;
+
+    public ApplicationCache() {
+    }
+
+    public ApplicationCache(ApplicationCacheType applicationCacheType, String keyMD5, byte[] value, DateTime expiryDate) {
+        this.applicationCacheType = applicationCacheType;
+        this.keyMD5 = keyMD5;
+        this.value = value;
+        this.expiryDate = expiryDate;
+    }
+
+    public byte[] getValue() {
+        return value;
+    }
+
+    public void setValue(byte[] value) {
+        this.value = value;
+    }
+
+    public DateTime getCreatedDate() {
+        return createdDate;
+    }
+
+    public void setCreatedDate(DateTime createdDate) {
+        this.createdDate = createdDate;
+    }
+
+    public DateTime getExpiryDate() {
+        return expiryDate;
+    }
+
+    public void setExpiryDate(DateTime expireDate) {
+        this.expiryDate = expireDate;
+    }
+}
--- a/webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCache.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCache.java
@@ -1,0 +1,206 @@
+package com.box.l10n.mojito.service.cache;
+
+import com.box.l10n.mojito.entity.ApplicationCache;
+import com.box.l10n.mojito.entity.ApplicationCacheType;
+import com.google.common.base.Preconditions;
+import org.apache.commons.codec.digest.DigestUtils;
+import org.slf4j.Logger;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Configurable;
+import org.springframework.cache.support.AbstractValueAdaptingCache;
+import org.springframework.cache.support.NullValue;
+import org.springframework.core.serializer.support.SerializingConverter;
+import org.springframework.transaction.annotation.Transactional;
+import org.springframework.util.ObjectUtils;
+
+import javax.annotation.Nullable;
+import java.time.temporal.ChronoUnit;
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.concurrent.Callable;
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+/**
+ * Implements a cache that is stored in the database.
+ * <p>
+ * The key objects are serialized and an MD5 of the serialized result is used, together with the cache type, as the
+ * identifying key in the database.
+ * The value objects are serialized/deserialized to byte arrays and back.
+ *
+ * @author garion
+ */
+@Configurable
+public class DatabaseCache extends AbstractValueAdaptingCache {
+    /**
+     * logger
+     */
+    static Logger logger = getLogger(DatabaseCache.class);
+
+    private static final byte[] BINARY_NULL_VALUE = (new SerializingConverter()).convert(NullValue.INSTANCE);
+
+    @Autowired
+    ApplicationCacheUpdaterService applicationCacheUpdaterService;
+
+    @Autowired
+    ApplicationCacheTypeRepository applicationCacheTypeRepository;
+
+    @Autowired
+    ApplicationCacheRepository applicationCacheRepository;
+
+    private final String cacheName;
+    private final DatabaseCacheConfiguration cacheConfig;
+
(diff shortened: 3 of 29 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCache.java
```diff
@@ -0,0 +1,226 @@
...
+    @Transactional
+    @Override
+    public void clear() {
+        checkCacheTypeConfigured();
+        applicationCacheRepository.clearCache(cacheType.getId());
+    }
+
+    @Pollable(async = true, message = "Cleaning up expired cache entries.")
+    private void scheduleProbabilisticExpiredEvictionAsync() {
+        if (rand.nextDouble() < cacheConfig.getWritesToExpiredEvictionRatio()) {
```
aurambaj, line 170: the "WritesToExpiredEvictionRatio" concept is not obvious to me at first glance.

webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCacheConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Scope;
+
+import java.time.Duration;
+
+/**
+ * This class enables custom configurations for a database-backed cache.
+ * By default, no TTL is configured (an expiry date isn't writen for entries)
+ * and cache evictions for expired entries are attempted once every 10,000 writes.
```
aurambaj, line 14: not sure to understand the eviction based on write, specifically for current use case, we'll never reach this number of writes?

webapp/src/main/java/com/box/l10n/mojito/CachingConfig.java
```diff
@@ -1,34 +1,62 @@
...
 /**
- *
  * @author jaurambault
  */
 @Configuration
 @EnableCaching(mode = AdviceMode.ASPECTJ)
 public class CachingConfig extends CachingConfigurerSupport {
 
+    @Autowired
+    DatabaseCacheConfiguration mtDbCacheConfiguration;
```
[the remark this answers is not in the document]
garionpin: The DatabaseCache/CacheConfiguration classes are supposed to be agnostic of what purpose they are used for (as any cache would) - and the specific cache settings we define here are purely for the MT calls. 

We can have additional caches defined for other APIs - similar to how we have different caches for locales, pluralForms etc. - and they would have require their own configuration/TTL etc. This follows the same model as for CaffeineCache/RedisCache etc.

webapp/src/main/java/com/box/l10n/mojito/entity/ApplicationCache.java
```diff
@@ -0,0 +1,83 @@
...
+ * @author garion
+ */
+@Entity
+@Table(name = "application_cache",
+        indexes = {
+                @Index(name = "I__APPLICATION_CACHE__CACHE_TYPE_ID_EXPIRY_DATE", columnList = "cache_type_id, expiry_date"),
+        })
+@BatchSize(size = 1000)
+public class ApplicationCache {
+    @EmbeddedId
```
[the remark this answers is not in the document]
garionpin: I followed the best practices defined here for the composite key setup: https://vladmihalcea.com/the-best-way-to-map-a-composite-primary-key-with-jpa-and-hibernate/ 

I generally use a surrogate PK key for most entities (to help with joins performance etc.) - but in this case a natural key seems a better fit - as this is a completely self contained table - and an auto-incremented ID PK not only doesn't give us anything extra - but it makes the lookups slightly slower.

As we primarily lookup by the cache id + hash of the key (and we would never lookup by an auto-incremented ID for a cache) - having these as a natural key used for clustering data is likely close to the ideal setup.

That said - I am interested in learning more about the best practices you mentioned - can you expand on what should be the ideal for this specific scenario?

webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCache.java
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    @Transactional
+    @Override
+    public void clear() {
+        checkCacheTypeConfigured();
+        applicationCacheRepository.clearCache(cacheType.getId());
+    }
+
+    @Pollable(async = true, message = "Cleaning up expired cache entries.")
```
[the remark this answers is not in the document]
garionpin: Only one execution is needed (not across the cluster at once) - as the data to be cleared is stored centrally in the DB.
The fire and forget nature is to ensure it doesn't impact the performance for the regular write operations.
I feel there might/should be a better way to do this though..

webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCacheConfiguration.java
garionpin: We can certainly configure that ratio / number to something we think is more appropriate to our use case.

The goal is to have a cache that takes care of its own clean-up/eviction - and doesn't require additional setup (e.g. external scheduled tasks) for clearing up expired entries - as that part will likely be missed and will result in stale data never getting cleaned.

There might a better mechanic out there to achieve this.
The way that goal is achieved in the current implementation is by scheduling an async cleanup of expired entries probabilistically every x writes (as the number of entries to be cleaned is generally proportional to the number of total entries - which is in turn proportional to the number of writes we do in the system).

Can you let me know if you can think of some better ways to achieve the same goal, while also keeping the cache self-contained?

webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCacheConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public DatabaseCache getConfiguredDatabaseCache(String cacheName, DatabaseCacheConfiguration databaseCacheConfiguration) {
+        return this.getDatabaseCache(cacheName, databaseCacheConfiguration);
+    }
+
+    /**
+     * Method-based dependency injection is required here to be able to create multiple {@Link DatabaseCache} instances
+     * at runtime with different configurations that get their own dependencies auto-wired.
+     */
+    @Lookup
```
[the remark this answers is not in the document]
garionpin: Yes, that's the recommended approach by Spring for this scenario: https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-factory-method-injection


As we can have different caches instantiated for different scenarios (MT, locales, pluralForms etc.) - we need different instances for the Caches. 
The class itself has autowired dependencies - so we can not new() it up manually - those objects need to be created by Spring so that all of the dependencies are resolved correctly. The key thing here is that we do not have only one Cache instantiated - we can have multiple ones configured (as with any cache). I initially attempted to create Bean definitions for each with @Qualifier annotations - however that doesn't help in this scenario.

The cache instances also have different configurations associated with them (name, settings - similar to RedisCache and others) - information which is only available at runtime. The cacheConfig objects similarly need to be separate objects - as different caches will likely use different setting configurations.


Now, the only way I could get all of the above to work is by using the @Lookup annotation provided by Spring to create prototype-scoped instances of our component. (The other option I found was to grab a reference to a BeanFactory to essentially create the components through the Spring factory - however, that leaks Spring-specific code in the classes - which I thought was best avoided).


I couldn't figure out a way to do that using standard constructors - how would it work given the above constraints - while maintaining the ability to have different caches in place that get configured differently at runtime (different names, ttl configs etc.)?

webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCache.java
```diff
@@ -0,0 +1,226 @@
...
+        if (maybeCacheResult.isPresent()) {
+            byte[] bytes = maybeCacheResult.get().getValue();
+
+            // Permanently evict cache entries that cause deserialization failures (e.g.: due to breaking changes in
+            // target class structure).
+            try {
+                Object deserializedValue = deserializeCacheValue(bytes);
+                return this.fromStoreValue(deserializedValue);
+            } catch (Exception e) {
+                logger.error(String.format("Could not deserialize cache entry. Problematic serialized value: %s . Exception: ", Arrays.toString(bytes)), e);
```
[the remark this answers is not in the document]
garionpin: Yes! That was my initial intent as well - forgot to change it.

webapp/src/main/java/com/box/l10n/mojito/entity/ApplicationCache.java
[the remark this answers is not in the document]
aurambaj: I don't remember exactly at this point... more of a vague memory. Not sure if it was an hibernate thing (the embedded id maybe have some complication with implementing serializable, or maybe it can lead to doing a lot of overcomplicated things) or maybe more of general DB design recommandation (as you said having surrogate PK just make it simpler for join and understanding things quickly). Whatever I remember might be outdated too.

webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCache.java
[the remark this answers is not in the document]
aurambaj: after our discussion, the cleanup job shoud probably similar to what we have for other system, see job annotated with `@DisallowConcurrentExecution`

webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCache.java
aurambaj: i don't have strong feeling on this one but I think we'd fine with a cronjob for now. We do have similar job already to perform cleanups

webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCacheConfiguration.java
aurambaj: i can see how the self contained clean is better for in-memory cache but since this is a distributed cache assuming many clients, they'll try to do perform clean concurrently which does not seem great. So we'd need some kind of external scheduling system that prevent concurrent cleanups. We can do that with Quartz (`@DisallowConcurrentExecution` jobs): we have the capability to run 1 single task in the whole cluster. Still this can be trigger within the write based on the ratio or with a cron job.

webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCacheConfiguration.java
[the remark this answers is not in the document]
aurambaj: just for the record as discussed, @Configurable can be similar to inject field using `new`.

Here are some suggestions. 

1) just be very specific about create a cache for the MTService

```properties
l10n.mtservice.cache.inmemory.ttl=1000
l10n.mtservice.cache.inmemory.max-size=5000
l10n.mtservice.cache.database.enabled=true
l10n.mtservice.cache.database.ttl=5000
l10n.mtservice.cache.database.eviction-cron=* * * * 5
```
```java

@Configuration
@ConfigurationProperties(prefix = "l10n.mtservice.cache")
public class MTServiceCacheConfig {

    Database database = new Database();

    InMemory inMemory = new InMemory();

    static class Database {
        boolean enabled = false;
        int ttl = -1;
        String evictionCron = null;

        public boolean isEnabled() {
            return enabled;
        }

        public void setEnabled(boolean enabled) {
            this.enabled = enabled;
        }

        public int getTtl() {
            return ttl;
        }

        public void setTtl(int ttl) {
            this.ttl = ttl;
        }

        public String getEvictionCron() {
            return evictionCron;
        }

        public void setEvictionCron(String evictionCron) {
            this.evictionCron = evictionCron;
        }
    }

    static class InMemory {
        int ttl = -1;

        public int getTtl() {
            return ttl;
        }

        public void setTtl(int ttl) {
            this.ttl = ttl;
        }
    }

    public Database getDatabase() {
        return database;
    }

    public void setDatabase(Database database) {
        this.database = database;
    }

    public InMemory getInMemory() {
        return inMemory;
    }

    public void setInMemory(InMemory inMemory) {
        this.inMemory = inMemory;
    }
}
```
Then you autowrite that config in CacheConfig and create the cache instance just for the mt service

2) enable configuration of many generic database cache that can be used kind of what spring provides.

```properties
l10n.caches.database.guinness.ttl=100
l10n.caches.database.guinness.eviction-cron=* * * * 5
l10n.caches.database.kilkenny.ttl=200
l10n.caches.database.kilkenny.eviction-cron=* * * * 5
```

```java
package com.box.l10n.mojito.service.mtcache;

import org.springframework.boot.context.properties.ConfigurationProperties;
import org.springframework.context.annotation.Configuration;

import java.util.LinkedHashMap;
import java.util.Map;

@Configuration
@ConfigurationProperties(prefix = "l10n.caches")
public class DatabaseCachesConfig {

    Map<String, DatabaseCache> database = new LinkedHashMap<>();

    static class DatabaseCache {
        int ttl = -1;
        String evictionCron;

        public int getTtl() {
            return ttl;
        }

        public void setTtl(int ttl) {
            this.ttl = ttl;
        }

        public String getEvictionCron() {
            return evictionCron;
        }

        public void setEvictionCron(String evictionCron) {
            this.evictionCron = evictionCron;
        }
    }

    public Map<String, DatabaseCache> getDatabase() {
        return database;
    }

    public void setDatabase(Map<String, DatabaseCache> database) {
        this.database = database;
    }
}

```

You can iterate on the list like that, so basically in the CacheConfig class you can just instanciate/register has many as you want
```
databaseCachesConfig.getDatabase().entrySet().stream().forEach(e ->
                System.out.println(e.getKey() + " : " + e.getValue().getTtl() + "," + e.getValue().getEvictionCron()));
```

webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCache.java
```diff
@@ -0,0 +1,226 @@
...
+ *
+ * The key objects are serialized and an MD5 of the serialized result is used, together with the cache type, as the
+ * identifying key in the database.
+ * The value objects are serialized/deserialized to byte arrays and back.
+ *
+ *
+ * @author garion
+ */
+@Component
+@Scope(BeanDefinition.SCOPE_PROTOTYPE)
```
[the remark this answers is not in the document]
garionpin: Using @Configurable instead now.

webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCacheConfiguration.java
garionpin: Using a Quartz job for the cleanup now.

webapp/src/main/java/com/box/l10n/mojito/service/cache/DatabaseCacheConfiguration.java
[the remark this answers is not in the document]
garionpin: Thank you for the guidance there - that was very useful.
I've implemented the suggested pattern now.

webapp/src/main/java/com/box/l10n/mojito/service/cache/TieredCache.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Enables a tiered caching setup where items are stored in both caches and retrieval
+ * is done preferentially from the tier 1 cache first and then on cache miss from tier 2.
+ *
+ * Usually expected to be used with an ephemeral in-process memory cache as tier 1 and a
+ * durable/persistent tier 2 cache.
+ *
+ * @author garion
+ */
+public class TieredCache extends AbstractValueAdaptingCache {
```
aurambaj, line 19: Spring doesn't expose something similar? Is it only in the annotation that you can chain caches to use?

webapp/src/main/java/com/box/l10n/mojito/service/leveraging/LeveragerByContentAndRepository.java
```diff
@@ -15,7 +16,7 @@
  *
  * @author garion
  */
-@Component
+@Configurable
```
aurambaj, line 19: revert?

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/TextUnitTranslationGroupDTO.java
```diff
@@ -8,7 +9,7 @@
  *
  * @author garion
  */
-public class TextUnitTranslationGroupDTO {
+public class TextUnitTranslationGroupDTO implements Serializable {
```
aurambaj, line 12: is that for the cache?

webapp/src/main/java/com/box/l10n/mojito/CachingConfig.java
```diff
@@ -1,38 +1,119 @@
...
-        Cache defaultCache = new ConcurrentMapCache("default");
-        Cache localesCache = new ConcurrentMapCache("locales");
-        Cache pluralForm = new ConcurrentMapCache("pluralForms");
+        Cache defaultCache = new ConcurrentMapCache(CacheType.Names.DEFAULT);
+        Cache localesCache = new ConcurrentMapCache(CacheType.Names.LOCALES);
+        Cache pluralForm = new ConcurrentMapCache(CacheType.Names.PLURAL_FORMS);
+
+        TieredCache machineTranslationTieredCache = getMachineTranslationCache();
+
+        List<DatabaseCache> databaseCaches = getRuntimeDatabaseCaches();
```
aurambaj, line 56: oh, you implemented both solutions (specific cache init for MT and generic case init)? I was thinking of an "or" in my previous comment, but I guess that works! The generic configuration won't be use right now, right?

webapp/src/main/java/com/box/l10n/mojito/entity/ApplicationCache.java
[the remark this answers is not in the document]
aurambaj: ooopss I meant sticking with simple id. A bit late now, but DB schema is hard to change later. Are you sure about the benefis for the composed key?

webapp/src/test/java/com/box/l10n/mojito/service/cache/DatabaseCacheTest.java
```diff
@@ -0,0 +1,228 @@
...
+        Assertions.assertNull(databaseCacheWith1SecondTTL.lookup(KEY));
+    }
+
+    @Test
+    public void testExpiryBasedEviction() throws InterruptedException {
+        Duration ttl = Duration.ofSeconds(1);
+        databaseCacheConfiguration.setTtl(ttl);
+        DatabaseCache databaseCacheWithAggressiveEviction = new DatabaseCache("databaseCacheWithAggressiveEviction", databaseCacheConfiguration);
+
+        // This test is not supported on H2
```
aurambaj, line 163: - `H2` --> `Hsql`
- Use assume?
- why not using the mysql check since we haven't tested that code on other dbs?

webapp/src/test/java/com/box/l10n/mojito/service/cache/DatabaseCacheTest.java
```diff
@@ -0,0 +1,228 @@
...
+        if (dbUtils.isHsql()) {
+            return;
+        }
+
+        putKeyValue(databaseCacheWithAggressiveEviction, KEY, VALUE);
+        Thread.sleep(ttl.toMillis());
```
aurambaj, line 169: Sleeping in test eventually makes the test suite slow. Can we avoid?

webapp/src/main/java/com/box/l10n/mojito/service/cache/TieredCache.java
garionpin: Using a @Cacheable annotation where you can chain caches seems to be the only way Spring supports this as far as I can tell. Though it's not very transparent in that case as to how the system deals with cache misses and re-population in a chained scenario.

webapp/src/main/java/com/box/l10n/mojito/service/leveraging/LeveragerByContentAndRepository.java
garionpin: Needed - as we new() up this class in order to be able to inject the repository search params in the contructor - and the main matching method relies on this class from AbstractLeverager being initialized correctly (non-null).
```
@Autowired
    protected TextUnitSearcher textUnitSearcher;
```

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/TextUnitTranslationGroupDTO.java
garionpin: Yes, to serialize and store the response as a binary blob.

webapp/src/main/java/com/box/l10n/mojito/CachingConfig.java
garionpin: Removed generic case - not needed currently.

webapp/src/test/java/com/box/l10n/mojito/service/cache/DatabaseCacheTest.java
garionpin: Done

webapp/src/test/java/com/box/l10n/mojito/service/cache/DatabaseCacheTest.java
garionpin: Avoided now

webapp/src/main/java/com/box/l10n/mojito/service/leveraging/LeveragerByContentAndRepository.java
aurambaj: oh that's from another PR?

aurambaj (approved):

PR #800, opened 2022-04-27
JDK8 XML backwards compatibility fix in WSTestBase

garionpin:


aurambaj (commented):
that'd be good enough then? looks like a good option if sharing the code in function, no?

PR #803, opened 2022-05-05
Fixes for compiler type & assignment warnings

garionpin:
Fixed multiple instances of:
- Raw use of parameterized class
- Unchecked assignment


aurambaj (approved):
thanks @geoarion !!

PR #813, opened 2022-05-23
Run all integration tests using a MySQL test container instead of HSQL

garionpin:
Replace HSQL with MySQL for higher fidelity integration tests and to
be able to also test native queries and platform-specific functionality.

### DIFF
--- a/cli/src/test/resources/application.properties
+++ b/cli/src/test/resources/application.properties
@@ -1,4 +1,7 @@
 info.build.version=@project.version@
 
+spring.flyway.enabled=true
+spring.datasource.url=jdbc:tc:mysql:5.7://localhost:3306/mojito?characterEncoding=UTF-8&useUnicode=true&TC_INITSCRIPT=db/mysql/test_init.sql
+
 l10n.console-writer.ansi-code-enabled=false
 l10n.console-writer.output-type=ANSI_LOGGER
--- a/docs/_docs/guides/open-source-contributors.md
+++ b/docs/_docs/guides/open-source-contributors.md
@@ -242,7 +242,20 @@ For example to create Demo data, you can now run: `mojito demo-create -n DemoCLI
 Alternatively, install the CLI using the [install scripts]({{ site.url }}/docs/guides/install/#cli-install-script).
 
 ## Run Unit Tests
-   
+
+
+### Testcontainers
+By default, if no datasource is configured in the application.properties file, integration tests rely on 
+[Testcontainers](https://www.testcontainers.org/)
+to 
+provide a configured
+MySql instance in a test-scoped container.
+
+For that to work,  you will need Docker installed and running on the local machine. 
+
+See [general Docker requirements](https://www.testcontainers.org/supported_docker_environment/) for Testcontainers.
+
+### Test command
 ```sh
 cd ${PROJECT_DIR}
 mvn test
--- a/pom.xml
+++ b/pom.xml
@@ -76,6 +76,24 @@
             </exclusions>
         </dependency>
 
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <version>1.17.1</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>junit</groupId>
+                    <artifactId>junit</artifactId>
+                </exclusion>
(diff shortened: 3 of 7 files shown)


docs/_docs/guides/open-source-contributors.md
```diff
@@ -242,7 +242,20 @@ For example to create Demo data, you can now run: `mojito demo-create -n DemoCLI
 Alternatively, install the CLI using the [install scripts]({{ site.url }}/docs/guides/install/#cli-install-script).
 
 ## Run Unit Tests
-   
+
+
+### Testcontainers
+By default, if no datasource is configured in the application.properties file, integration tests rely on
```
aurambaj, line 248: One of the advantage of the current setup is that everything can be running out of the box for simple development. Adding docker as requirement adds complexity and might flakiness/impact perf. I'd keep the default settings to run in memory and just activate container testing with a spring profile that we can use or not locally and in dev. On current CI we could run both tests in parallel at the beginning?

pom.xml
```diff
@@ -76,6 +76,24 @@
...
+                    <groupId>junit</groupId>
+                    <artifactId>junit</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>mysql</artifactId>
+            <version>1.17.1</version>
```
aurambaj, line 94: should that be a "test" dependency?

webapp/src/test/java/com/box/l10n/mojito/db/MySQLIntegrationTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringBootTest(classes = {DBUtils.class})
+public class MySQLIntegrationTest {
+    @Autowired
+    DBUtils dbUtils;
+
+    @Test
+    public void testIntegrationTestsRunOnMysql() {
+        Assert.assertTrue(dbUtils.isMysql());
```
aurambaj, line 23: that'd fail if we run the test against anything but mysql?

cli/src/test/resources/application.properties
```diff
@@ -1,4 +1,7 @@
 info.build.version=@project.version@
 
+spring.flyway.enabled=true
+spring.datasource.url=jdbc:tc:mysql:5.7://localhost:3306/mojito?characterEncoding=UTF-8&useUnicode=true&TC_INITSCRIPT=db/mysql/test_init.sql
```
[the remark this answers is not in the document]
garionpin: It won't cause any issues, it's just a placeholder value when Testcontainers is used, see https://www.testcontainers.org/modules/databases/jdbc/ for details.

pom.xml
garionpin: This was not made a test dependency explicitly so that someone could decide to configure their local instance to run with TC as well out of the box with a configuration change, without needing to separately install MySQL.

webapp/src/test/java/com/box/l10n/mojito/db/MySQLIntegrationTest.java
garionpin: That is the intent - to make sure that integration tests run in an environment that is similar to production, otherwise the tests risk not being as valuable and missing real world issues (as we've seen HSQL has missed several times before).

docs/_docs/guides/open-source-contributors.md
garionpin: Given that other popular Java frameworks now rely on similar dependencies for testing (e.g.: Quarkus) - I don't have any specific reason to believe that flakiness/perf is going to be significantly impacted by this change.

The biggest value for integration tests is that they find issues locally, before deploying the application live. I believe ensuring the default behaviour is as close to production and finds as many bugs should be "base" state then. (opt-out vs opt-in).

If the target is simplifying setup - we could make the existing setup steps significantly simpler by removing the MySQL install, configuration & application configuration setup steps and replacing that with the requirement to just have Docker instead (+ also changing the default for the local application to use TC). 
I would expect it's more likely today to have Docker already installed on one's machine than to find the right version of MySQL that this application depends on - but I wouldn't have any data backing that.

PR #814, opened 2022-05-30
Alter asset_text_unit_usages.usages size migration

montedonio-pinterest:


webapp/src/main/resources/db/migration/V59__Alter_text_unit_usages.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE mojitodb.asset_text_unit_usages MODIFY COLUMN usages varchar(512) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NULL;
```
aurambaj, line 1: For the script to work on any database, remove the db name `mojitodb`. 

Do we really need this part `CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NULL;`?

aurambaj (approved):

PR #815, opened 2022-06-01
Created before/after filters branches page

montedonio-pinterest:


webapp/package-lock.json
```diff
@@ -341,11 +341,6 @@
                   "dev": true
```
maallen, line 341: Are the updates to this file required? Usually I would exclude local changes to this file from my commit

webapp/src/main/resources/public/js/components/branches/BranchesStatusDropdown.js
```diff
@@ -17,10 +20,10 @@ class BranchesStatusDropdown extends React.Component {
         "onEmptyChanged": PropTypes.func.isRequired,
         "onNotEmptyChanged": PropTypes.func.isRequired,
         "onOnlyMyBranchesChanged": PropTypes.func.isRequired,
-    }
-
-    onFilterSelected(filter) {
```
aurambaj, line 22: why is this removed, can you explain in the commit description?

webapp/src/main/resources/public/js/stores/PaginatorStore.js
```diff
@@ -34,7 +34,9 @@ class PaginatorStore {
     }
 
     changeCurrentPageNumber(currentPageNumber) {
-        this.currentPageNumber = currentPageNumber;
```
aurambaj, line 37: why is this added?

webapp/src/main/resources/public/js/stores/branches/BranchesHistoryStore.js
```diff
@@ -45,6 +46,20 @@ class BranchesHistoryStore {
         this.openBranchStatistic = openBranchStatistic;
     }
 
+    changeCreatedBefore(createdBefore) {
+        this.createdBefore = createdBefore
+            ? moment(createdBefore)
+                .format("YYYY-MM-DDTHH:mm")
```
aurambaj, line 52: Do we use this construct in the other filter?

aurambaj (commented):
Can you squash the changes into a single commit, and add explanation on what it does exactly.

webapp/src/main/resources/public/js/actions/branches/BranchesDataSource.js
```diff
@@ -38,6 +39,20 @@ const BranchesDataSource = {
                 branchStatisticSearcherParameters.empty(false);
             }
 
+            if (branchesSearchParamState.createdBefore) {
+                branchStatisticSearcherParameters.createdBefore(
+                    moment(branchesSearchParamState.createdBefore)
+                        .format('YYYY-MM-DDTHH:mm')
```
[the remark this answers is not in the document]
montedonio-pinterest: I preferred to handle using moment because this way we can ensure the format that will be correct both from the input change and the queryString retrieval. This format makes the url cleaner as well; easier to change the link by hand because it won't have all of the escape characters.

webapp/src/main/resources/public/js/components/branches/BranchesSearchResults.js
```diff
@@ -202,9 +202,9 @@ class BranchesSearchResults extends React.Component {
 
         let className = ClassNames("branches-branchstatistic-textunit", {"branches-branchstatistic-textunit-open": this.isBranchStatisticOpen(branchStatistic)});
 
-        return (<Collapse in={this.isBranchStatisticOpen(branchStatistic)}>
+        return (<Collapse in={this.isBranchStatisticOpen(branchStatistic)} key={"branchStatisticTextUnit-" + branchTextUnitStatistic.id}>
```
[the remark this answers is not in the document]
montedonio-pinterest: React requires each item iterated inside of a .map function to have a key prop. I put this here to clean console warnings

webapp/src/main/resources/public/js/stores/PaginatorStore.js
montedonio-pinterest: There was a console warning that the prop was expecting a string but got number (or vice versa). This ensures that the expected type will be received

webapp/src/main/resources/public/js/stores/branches/BranchesHistoryStore.js
montedonio-pinterest: No, we are using react state to control the filters in the other filter, but are not here. I had to change some settings to get the page behavior to work correctly

PR #817, opened 2022-06-22
createdBefore & after filters in branches page

montedonio-pinterest:
- createdBefore & after filters in branches page
- cleanup code & remove warns from console
    - add key prop to .map functions in render method
    - component prop expected string but got num
    - component prop expected value but got null at page rendering

### DIFF
--- a/webapp/src/main/resources/public/js/stores/PaginatorStore.js
+++ b/webapp/src/main/resources/public/js/stores/PaginatorStore.js
@@ -1,10 +1,9 @@
-
 class PaginatorStore {
 
     constructor(actions) {
         this.setDefaultState();
     }
-    
+
     setDefaultState() {
         this.currentPageNumber = 1;
         this.hasNextPage = true;
@@ -16,7 +15,7 @@ class PaginatorStore {
     resetSearchParams() {
         this.setDefaultState();
     }
-    
+
     goToNextPage() {
         if (this.hasNextPage) {
             this.currentPageNumber++;
@@ -51,7 +50,7 @@ class PaginatorStore {
         this.disabled = false;
         this.shown = false;
     }
-   
+
 }
 
 export default PaginatorStore;
--- a/webapp/src/main/resources/public/js/stores/branches/BranchesHistoryStore.js
+++ b/webapp/src/main/resources/public/js/stores/branches/BranchesHistoryStore.js
@@ -45,6 +45,14 @@ class BranchesHistoryStore {
         this.openBranchStatistic = openBranchStatistic;
     }
 
+    changeCreatedBefore(createdBefore) {
+        this.createdBefore = createdBefore
+    }
+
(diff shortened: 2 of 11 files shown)


webapp/src/main/resources/public/js/stores/PaginatorStore.js
```diff
@@ -34,7 +34,9 @@ class PaginatorStore {
     }
 
     changeCurrentPageNumber(currentPageNumber) {
-        this.currentPageNumber = currentPageNumber;
+        this.currentPageNumber = currentPageNumber
```
garionpin, line 37: Do you have any context why this change is needed (were we not displaying the page number in certain scenarios?)?

garionpin: Thanks for clarifying!

webapp/src/main/resources/public/js/stores/branches/BranchesHistoryStore.js
```diff
@@ -45,6 +46,20 @@ class BranchesHistoryStore {
         this.openBranchStatistic = openBranchStatistic;
     }
 
+    changeCreatedBefore(createdBefore) {
+        this.createdBefore = createdBefore
```
aurambaj, line 50: all other methods are simple setters so this should stay a simple setter as well. What call sites needs to be updated to mimic that behavior if we're keeping a normal setter?

webapp/src/main/resources/public/js/stores/branches/BranchesHistoryStore.js
```diff
@@ -119,6 +135,12 @@ class BranchesHistoryStore {
         BranchesSearchParamsActions.changeEmpty(empty === "true");
         BranchesSearchParamsActions.changeNotEmpty(notEmpty === "true");
         BranchesSearchParamsActions.changeOnlyMyBranches(onlyMyBranches === "true");
+        if (createdBefore !== null) {
+            BranchesSearchParamsActions.changeCreatedBefore(moment(createdBefore))
```
aurambaj, line 139: looks like that's where we want to apply those conversion. similar to what we do for `selectedBranchTextUnitIds`

webapp/src/main/resources/public/js/stores/branches/BranchesHistoryStore.js
montedonio-pinterest: Was able to simplify these setters by handling the value as strings and letting the Datetime component convert to the moment object and handle the parsing (which it already did by default 😅 )

webapp/src/main/resources/public/js/stores/branches/BranchesHistoryStore.js
montedonio-pinterest: Handling as strings now, don't even need to convert!

aurambaj (approved):
Looks good!

PR #818, opened 2022-06-23
Change xliff refs from v2.0 to v1.2

montedonio-pinterest:


cli/src/test/resources/com/box/l10n/mojito/cli/command/DropExportCommandTest_IO/export/input/source/source-xliff.xliff
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<xliff xmlns="urn:oasis:names:tc:xliff:document:2.0" version="2.0">
```
aurambaj, line 2: looks like all other `1.2` files use `<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:okp="okapi-framework:xliff-extensions">` right now, I'd suggested reusing the same

Issue #811, opened 2022-05-20
XLIFF version 2


danjor:
Hello,

I tried to push an xliff file version 2 to but it failed.
**Is Xliff version 2 supported ?**

https://www.mojito.global//docs/refs/mojito-file-formats/
This page indicated supported files, and for xliff, the example is an xml that have 
``xmlns="urn:oasis:names:tc:xliff:document:2.0" version="2.0"``
but the content looks like xliff 1.2 and not 2.0
because it contains trans-unit tags instead of unit tags, does not have segment tag, etc.

![image](https://user-images.githubusercontent.com/3473673/169499838-84ab3387-511d-43b9-94b6-9e1f76a2abad.png)

xliff 1.2 
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
  <file source-language="en" datatype="plaintext" original="ng2.template">
    <body>
      <trans-unit id="1014715239941102606" datatype="html">
        <source>Remove <x id="linkedMethodName" equiv-text="this.linkedMethods.value[index].name"/></source>
        <note priority="1" from="description">Action verb</note>
      </trans-unit>
    </body>
  </file>
</xliff>
```

xliff 2
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<xliff version="2.0" xmlns="urn:oasis:names:tc:xliff:document:2.0" srcLang="en">
  <file id="ngi18n" original="ng.template">
    <unit id="1014715239941102606">
      <segment>
        <source>Remove <ph id="0" equiv="linkedMethodName" disp="this.linkedMethods.value[index].name"/></source>
      </segment>
    </unit>
  </file>
</xliff>
```

All xliff files in the code used for tests are wrong also, as they declared to use version 2 and related schema but contain 1.2 tags.

aurambaj:
@danjor indeed, the documentation snippet is wrong. Mojito supports 1.2 as provided by the [okapi filter ](https://okapiframework.org/wiki/index.php/XLIFF_Filter).

@montedonio-pinterest could you review / try to fix the XML headers so it is 1.2 in all XLIFF? and if there is no issue update the documentation?

PR #820, opened 2022-06-23
change xliff refs from v1.0 to v1.2

montedonio-pinterest:
Resolving issue #811

### DIFF
--- a/docs/_docs/refs/mojito-file-formats.md
+++ b/docs/_docs/refs/mojito-file-formats.md
@@ -320,7 +320,7 @@ Source Resource File (English): `resource.xliff`
 
 ```xml
 <?xml version="1.0" encoding="UTF-8"?>
-<xliff xmlns="urn:oasis:names:tc:xliff:document:2.0" version="2.0">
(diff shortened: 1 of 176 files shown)


docs/_docs/refs/mojito-file-formats.md
```diff
@@ -320,34 +320,34 @@ Source Resource File (English): `resource.xliff`
 
 ```xml
 <?xml version="1.0" encoding="UTF-8"?>
-<xliff xmlns="urn:oasis:names:tc:xliff:document:2.0" version="2.0">
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:okp="okapi-framework:xliff-extensions" version="1.2">
   <file original="" source-language="en" datatype="x-undefined">
     <body>
-      <trans-unit id="1" resname="hello" datatype="php">
+      <trans-unit id="hi">
```
aurambaj, line 326: how did this get changed?

montedonio-pinterest: @aurambaj wasnt sure if resname and datatype were part of the 1.2 version so I just left the id prop. Just checked the specs and they do, so I'll place them back

PR #824, opened 2022-06-30
Fixes for the DoNotTranslate bug in Mojito Workbench

baktayPin:
Fixes for the bug where setting text unit state to "do not translate = false" doesn't seem to work in workbench. 

The fixes include:
- UI to show the 'do not translate' state of the selected text units (instead of showing a default value, which is confusing for the user)
- Updating the TextUnitDTOsCacheService so that we include the changes in "Do Not Translate" while comparing the cache to the database values.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/tm/textunitdtocache/TextUnitDTOsCacheService.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/tm/textunitdtocache/TextUnitDTOsCacheService.java
@@ -216,8 +216,8 @@ public class TextUnitDTOsCacheService {
     Stream<Long> getTmTextUnitIdsForChangedTranslateStatus(ImmutableSet<Long> idsOfDoNotTranslateTextUnits, ImmutableList<TextUnitDTO> textUnitDTOsToUpdate) {
         return textUnitDTOsToUpdate.stream()
                 .filter(t -> {
-                    boolean doNotTranslateCurrent = idsOfDoNotTranslateTextUnits.contains(t.getTmTextUnitId());
-                    return doNotTranslateCurrent != t.isDoNotTranslate();
+                    boolean newDoNotTranslate = idsOfDoNotTranslateTextUnits.contains(t.getTmTextUnitId());
+                    return newDoNotTranslate != t.isDoNotTranslate();
                 })
                 .map(TextUnitDTO::getTmTextUnitId);
     }


webapp/src/main/java/com/box/l10n/mojito/service/tm/textunitdtocache/TextUnitDTOsCacheService.java
```diff
@@ -211,6 +213,17 @@ Stream<Long> getTmTextUnitIdsForNewTranslations(List<TMTextUnitCurrentVariantDTO
                 .map(TMTextUnitCurrentVariantDTO::getTmTextUnitId);
     }
 
+    Stream<Long> getTmTextUnitIdsForChangedTranslateStatus(ImmutableSet<Long> idsOfDoNotTranslateTextUnits, ImmutableList<TextUnitDTO> textUnitDTOsToUpdate) {
+        return textUnitDTOsToUpdate.stream()
+                .filter(t -> {
+                    boolean doNotTranslateCurrent = idsOfDoNotTranslateTextUnits.contains(t.getTmTextUnitId());
+                    boolean doNotTranslateInNew = t.isDoNotTranslate();
+
+                    return (doNotTranslateCurrent && !doNotTranslateInNew || (!doNotTranslateCurrent && doNotTranslateInNew));
```
aurambaj, line 222: can it be simplified in

```
boolean newDoNotTranslate = idsOfDoNotTranslateTextUnits.contains(t.getTmTextUnitId());
return newDoNotTranslate != t.isDoNotTranslate();
```

webapp/src/test/java/com/box/l10n/mojito/service/asset/VirtualAssetServiceTest.java
```diff
@@ -207,6 +207,54 @@ public void testGetTextunitWithDoNoTranslateFilter() throws Exception {
...
 
+    @Test
+    public void testResetOfDoNoTranslateFilter() throws Exception {
+        Repository repository = repositoryService.createRepository(testIdWatcher.getEntityName("testResetOfDoNoTranslateFilter"));
+        VirtualAsset virtualAsset = new VirtualAsset();
+        virtualAsset.setRepositoryId(repository.getId());
+        virtualAsset.setPath("default");
+        virtualAsset = virtualAssetService.createOrUpdateVirtualAsset(virtualAsset);
+
+        virtualAssetService.addTextUnit(
```
aurambaj, line 218: Does that test reproduce the issue if ran against the old implementation?

webapp/src/main/java/com/box/l10n/mojito/service/tm/textunitdtocache/TextUnitDTOsCacheService.java
baktayPin: Yeah I was thinking whether I should simplify it that way, but I think it makes it not easy to read. What do you think?

aurambaj: the simpler the better also to have a consistent use of "new" with the other function. "new" meaning the state from the DB (vs the state from the blob t.xyz())

webapp/src/test/java/com/box/l10n/mojito/service/asset/VirtualAssetServiceTest.java
baktayPin: I thought it failed, but checking it again it doesn't seem to! I'll taking a look at it

webapp/src/main/java/com/box/l10n/mojito/service/tm/textunitdtocache/TextUnitDTOsCacheService.java
baktayPin: @aurambaj shouldn't it be current actually (as opposed to new) as the DB is only be updated after we detect the change?

webapp/src/test/java/com/box/l10n/mojito/service/asset/VirtualAssetServiceTest.java
baktayPin: Updated the UT, now it does fail without my changes.

webapp/src/main/java/com/box/l10n/mojito/service/tm/textunitdtocache/TextUnitDTOsCacheService.java
baktayPin: oh true, okay updated it!

aurambaj (approved):

PR #830, opened 2022-07-07
Upgrade spring-boot-start-parent to 2.5.14

montedonio-pinterest:
- Fixes vulnerability from `org.springframework:spring-beans | 5.2.9.RELEASE (CVE-2022-22965)`
    - `org.springframework` package upgraded to `5.3.20`
- Fixes import for `org.junit`
- Mark default app.properties classpaths for tests as optional so as to not break the CI which should instead load app-test.properties

### DIFF
--- a/test-common/pom.xml
+++ b/test-common/pom.xml
@@ -35,6 +35,17 @@
             <artifactId>assertj-core</artifactId>
             <version>3.16.1</version>
         </dependency>
+
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.hamcrest</groupId>
+                    <artifactId>hamcrest-core</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
     </dependencies>
 
 </project>
--- a/webapp/pom.xml
+++ b/webapp/pom.xml
@@ -225,12 +225,6 @@
         </dependency>
 
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
             <groupId>com.box.l10n.mojito</groupId>
             <artifactId>mojito-common</artifactId>
             <version>0.111-SNAPSHOT</version>
@@ -439,7 +433,7 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <!-- Sets the VM argument line used when unit tests are run. -->
-                    <argLine>-Dspring.config.additional-location=file://${user.home}/.l10n/config/webapp/
+                    <argLine>-Dspring.config.additional-location=optional:file://${user.home}/.l10n/config/webapp/
                         -Dspring.profiles.active=${user.name},test -Xmx1024m -Duser.timezone=UTC
                     </argLine>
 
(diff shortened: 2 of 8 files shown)


webapp/pom.xml
```diff
@@ -434,27 +434,6 @@
                 </executions>
             </plugin>
 
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-surefire-plugin</artifactId>
-                <configuration>
-                    <!-- Sets the VM argument line used when unit tests are run. -->
-                    <argLine>-Dspring.config.additional-location=file://${user.home}/.l10n/config/webapp/
-                        -Dspring.profiles.active=${user.name},test -Xmx1024m -Duser.timezone=UTC
```
aurambaj, line 443: - the memory and timezone should be at the very least kept
- what's the "new" way of doing what was done with `spring.config.additional-location`? ideally we should replace it, not remove it.

test-common/pom.xml
```diff
@@ -35,6 +35,10 @@
             <artifactId>assertj-core</artifactId>
             <version>3.16.1</version>
         </dependency>
+      <dependency>
+        <groupId>junit</groupId>
```
aurambaj, line 39: they removed from the transitive dependencies? where was it coming from before?

webapp/pom.xml
garionpin: The recommended way for files that might not exist is to use the "optional:" qualifier in the file path https://docs.spring.io/spring-boot/docs/current/reference/html/features.html

This configuration should be kept - and optional should be used on the path.

aurambaj: agree let's try to use: `optional:-Dspring.config.additional-location=file://${user.home}/.l10n/config/webapp/...`

test-common/pom.xml
aurambaj: actually from the release note it seems the recommended way is this: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.4-Release-Notes#junit-5s-vintage-engine-removed-from-spring-boot-starter-test. Can you update accordingly? Thanks!

webapp/pom.xml
garionpin: *or -Dspring.config.additional-location=optional:file://${user.home}/.l10n/config/webapp/...
(whatever the syntax is)

webapp/src/main/resources/config/application.properties
```diff
@@ -36,8 +36,11 @@ spring.jpa.properties.org.hibernate.envers.track_entities_changed_in_revision=tr
 #spring.jpa.open-in-view=false
 
 spring.datasource.url=jdbc:hsqldb:mem:testdb;DB_CLOSE_DELAY=-1
-spring.datasource.initialization-mode=embedded
-spring.datasource.data=classpath:/db/hsql/data.sql
+#spring.datasource.initialization-mode=embedded
```
aurambaj, line 39: remove commented

test-common/pom.xml
```diff
@@ -35,6 +35,17 @@
             <artifactId>assertj-core</artifactId>
             <version>3.16.1</version>
         </dependency>
+
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
```
aurambaj, line 40: can you add a comment in the commit description and point to release not for future reference ... i'm sure will need it!

aurambaj: in fact i quickly tried on my side, i doesn't seem to be needed. Please try without it now.

PR #835, opened 2022-08-03
Delta generation service

garionpin:
Delta generation service providing a time range generation method for
translation deltas as well as a push/pull run state-based generation
method.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/rest/commit/CommitWS.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/rest/commit/CommitWS.java
@@ -1,6 +1,8 @@
 package com.box.l10n.mojito.rest.commit;
 
 import com.box.l10n.mojito.entity.Commit;
+import com.box.l10n.mojito.entity.PullRun;
+import com.box.l10n.mojito.entity.PushRun;
 import com.box.l10n.mojito.entity.Repository;
 import com.box.l10n.mojito.rest.PageView;
 import com.box.l10n.mojito.rest.View;
@@ -13,7 +15,6 @@ import com.box.l10n.mojito.service.repository.RepositoryRepository;
 import com.fasterxml.jackson.annotation.JsonView;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort;
@@ -37,11 +38,14 @@ public class CommitWS {
      */
     static Logger logger = LoggerFactory.getLogger(CommitWS.class);
 
-    @Autowired
     RepositoryRepository repositoryRepository;
 
-    @Autowired
     CommitService commitService;
+
+    public CommitWS(RepositoryRepository repositoryRepository, CommitService commitService) {
+        this.repositoryRepository = repositoryRepository;
+        this.commitService = commitService;
+    }
 
     /**
      * Gets information about commits based on the specified search criteria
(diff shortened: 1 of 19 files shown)


webapp/src/main/java/com/box/l10n/mojito/rest/commit/CommitWS.java
```diff
@@ -98,17 +100,72 @@ public Page<Commit> getCommits(@RequestParam(value = "repositoryId") Long reposi
...
      */
     @JsonView(View.Commit.class)
     @RequestMapping(value = "/api/commits/lastPushed/", method = RequestMethod.POST)
-    public Commit getLastPushedCommit(@RequestBody LastPushedCommitBody lastPushedCommitBody) {
-        return commitService.getLastPushedCommit(lastPushedCommitBody.getCommitNames(),
-                                                 lastPushedCommitBody.getRepositoryId())
+    public Commit getLastPushedCommit(@RequestBody CommitListWithRepositoryIdBody commitListWithRepositoryIdBody) {
+        return commitService.getLastPushedCommit(commitListWithRepositoryIdBody.getCommitNames(),
+                                                 commitListWithRepositoryIdBody.getRepositoryId())
+                .orElse(null);
```
aurambaj, line 115: what is this actually going to return? Should it be remapped to an HTTP error?

webapp/src/main/java/com/box/l10n/mojito/rest/commit/CommitWS.java
```diff
@@ -98,17 +100,72 @@ public Page<Commit> getCommits(@RequestParam(value = "repositoryId") Long reposi
...
+     *
+     * @param commitListWithRepositoryIdBody The commit names to search for together with the repositoryID.
+     * @return {@link View.Commit}
+     */
+    @JsonView(View.CommitDetailed.class)
+    @RequestMapping(value = "/api/commits/lastPushRun/", method = RequestMethod.POST)
+    public PushRun getLastPushRun(@RequestBody CommitListWithRepositoryIdBody commitListWithRepositoryIdBody) {
+        return commitService.getLastPushRun(commitListWithRepositoryIdBody.getCommitNames(),
+                                                 commitListWithRepositoryIdBody.getRepositoryId())
+                .orElse(null);
```
aurambaj, line 133: Spring doesn't handle optional out of the box?

webapp/src/main/java/com/box/l10n/mojito/service/delta/DeltaService.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Service that enables delta generation.
+ *
+ * @author garion
+ */
+@Service
+public class DeltaService {
+    final
```
aurambaj, line 34: in most places we don't use final, not saying it is better but maybe for consistency we should keep doing that

webapp/src/main/java/com/box/l10n/mojito/service/delta/DeltaService.java
```diff
@@ -0,0 +1,192 @@
...
+     *      as before.
+     *
+     * @return The delta of text unit variants, their translations and corresponding metadata.
+     */
+    public DeltaResponseDTO getDeltasForRuns(Repository repository,
+                                             List<Locale> locales,
+                                             List<PushRun> pushRuns,
+                                             List<PullRun> pullRuns) {
+
+        List<Long> localeIds = locales == null ? Collections.emptyList() :
```
aurambaj, line 121: Optional.ofNullable(..) to replace the null check in the following lines?

webapp/src/main/java/com/box/l10n/mojito/service/delta/DeltaService.java
```diff
@@ -0,0 +1,192 @@
...
+
+        DeltaResponseDTO deltaResponseDTO = new DeltaResponseDTO();
+        deltaResponseDTO.setTranslationsPerLocale(deltaLocaleDataByBcp47Tags);
+        deltaResponseDTO.setDeltaMetadataDTO(deltaMetadataDTO);
+
+        return deltaResponseDTO;
+    }
+
+    private TextUnitVariantDelta getTextUnitVariantDelta(TMTextUnitVariant tmTextUnitVariant) {
+        return new TextUnitVariantDelta() {
```
aurambaj, line 146: might be better to have an actual implementation vs the annonymous class

webapp/src/main/java/com/box/l10n/mojito/service/tm/TMTextUnitVariantRepository.java
```diff
@@ -17,9 +27,130 @@ public interface TMTextUnitVariantRepository extends JpaRepository<TMTextUnitVar
...
+     * {@link PullRun}.
+     * <p>
+     * Additional filtering done by {@link Repository} ID and a list of
+     * {@link Locale} IDs.
+     * <p>
+     * To note: unused text units and deleted assets are also included,
+     * as this method is indented to provided deltas relevant for a specific snapshot.
+     */
+    @Query(nativeQuery = true,
+            value = "select distinct base_tu.name as tmTextUnitName, " +
```
aurambaj, line 114: do you have a sense of the performance impact of that query?

aurambaj (commented):
Looks good!

webapp/src/main/java/com/box/l10n/mojito/service/delta/DeltaService.java
garionpin: That's the default transformation that the latest version of IntelliJ applies when moving from prop to ctor injection.

It seems like good practice to use final as these properties should never be assigned outside the ctor - will remove the final keywords for now for the sake of consistency.

webapp/src/main/java/com/box/l10n/mojito/service/delta/DeltaService.java
garionpin: Not sure if I understood this, do you mean doing something like?
`        List<Long> localeIds = Optional.ofNullable(locales)
                        .map(l -> locales.stream().map(BaseEntity::getId).collect(Collectors.toList()))
                        .orElse(Collections.emptyList());`

That seems quite a bit more verbose than the current implementation.

webapp/src/main/java/com/box/l10n/mojito/rest/commit/CommitWS.java
garionpin: This should not return an error (404) as it's expected that no valid commit exists for the list of commit names provided. (E.g.: I could pass in a list of local commit hashes that Mojito never processed a push on - that's not an error situation).
When a valid commit is not found, the client is expected to loop back and call this API with additional older commit names until something is found (or a max threshold is reached).

This is the same scenario as searching the Mojito workbench / text units search API for content that doesn't exist.

webapp/src/main/java/com/box/l10n/mojito/rest/commit/CommitWS.java
garionpin: We never return Optional<T> in our API signatures - planning to stick with the same for consistency and not mixing return types.

webapp/src/main/java/com/box/l10n/mojito/service/tm/TMTextUnitVariantRepository.java
garionpin: Not empirically, though the query is optimized as much as it can be for now with predicates being pushed down for both subqueries, the number of rows returned should be quite small in practice.

Once we have production data from pull on large repos - we can measure and tweak this further.

webapp/src/main/java/com/box/l10n/mojito/service/delta/DeltaService.java
garionpin: Done

webapp/src/main/java/com/box/l10n/mojito/service/delta/DeltaService.java
aurambaj: Similar. I'd start with something like that (which could be put into a static helper actually since it is doing it 3 times in a row) to get a stream from a nullable list

```
Optional.ofNullable(locales).map(List::stream).orElseGet(Stream::empty);
```

and have the specific code

garionpin: I've refactor the common stream code/ID parsing now.

PR #838, opened 2022-08-08
Added machine translation retrieval to chrome extension

maallen:
Added functionality to the chrome extension to retrieve machine translations for flagged strings by calling to a remote endpoint that takes the requested locale and message as query parameters in a url.

The MT url format should be provided in the background configuration with `{mt_locale}` and `{mt_message}` as placeholders in the url which will be replaced with the locale and message required for machine translation.

### DIFF
--- a/chromeextension/background.js
+++ b/chromeextension/background.js
@@ -3,7 +3,8 @@ var backgroundConfig = {
     enabled: false,
     headerName: 'X-Mojito-Ict',
     headerValue: 'on',
-    actionButtons: []
+    actionButtons: [],
+    mtEndpointUrlFormat: ''
 };
 
 chrome.storage.sync.get(backgroundConfig, (items) => {
@@ -23,6 +24,36 @@ function addMojitoIctHeader(details) {
     }
     return {requestHeaders: details.requestHeaders};
 }
+
+chrome.runtime.onMessage.addListener(function(request, sender, sendResponse) {
+    if (backgroundConfig.mtEndpointUrlFormat) {
+        var message = request['textUnitVariant']
+        var locale = request['locale']
+        var storageKey = "I18N_MT_" + locale + "_" + message
+        chrome.storage.local.get([storageKey], function (result){
+            if (result[storageKey]) {
+                // Retrieve translation from local storage
+                sendResponse({data: result[storageKey]})
+            } else {
+                var requestUrl = backgroundConfig.mtEndpointUrlFormat.replace("{mt_locale}", locale).replace("{mt_message}", message);
+                fetch(requestUrl)
+                    .then(response => response.json())
+                    .then(json => {
+                        chrome.storage.local.set({[storageKey]: json.data}, function (){
+                            // Store translation in local storage for future retrieval
+                            sendResponse(json)
+                        })
+                    })
+                    .catch(err => sendResponse({data: ''}));
+            }
+        })
+
+        return true;
+    } else {
+        sendResponse({data: ''});
+    }
+
+});
 
 chrome.webRequest.onBeforeSendHeaders.addListener(
     addMojitoIctHeader,
--- a/webapp/src/main/resources/public/js/ict/Ict.js
+++ b/webapp/src/main/resources/public/js/ict/Ict.js
@@ -92,44 +92,40 @@ class Ict {
             node = node.parentNode;
         }
 
-        var mtRequired = textUnits[0]['isTranslated'] === 'false';
-        if (this.mtEnabled && mtRequired){
-            if (!node.classList.contains("mojito-ict-string")) {
-                try {
-                    node.className += " mojito-ict-string";
-                } catch (e) {
-                }
 
-                node.addEventListener("mouseenter", (e) => {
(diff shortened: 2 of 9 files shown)


chromeextension/background.js
```diff
@@ -24,6 +25,22 @@ function addMojitoIctHeader(details) {
     return {requestHeaders: details.requestHeaders};
 }
 
+chrome.runtime.onMessage.addListener(function(request, sender, sendResponse) {
+    if(backgroundConfig.mtEndpointUrlFormat) {
```
aurambaj, line 29: indentation

webapp/src/main/resources/public/js/ict/IctMetadataExtractor.js
```diff
@@ -37,10 +37,13 @@ class IctMetadataExtractor {
                         textUnit['textUnitName'] = spitTextUnitMetadata[2];
                         textUnit['locale'] = spitTextUnitMetadata[3];
                         textUnit['stack'] = spitTextUnitMetadata[4];
+                        textUnit['mtRequired'] = spitTextUnitMetadata[5];
+                        textUnit['mtLocale'] = spitTextUnitMetadata[6];
```
aurambaj, line 41: how is "mtLocale" different from "locale"?

webapp/src/main/resources/public/js/ict/IctMetadataExtractor.js
```diff
@@ -37,10 +37,13 @@ class IctMetadataExtractor {
...
                     } catch (e) {
                         console.log("Can't extract ict metadata from string: ", string, e);
                     }
 
+
```
aurambaj, line 46: remove?

webapp/src/main/resources/public/js/ict/IctMetadataExtractor.js
maallen: I meant to update this, there was handling in the backend endpoint that converted pinterest locale to the mojito locale, if we were already in the mojito locale the map lookup failed and the original string got returned. I've updated the endpoint logic this morning to allow both pinterest & mojito locales so will remove the mtLocale field as it was a workaround to verify against the backend servers existing endpoint

chromeextension/background.js
```diff
@@ -24,6 +25,22 @@ function addMojitoIctHeader(details) {
     return {requestHeaders: details.requestHeaders};
 }
 
+chrome.runtime.onMessage.addListener(function(request, sender, sendResponse) {
+    if (backgroundConfig.mtEndpointUrlFormat) {
+        var message = request['textUnitName'].split('---')[0]
```
aurambaj, line 30: oh actually it should be ['textUnitVariant'] (`---` is gettext specific)

webapp/src/main/resources/public/js/ict/Ict.js
```diff
@@ -87,19 +92,44 @@ class Ict {
...
+
+                node.addEventListener("mouseleave", (e) => {
+                    e.target.classList.remove("mojito-ict-string-mt");
+                });
             }
 
-            node.addEventListener("mouseenter", (e) => {
-                e.target.classList.add("mojito-ict-string-active");
-            });
+            chrome.runtime.sendMessage(textUnits[0], function (response){
```
aurambaj, line 112: indent `function (response){`

refactor suggestion, probably want to share the event registration code?

```
var mtRequired = this.mtEnabled && textUnits[0]['isTranslated'] === 'false';
eventListenerClassName = mtRequired  ? "mojito-ict-string-mt" : "mojito-ict-string-active"

if (!node.classList.contains("mojito-ict-string")) {
      try {
          node.className += " mojito-ict-string";
      } catch (e) {
      }

      node.addEventListener("mouseenter", (e) => {
          e.target.classList.add(eventListenerClassName);
      });

      node.addEventListener("mouseleave", (e) => {
          e.target.classList.remove(eventListenerClassName);
      });
  }

if (mtRequired) {
    chrome.runtime.sendMessage(textUnits[0], function (response){ 
...

```

webapp/src/main/resources/public/js/ict/IctModal.js
```diff
@@ -21,6 +21,7 @@ class ModalTextUnit extends React.Component {
         return (
             <div className={className} 
                  onClick={() => this.props.onSelect(textUnit)}>
+                {textUnit.isMachineTranslated && <div className="mbs">This text unit has been machine translated. Please wait for final translations before committing your changes!</div> }
```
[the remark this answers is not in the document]
maallen: Good point on the committing changes message, I'll remove it. I haven't really looked much into adding the overlay icon as we now wrap any MT'd strings in an orange box on hover, I can take a look into appending some HTML to the DOM node if we MT the string to make it more clear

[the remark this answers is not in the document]
maallen: Updated to append an orange asterisk to machine translated strings to provide a visual signal that a string has been machine translated
Screenshot 2022-08-11 at 15 44 01
.

PR #840, opened 2022-08-10
Migrate extension to manifest v3

maallen:
Upgraded the chrome extension to work with Manifest V3 changes.

### DIFF
--- a/chromeextension/background.js
+++ b/chromeextension/background.js
@@ -7,6 +7,32 @@ var backgroundConfig = {
     mtEndpointUrlFormat: ''
 };
 
+function addMojitoIctHeader() {
+    // First removes any pre-existing rule with id 1, then adds a new one with same id. The remove
+    // occurs before the addition operation as documented at
+    // https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/#method-updateDynamicRules
+    chrome.declarativeNetRequest.updateDynamicRules({
+        addRules:[{
+            "id": 1,
+            "priority": 1,
+            "action":  {
+                "type": "modifyHeaders",
+                "requestHeaders": [
+                    {"header": backgroundConfig.headerName, "operation": "set", "value": backgroundConfig.headerValue}
+                ]
+            },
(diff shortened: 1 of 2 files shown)


webapp/src/main/resources/public/js/app.js
```diff
@@ -85,7 +85,7 @@ function instrumentMessagesForIct(messages, locale) {
 
     Object.keys(messages).map((key) => {
         let stack = new Error().stack; // stack is useless here but for tests
-        messages[key] = IctMetadataBuilder.getTranslationWithMetadata("mojito", null, key, locale, stack, messages[key]);
+        messages[key] = IctMetadataBuilder.getTranslationWithMetadata("mojito", null, key, locale, stack, messages[key], true);
```
aurambaj, line 88: `messages[key] != null` or whatever JS returns in that case ?

webapp/src/main/resources/public/js/ict/Ict.js
```diff
@@ -80,25 +85,46 @@ class Ict {
...
+        if (mtRequired) {
+            chrome.runtime.sendMessage(textUnits[0], function (response) {
+                if (response.data && node.textContent !== response.data) {
+                    if (node.nodeType === Node.ELEMENT_NODE && node.getAttribute('placeholder')) {
+                        node.setAttribute('placeholder', response.data)
+                    } else {
+                        textUnits[0]['isMachineTranslated'] = true
+                        node.textContent = response.data;
+                        node.classList.add("mojito-mt-ict-string-static");
+                        node.addEventListener("click", (e) => addMTCSS(e, node))
```
aurambaj, line 123: i'm not sure to understand the event listeners below. Why do we need them?

maallen: I had an issue where if you clicked an MT'd text unit than the CSS MT label was removed from the content, I suspect it's to do with React re-rendering the page which updates the content and removes the CSS appended 'MT' label

PR #843, opened 2022-08-16
Ota proto

aurambaj:


webapp/pom.xml
```diff
@@ -24,6 +24,12 @@
 
     <dependencies>
 
+        <dependency>
+            <groupId>p6spy</groupId>
```
aurambaj, line 28: remove

webapp/src/main/java/com/box/l10n/mojito/entity/PullRunTextUnitVariant.java
```diff
@@ -22,12 +22,7 @@
  * @author garion
  */
 @Entity
-@Table(
-        name = "pull_run_text_unit_variant",
-        indexes = {
-                @Index(name = "UK__PULL_RUN_TEXT_UNIT_VARIANT__EVA_ID__TM_TUV_ID", columnList = "pull_run_asset_id, tm_text_unit_variant_id", unique = true)
```
aurambaj, line 28: remove the constraint since we can have duplicated tm text unit variant id. 

If we want to put it back we need to add a `locale_id` column

webapp/src/main/java/com/box/l10n/mojito/service/pullrun/PullRunAssetService.java
```diff
@@ -1,26 +1,172 @@
...
+     * Ran locally this was very slow.. while we expected it to be the faster option
+     */
+    @StopWatch
+    @Transactional
+    public void saveTextUnitVariantsServer(PullRunAsset pullRunAsset, List<Long> tuvIds, Long localeId) {
+        Repository repository = pullRunAsset.getPullRun().getRepository();
+        JdbcTemplate jdbcTemplate = new JdbcTemplate(dataSource);
+        meterRegistry.timer("PullRunAssetService.saveTextUnitVariantsMultiRow", Tags.of("repositoryId", Objects.toString(repository.getId())))
+                .record(() -> {
+                    // TODO(jean) might be the fastest option (if data is clean on the server side),
```
aurambaj, line 117: @geoarion  on my localhost it turned out to be slow ... but everything was slow compared to my original test

webapp/src/main/java/com/box/l10n/mojito/service/pullrun/PullRunAssetService.java
```diff
@@ -1,26 +1,172 @@
...
+            pullRunTextUnitVariant.setTmTextUnitVariant(entityManager.getReference(TMTextUnitVariant.class, tuvId));
+            pullRunTextUnitVariantRepository.save(pullRunTextUnitVariant);
+        }
+        entityManager.flush();
+        entityManager.clear();
+    }
+
+    private void saveTextUnitVariantsPreparedStatementBatch(JdbcTemplate jdbcTemplate, PullRunAsset pullRunAsset, List<Long> tuvIds, Instant now) {
+        // TODO(jean) is providing the date faster?
+        jdbcTemplate.batchUpdate("insert into pull_run_text_unit_variant (pull_run_asset_id, tm_text_unit_variant_id, created_date) values (?, ?, ?)", new BatchPreparedStatementSetter() {
```
aurambaj, line 144: @garionpin any feeling on passing the date vs auto generating on the server? Obviously the advantage to pass it, is to have the a single value

webapp/src/main/java/com/box/l10n/mojito/service/pullrun/PullRunAssetService.java
```diff
@@ -1,26 +1,172 @@
...
+                return tuvIds.size();
+            }
+        });
+    }
+
+    private void saveTextUnitVariantsMultiRowBatch(JdbcTemplate jdbcTemplate, PullRunAsset pullRunAsset, List<Long> tuvIds, Instant now) {
+        // TODO(jean) is providing the date faster?
+        // TODO(jean) no protection from preparedstatement...
+        String sql = "insert into pull_run_text_unit_variant(pull_run_asset_id, tm_text_unit_variant_id, created_date) values "
+                + tuvIds.stream().map(id -> String.format("(%s, %s, '%s') ", pullRunAsset.getId(), id, Timestamp.from(now))).collect(Collectors.joining(","));
```
aurambaj, line 163: @garionpin added the date... so now it is duplicated in the payload.

webapp/src/main/java/com/box/l10n/mojito/entity/PullRunTextUnitVariant.java
garionpin: Would we not need locale_id added anyways after removing this constraint?
Otherwise - we'd have exactly the same TUVs duplicated - and we wouldn't be able to discern any useful information from the duplicates - if we don't specifically know what locales they were intended for.

Just duplicating the data without a locale_id will bloat the DB.

aurambaj: I agree, I feel like having the `locale_id` makes more sense to cover inheritance. Do we want to denormalize the `tm_text_unit_id` also, we'd have a table pretty similar to the current variants then

cli/src/main/java/com/box/l10n/mojito/cli/command/PushService.java
```diff
@@ -2,13 +2,15 @@
...
 import com.box.l10n.mojito.rest.client.AssetClient;
+import com.box.l10n.mojito.rest.client.CommitClient;
 import com.box.l10n.mojito.rest.client.RepositoryClient;
 import com.box.l10n.mojito.rest.client.exception.PollableTaskException;
 import com.box.l10n.mojito.rest.entity.Branch;
 import com.box.l10n.mojito.rest.entity.PollableTask;
 import com.box.l10n.mojito.rest.entity.Repository;
 import com.box.l10n.mojito.rest.entity.SourceAsset;
 import com.google.common.collect.Sets;
+import org.apache.commons.lang3.StringUtils;
```
aurambaj, line 13: trying to stop using common lang in favor of guava

cli/src/main/java/com/box/l10n/mojito/cli/command/PushCommand.java
```diff
@@ -104,13 +150,16 @@ public void execute() throws CommandException {
...
                     sourceAsset.setFilterConfigIdOverride(sourceFileMatch.getFileType().getFilterConfigIdOverride());
-                    sourceAsset.setFilterOptions(commandHelper.getFilterOptionsOrDefaults(sourceFileMatch.getFileType(), filterOptionsParam));
+                    sourceAsset.setFilterOptions(commandHelper.getFilterOptionsOrDefaults(sourceFileMatch.getFileType(),
+                                                                                          filterOptionsParam));
 
                     return sourceAsset;
                 });
 
         pushService.push(repository, sourceAssetStream, branchName, pushType);
+        pushService.associatePushRun(repository, pushRunName, commitHash);
```
aurambaj, line 162: should this be gated by the `recordPushRun`?

cli/src/main/java/com/box/l10n/mojito/cli/command/PushService.java
```diff
@@ -104,6 +121,13 @@ void waitForPollableTasks(List<PollableTask> pollableTasks) throws CommandExcept
         }
     }
 
+    public void associatePushRun(Repository repository, String pushRunName, String commitHash) {
+        if (!StringUtils.isBlank(pushRunName) && !StringUtils.isBlank(commitHash)) {
```
aurambaj, line 125: should we prevent those to be blank in a first place?

PR #846, opened 2022-08-17
Pull recording

aurambaj:


webapp/src/main/java/com/box/l10n/mojito/service/tm/TMService.java
```diff
@@ -961,15 +980,29 @@ public String generateLocalized(
...
+            saveUsedTmTextUnitVariantIds(asset, pullRunName, repositoryLocale.getLocale(), translateStep.getUsedTmTextUnitVariantIds());
+        }
+
+        return generateLocalizedBase;
+    }
+
+    void saveUsedTmTextUnitVariantIds(Asset asset, String pullRunName, Locale locale, List<Long> usedTmTextUnitVariantIds) {
+        PullRun pullRun = pullRunService.getOrCreate(pullRunName, asset.getRepository());
+        PullRunAsset pullRunAsset = pullRunAssetService.getOrCreate(pullRun, asset);
+        List<Long> uniqueUsedTmTextUnitVariantIds = usedTmTextUnitVariantIds.stream().distinct().collect(Collectors.toList());
```
aurambaj, line 997: @garionpin Not sure if we should record exactly what's used to generate the file (ie potential duplicates, will need to change the UK again then) or if we record distinct entries (current implementation).

cli/src/main/java/com/box/l10n/mojito/cli/command/CommitToPullRunCommand.java
```diff
@@ -0,0 +1,87 @@
...
+    @Autowired
+    ConsoleWriter consoleWriter;
+
+    @Parameter(names = {Param.REPOSITORY_LONG, Param.REPOSITORY_SHORT}, arity = 1, required = true, description = Param.REPOSITORY_DESCRIPTION)
+    String repositoryParam;
+
+    @Parameter(names = {Param.COMMIT_HASH_LONG, Param.COMMIT_HASH_SHORT}, arity = 1, required = true, description = "The commit to map to a pull run")
+    String commitHash;
+
+    @Parameter(names = {"--input-directory", "-i"}, arity = 1, required = false, description = "The input directory where to read the pull run name, filename: pull-run-name.txt)")
```
garionpin, line 42: Should we allow the full path of the file to be specified instead of just the input dir? (--input-file-path)

cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
```diff
@@ -148,9 +156,26 @@ public void execute() throws CommandException {
...
 
+    void initPullRunName() {
+        if (recordPullRun) {
+            pullRunName = UUID.randomUUID().toString();
+        }
+    }
+
+    void writePullRunFileIfNeeded() {
+        if (recordPullRun) {
+            Path pullRunOuputFile = commandDirectories.getTargetDirectoryPath().resolve("pull-run-name.txt");
```
garionpin, line 172: Should the full output file path be explicitly be passed in as a param?

cli/src/test/java/com/box/l10n/mojito/cli/command/PullCommandTest.java
```diff
@@ -896,4 +934,269 @@ public void pullFullyTranslated() throws Exception {
...
+        tmTextUnitCurrentVariantService.removeCurrentVariant(tmTextUnitCurrentVariantRepository.findByLocale_IdAndTmTextUnit_Id(jaJP.getId(), tmTextUnit.getId()).getId());
+
+        logger.debug("Check that the delta for the pull run generated returns 1 modifications");
+        DeltaResponseDTO deltaForAfterFrenchAndJapaneseImportsToSmallChanges = deltaService.getDeltasForRuns(
+                repository,
+                null,
+                ImmutableList.of(pushRun),
+                ImmutableList.of(pullRunAfterFrenchAndJapaneseImports));
+
+        // TODO(jean) We may have agreed on returning nothing for delete and have to insert either a fix or the
```
garionpin, line 1074: Say we return something - would we want to include that in the final output for a downstream client to use? And if so - would we load that up and expect that translation to be displayed to the end user?
I can't think of a scenario atm where that is desirable - and am thinking for that reason to not include any unusable data essentially.

webapp/src/main/java/com/box/l10n/mojito/service/tm/TMService.java
garionpin: Is this comment still applicable, now that you've added locale_id in the table?

garionpin: PS: having the same data recorded multiple times for a locale doesn't provide us with any additional information - I'd advocate for always uploading distinct().

webapp/src/main/resources/db/migration/V61__Refactor_PullRunTextUnitVariant.sql
```diff
@@ -0,0 +1,19 @@
...
+
+ALTER TABLE pull_run_text_unit_variant
+    ADD CONSTRAINT FK__PULL_RUN_TEXT_UNIT_VARIANT__PULL_RUN_ASSET_ID FOREIGN KEY (pull_run_asset_id) REFERENCES pull_run_asset (id);
+
+ALTER TABLE pull_run_text_unit_variant
+    ADD CONSTRAINT FK__PULL_RUN_TEXT_UNIT_VARIANT__TM_TEXT_UNIT_VARIANT_ID FOREIGN KEY (tm_text_unit_variant_id) REFERENCES tm_text_unit_variant (id);
+
+-- assume data has been removed first from pull_run_text_unit_variant
+alter table pull_run_text_unit_variant add column locale_id bigint not null;
+alter table pull_run_text_unit_variant add constraint UK__PULL_RUN_TEXT_UNIT_VARIANT__PRA_ID__TUV_ID__LOCALE_ID unique (pull_run_asset_id, tm_text_unit_variant_id, locale_id);
```
garionpin, line 18: This index would get more use by the query optimizer if locale_id is the second element with tm_text_unit_variant_id being last.
(tm_text_unit_variant_id is already a PK in it's own index based on the FK constraint).

webapp/src/main/java/com/box/l10n/mojito/entity/PullRunTextUnitVariant.java
```diff
@@ -25,7 +26,7 @@
 @Table(
         name = "pull_run_text_unit_variant",
         indexes = {
-                @Index(name = "UK__PULL_RUN_TEXT_UNIT_VARIANT__EVA_ID__TM_TUV_ID", columnList = "pull_run_asset_id, tm_text_unit_variant_id", unique = true)
+                @Index(name = "UK__PULL_RUN_TEXT_UNIT_VARIANT__PRA_ID__TUV_ID__LOCALE_ID", columnList = "pull_run_asset_id, tm_text_unit_variant_id, locale_id", unique = true)
```
garionpin, line 29: Same comment as for v61 - locale_id would help queries most as the 2nd column in the index.

cli/src/main/java/com/box/l10n/mojito/cli/command/CommitToPullRunCommand.java
aurambaj: that was my original intent but then i wanted to avoid having too many parameter until proven necessary. For our current usages, the file will be collocated with the localized files which I think is fine. For a repository where we'd scan from the root it would put that file in a place we probably don't want to have.

cli/src/main/java/com/box/l10n/mojito/cli/command/PullCommand.java
aurambaj: see previous comment.

Code wise I need to share the code at least

cli/src/test/java/com/box/l10n/mojito/cli/command/PullCommandTest.java
aurambaj: the scenario would be an invalid translation that you want to remove. the workaround would be to add the english (or parent) as current translation while re-translation happens but that might be confusing too

webapp/src/main/java/com/box/l10n/mojito/service/tm/TMService.java
aurambaj:
```
yes=it could happen

yes=it could happen
```

cli/src/test/java/com/box/l10n/mojito/cli/command/PullCommandTest.java
```diff
@@ -896,4 +945,360 @@ public void pullFullyTranslated() throws Exception {
...
+        // TODO(jean) so in real scenario will also have to create the commit -- replace with proper command or if inlined
+        String pullRunHash1 = "ddaa11";
+        commitService.saveCommit(repository, pullRunHash1, "coder@mail.com", "coder", DateTime.now());
+        getL10nJCommander().run("commit-to-pull-run", "-r", repository.getName(),
+                "-i", getTargetTestDir("target_translated").getAbsolutePath(),
+                "--commit-hash", pullRunHash1);
+
+        // For language like cs-CZ or ru-RU, PO file use only 3 plural forms while Mojito store 4 form (CLDR).
+        // This leads to the number of recorded entry in the pull run to lower that the number of current translations.
+        // This might look strange but is expected
```
aurambaj, line 1002: @garionpin fyi

PR #847, opened 2022-08-18
Add CLI command to store commit information

garionpin:
Add "commit-create" command to store commit details in the system.

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/CommitCreateCommand.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/CommitCreateCommand.java
@@ -1,0 +1,90 @@
+package com.box.l10n.mojito.cli.command;
+
+import com.beust.jcommander.Parameter;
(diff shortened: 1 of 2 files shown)


maallen (approved):

cli/src/main/java/com/box/l10n/mojito/cli/command/CommitCreateCommand.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+
+/**
+ * @author garion
+ */
+@Component
+@Scope("prototype")
+@Parameters(commandNames = {"commit-create", "cc"}, commandDescription = "Create commit information in Mojito.")
```
aurambaj, line 24: haha, not sure about the short name in that case. I've just drop it for some commands

PR #848, opened 2022-08-19
Adding the Virtual tag to the Text Unit Information View

baktayPin:
Adding the virtual field to the Text Unit View on the Mojito Workbench.

Screenshot 2022-08-22 at 12 40 16


webapp/src/main/java/com/box/l10n/mojito/service/tm/search/TextUnitSearcher.java
```diff
@@ -200,7 +200,8 @@ NativeCriteria getCriteriaForSearch(TextUnitSearcherParameters searchParameters)
                 addProjection("a.path", "assetPath").
                 addProjection("atu.id", "assetTextUnitId").
                 addProjection("tu.created_date", "tmTextUnitCreatedDate").
-                addProjection("atu.do_not_translate", "doNotTranslate")
+                addProjection("atu.do_not_translate", "doNotTranslate").
+                addProjection("a.virtual", "isVirtual")
```
aurambaj, line 204: For newer, less used fields, the idea has been to fetch them in the Gitblame logic to avoid growing the TextUnitSearcher query indefinitely. My recommendation is to update `GitBlameService#getGitBlameWithUsages()` and add one "enrich" function that basically fetch the virtual attribute instead of having it here

baktayPin: Yeah that does make sense. Let me change it 👍

PR #852, opened 2022-08-23
Add data retention job for Push and Pull run recording

garionpin:
Implement a retention policy to ensure old push and pull run data gets
deleted regularly.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/delta/PushPullRunCleanupJob.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/delta/PushPullRunCleanupJob.java
@@ -1,0 +1,71 @@
+package com.box.l10n.mojito.service.delta;
+
+import com.box.l10n.mojito.service.assetExtraction.AssetExtractionCleanupJob;
+import org.quartz.DisallowConcurrentExecution;
+import org.quartz.Job;
+import org.quartz.JobDetail;
+import org.quartz.JobExecutionContext;
+import org.quartz.JobExecutionException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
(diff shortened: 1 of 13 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/delta/PushPullRunCleanupJob.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * @author garion
+ */
+@Profile("!disablescheduling")
+@Configuration
+@Component
+@DisallowConcurrentExecution
+@ConditionalOnProperty(value = "l10n.delta.cleanup-job.enabled", havingValue = "true")
+public class PushPullRunCleanupJob implements Job {
```
aurambaj, line 32: should the property and the class name map?

webapp/src/main/java/com/box/l10n/mojito/service/delta/PushPullRunCleanupService.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import java.time.Duration;
+
+/**
+ * @author garion
+ */
+@Service
+public class PushPullRunCleanupService {
```
aurambaj, line 17: is this is not part of DeltaService, maybe rename the property `l10n.delta.cleanup-job.enabled` then?

aurambaj (approved):

webapp/src/main/java/com/box/l10n/mojito/service/delta/PushPullRunCleanupJob.java
```diff
@@ -0,0 +1,71 @@
...
+import java.time.Duration;
+
+/**
+ * @author garion
+ */
+@Profile("!disablescheduling")
+@Configuration
+@Component
+@DisallowConcurrentExecution
+@ConditionalOnProperty(value = "l10n.PushPullRun.cleanup-job.enabled", havingValue = "true")
```
aurambaj, line 31: probably my bad when i suggested the name. I think other job use `-` like `l10n.asset-extraction.cleanup-job.enabled`. Maybe we can keep it consistent. Merging as it anyway

aurambaj (approved):

PR #854, opened 2022-08-25
UTC fix for Push, Pull run recording and delta generation

garionpin:
Store push and pun run records in UTC format.
Use UTC when querying for state-based deltas.
Use current text unit variant's last_modified_date for date filtering
instead of created_date as existing variants get updated with new
translations keeping the created_date the same on the current variants
table.


aurambaj (approved):
thanks!

PR #857, opened 2022-08-26
Integrate Google Java Format on maven compile

garionpin:
Integrate the GoogleJavaFormat formatter in the project and trigger
the apply goal to auto-fix any issues when doing a maven compile.
This separately also enables formatting for pom.xml files.

The commit for the bulk changes is added to 
.git-blame-ignore-revs to ensure it does not pollute the history / 
git blame.


pom.xml
```diff
@@ -109,6 +111,52 @@
...
+                        </sortPom>
+                    </pom>
+                    <java>
+                        <googleJavaFormat>
+                            <version>${plugin.version.google-java-format}</version>
+                            <style>GOOGLE</style>
+                        </googleJavaFormat>
+                        <importOrder>
+                            <wildcardsLast>false</wildcardsLast>
+                            <order>com.box,com,org,javax,java,,\#,\\#</order>
```
garionpin, line 140: @aurambaj thoughts on import ordering?

pom.xml
```diff
@@ -109,6 +111,52 @@
...
+                    <execution>
+                        <goals>
+                            <goal>apply</goal>
+                        </goals>
+                        <phase>compile</phase>
+                    </execution>
+                </executions>
+                <configuration>
+                    <pom>
+                        <sortPom>
```
aurambaj, line 129: I don't know about sortPom plugin, I thought it was nice but at the same time there were some things i wasn't a big fan of. For example it doesn't respect max line length in the prolog which a bit weird

pom.xml
```diff
@@ -109,6 +111,52 @@
...
+                            <nrOfIndentSpace>4</nrOfIndentSpace>
+                        </sortPom>
+                    </pom>
+                    <java>
+                        <googleJavaFormat>
```
aurambaj, line 134: I liked those option for l10nmessages:

```
<toggleOffOn />
<removeUnusedImports />
```

pom.xml
```diff
@@ -109,6 +111,52 @@
...
+                            <version>${plugin.version.google-java-format}</version>
+                            <style>GOOGLE</style>
```
aurambaj, line 136: isn't that style the default value? if so remove?

pom.xml
aurambaj, line 140: isn't the order automatic?

pom.xml
```diff
@@ -109,6 +111,52 @@
...
+                        </googleJavaFormat>
+                        <importOrder>
+                            <wildcardsLast>false</wildcardsLast>
```
aurambaj, line 139: if we have auto formatting so we avoid wildcards? those in some cases it is nice to have for static imports.

pom.xml
garionpin: Will add them.

pom.xml
garionpin: Yes - it's to avoid wildcard imports. I'm generally not too bothered about wildcard imports - but I wanted to align it with what we tend to do in the project currently.

I haven't seen any setting that can be used to differentiate between the context of where that's used, either on/off everywhere. It's easiest to stick to the style guide in this instance (no wildcard imports).

According to the Google Java style guide, wildcard imports are not recommended for static imports either:
https://google.github.io/styleguide/javaguide.html
`3.3.1 No wildcard imports
Wildcard imports, static or otherwise, are not used.`

pom.xml
garionpin: The default one is quite different than what we use today - if you try it locally you'll see what I mean.

If we're happy with the default style the plugin offers - I'm Ok to rely on that.

pom.xml
garionpin: I can remove it from this PR - and we can re-consider adding/evaluating it as a separate change later on - there's plenty of changes without it already.

pom.xml
garionpin: Yes - added it for clarity, will remove.

pom.xml
garionpin: Nope - I mis-read what the setting does - it's just about ordering - will remove it.

pom.xml
aurambaj: the auto-formatter might be better than what we have today.  Actually I'm a bit surprise with our current order. In my mind, I have this reference

```
4.2 The import statements must follow the package statement. import statements should be sorted with the most fundamental packages first, and grouped with associated packages together and one blank line between groups.
```

this doesn't seem to be followed today, and maybe the formatter fixed it to follow those rules?

pom.xml
aurambaj: from what I've read in the past it is better to avoid them. One case i though was handy though with `l10nmessages` was to be able to wildcard the message ids, so you don't have to bother to get something to compile as you're editing the properties file. that's a very specific use case though

pom.xml
garionpin: I agree with the intent - however what the plugin does without that - is that it separates static imports - and everything else seems to be sorted alphabetically into one block without any spaces in between, e.g. for BranchDeleteCommand this is the result:

`package com.box.l10n.mojito.cli.command;

import static com.box.l10n.mojito.cli.command.param.Param.BRANCH_CREATED_BEFORE_LAST_WEEK_DESCRIPTION;
import static com.box.l10n.mojito.cli.command.param.Param.BRANCH_CREATED_BEFORE_LAST_WEEK_LONG;
import static com.box.l10n.mojito.cli.command.param.Param.BRANCH_CREATED_BEFORE_LAST_WEEK_SHORT;
import static com.box.l10n.mojito.cli.command.param.Param.BRANCH_NULL_BRANCH_DESCRIPTION;
import static com.box.l10n.mojito.cli.command.param.Param.BRANCH_NULL_BRANCH_LONG;
import static com.box.l10n.mojito.cli.command.param.Param.BRANCH_NULL_BRANCH_SHORT;
import static com.box.l10n.mojito.cli.command.param.Param.BRANCH_TRANSLATED_DESCRIPTION;
import static com.box.l10n.mojito.cli.command.param.Param.BRANCH_TRANSLATED_LONG;
import static com.box.l10n.mojito.cli.command.param.Param.BRANCH_TRANSLATED_SHORT;

import com.beust.jcommander.Parameter;
import com.beust.jcommander.Parameters;
import com.box.l10n.mojito.cli.command.param.Param;
import com.box.l10n.mojito.cli.console.ConsoleWriter;
import com.box.l10n.mojito.rest.client.AssetClient;
import com.box.l10n.mojito.rest.client.RepositoryClient;
import com.box.l10n.mojito.rest.entity.Branch;
import com.box.l10n.mojito.rest.entity.PollableTask;
import com.box.l10n.mojito.rest.entity.Repository;
import java.util.List;
import org.fusesource.jansi.Ansi;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.context.annotation.Scope;
`

pom.xml
garionpin: Just to confirm: that setting didn't control the actual wildcard part - only where wildcards are placed within the list.
We'll still have:
`import static org.junit.Assert.*;`

pom.xml
garionpin: Actually, looking at the style guide that seems to match the intent for the most part:

> 3.3.3 Ordering and spacing
> Imports are ordered as follows:
> 
> All static imports in a single block.
> All non-static imports in a single block.
> If there are both static and non-static imports, a single blank line separates the two blocks. There are no other blank lines between import statements.
> 
> Within each block the imported names appear in ASCII sort order. (Note: this is not the same as the import statements being in ASCII sort order, since '.' sorts before ';'.)

pom.xml
```diff
@@ -109,6 +111,48 @@
...
+                        </excludes>
+                        <googleJavaFormat>
+                            <version>${plugin.version.google-java-format}</version>
+                        </googleJavaFormat>
+                        <importOrder>
+                            <order>\\#,java,javax,org,com,,com.box</order>
+                        </importOrder>
+                        <indent>
+                            <tabs>true</tabs>
+                            <spacesPerTab>2</spacesPerTab>
```
aurambaj, line 140: duplicated? remove?

PR #859, opened 2022-08-31
Add pagination data to the /api/deltas/date API.

garionpin:
Add pagination data and refactor response type for /api/deltas/date API
endpoint.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/tm/TMTextUnitVariantRepository.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/tm/TMTextUnitVariantRepository.java
@@ -42,11 +42,12 @@ public interface TMTextUnitVariantRepository extends JpaRepository<TMTextUnitVar
    */
   @Query(
       value =
-          "select distinct tuv.id, "
-              + "   tu.name as tmTextUnitName, "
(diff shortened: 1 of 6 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/tm/TMTextUnitVariantRepository.java
```diff
@@ -42,11 +42,12 @@ List<TMTextUnitVariant> findByTmTextUnitTmRepositoriesAndLocale_Bcp47TagNotOrder
    */
   @Query(
       value =
-          "select distinct tuv.id, "
-              + "   tu.name as tmTextUnitName, "
-              + "   tuv.locale.bcp47Tag as bcp47Tag,"
-              + "   tuv.content as content, "
-              + "   'UNKNOWN' as deltaType "
+          "select distinct new com.box.l10n.mojito.service.tm.TextUnitVariantDeltaDTO("
```
aurambaj, line 45: it was working without an actual DTO instance before? just wondering why the change?

garionpin: Needed a DTO for the return type of the api/delta/date API anyways and didn't think an extra conversion was warranted.

PR #864, opened 2022-09-08
Truncate the Branch TextUnit View in Mojito Workbench with large text unit counts

baktayPin:
Truncate the Branch TextUnit View in Mojito Workbench with large text unit counts
- Retrieve the branches with greater than 30,000 text units without retrieving the TextUnit table
- Show the branches with the high count in the workbench, with the message 'Results cannot be shown as the branch contains high number of Text Units (more than 30,000).'

Screenshot 2022-09-13 at 16 54 13

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticWS.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticWS.java
@@ -8,7 +8,6 @@ import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.c
 import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.deletedEquals;
 import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.empty;
 import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.search;
-import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.totalCountLessThanOrEqualsTo;
 import static com.box.l10n.mojito.specification.Specifications.ifParamNotNull;
 import static org.slf4j.LoggerFactory.getLogger;
 import static org.springframework.data.jpa.domain.Specification.where;
@@ -22,6 +21,7 @@ import com.box.l10n.mojito.service.branch.BranchStatisticRepository;
 import com.box.l10n.mojito.service.branch.SparseBranchStatisticRepository;
 import com.fasterxml.jackson.annotation.JsonView;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -64,6 +64,7 @@ public class BranchStatisticWS {
       @RequestParam(value = "createdAfter", required = false) DateTime createdAfter,
       @PageableDefault(sort = "id", direction = Sort.Direction.DESC) Pageable pageable) {
     // Two phase querying: 1. retrieve BranchStatistic IDs for pagination first
+    // Retrieve all the Branch Statistic Ids without the totalCountLte filter:
     Page<Long> branchStatisticIds =
         sparseBranchStatisticRepository.findAllWithIdOnly(
             where(ifParamNotNull(createdByUserNameEquals(createdByUserName)))
@@ -73,15 +74,38 @@ public class BranchStatisticWS {
                 .and(ifParamNotNull(deletedEquals(deleted)))
                 .and(ifParamNotNull(empty(empty)))
                 .and(ifParamNotNull(createdBefore(createdBefore)))
-                .and(ifParamNotNull(createdAfter(createdAfter)))
-                .and(totalCountLessThanOrEqualsTo(totalCountLte)),
+                .and(ifParamNotNull(createdAfter(createdAfter))),
             pageable);
+
+    // Retrieve the branch statistics where the total text unit count is greater than the limit,
+    // by filtering the branchStatisticIds less than the limit.
+    List<BranchStatistic> branchStatisticsWithoutTextUnits =
+        branchStatisticRepository.findAllGreaterThanById(
+            branchStatisticIds.getContent(), totalCountLte);
+
+    // Set the BranchTextUnitStatistics to null as the branch has a large number of text Units.
+    branchStatisticsWithoutTextUnits.forEach(
+        branchStatistic -> {
+          branchStatistic.setBranchTextUnitStatistics(null);
+        });
+
+    List<Long> branchStatisticIdsWithoutTextUnits =
+        branchStatisticsWithoutTextUnits.stream()
+            .map(BranchStatistic::getId)
+            .collect(Collectors.toList());
+
+    List<Long> branchStatisticIdsWithLessThan =
+        branchStatisticIds.getContent().stream()
+            .filter(i -> !branchStatisticIdsWithoutTextUnits.contains(i))
+            .collect(Collectors.toList());
 
     // 2. Hydrate BranchStatistic entities for JSON response using eager loading to avoid N+1
     // queries
     List<BranchStatistic> branchStatistics =
-        branchStatisticRepository.findByIdIn(branchStatisticIds.getContent(), pageable.getSort());
+        branchStatisticRepository.findByIdIn(branchStatisticIdsWithLessThan, pageable.getSort());
 
+    // Merge the two lists:
+    branchStatistics.addAll(branchStatisticsWithoutTextUnits);
     PageImpl<BranchStatistic> page =
         new PageImpl<>(
             branchStatistics,
--- a/webapp/src/main/java/com/box/l10n/mojito/entity/BranchStatistic.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/entity/BranchStatistic.java
@@ -27,6 +27,19 @@ import javax.persistence.*;
               @NamedAttributeNode(value = "screenshots"),
               @NamedAttributeNode(value = "repository"),
             }))
+@NamedEntityGraph(
+    name = "BranchStatisticGraphWithoutTextUnits",
+    attributeNodes = {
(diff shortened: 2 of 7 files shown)


webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticWS.java
```diff
@@ -52,41 +44,70 @@ public class BranchStatisticWS {
...
+            branchStatisticIdsWithLessThan.getContent(), pageable.getSort());
+
+    // Merge the two lists:
+    branchStatistics.addAll(branchStatisticsWithoutTextUnits.getContent());
     PageImpl<BranchStatistic> page =
-        new PageImpl<>(
-            branchStatistics,
-            branchStatisticIds.getPageable(),
-            branchStatisticIds.getTotalElements());
+            new PageImpl<>(
```
garionpin, line 107: Looks like the formatting tool wasn't run locally and IntelliJ wasn't configured with the Google Java styleguide - could you run a "mvn compile" locally to fix?

webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticWS.java
```diff
@@ -52,41 +44,70 @@ public class BranchStatisticWS {
...
+                    ifParamNotNull(createdByUserNameEquals(createdByUserName))).and(
+                    ifParamNotNull(branchEquals(branchId))).and(
+                    ifParamNotNull(branchNameEquals(branchName))).and(
+                    ifParamNotNull(search(search))).and(
+                    ifParamNotNull(deletedEquals(deleted))).and(
+                    ifParamNotNull(empty(empty))).and(
+                    ifParamNotNull(createdBefore(createdBefore))).and(
+                    ifParamNotNull(createdAfter(createdAfter))).and(
+                    totalCountLessThanOrEqualsTo(totalCountLte))
+            , pageable);
```
garionpin, line 81: Using pageable here will result in results getting out of sync after a couple of pages, once a couple of larger branches are thrown in the mix. Passing in pageable to a query can only be done once for that reasons.

This query should take in branchStatisticIds as a query condition and filter for count based on those results and pageable should not be used here.

webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticWS.java
```diff
@@ -52,41 +44,70 @@ public class BranchStatisticWS {
...
+    // by filtering the branchStatisticIds less than the limit.
+    Page<BranchStatistic> branchStatisticsWithoutTextUnits = branchStatisticRepository.findAll(where(
+                    ifParamNotNull(createdByUserNameEquals(createdByUserName))).and(
+                    ifParamNotNull(branchEquals(branchId))).and(
+                    ifParamNotNull(branchNameEquals(branchName))).and(
+                    ifParamNotNull(search(search))).and(
+                    ifParamNotNull(deletedEquals(deleted))).and(
+                    ifParamNotNull(empty(empty))).and(
+                    totalCountGreaterThan(totalCountLte))
+            , pageable);
```
garionpin, line 93: Same comment from above applies here - it's not safe to use pageable again - this query should use as input the IDs from the very first query and the count desired.

However, we shouldn't go to the DB for this set of entries - we can take branchStatisticIds, eliminate IDs  from branchStatisticIdsWithLessThan - and the resulting IDs are exactly what we're looking for.

webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticWS.java
```diff
@@ -52,41 +44,70 @@ public class BranchStatisticWS {
...
+
+    // Set the BranchTextUnitStatistics to null as the branch has a large number of text Units.
+    branchStatisticsWithoutTextUnits.forEach(
+            branchStatistic -> {branchStatistic.setBranchTextUnitStatistics( null );});
+
+
+    // 2. Hydrate BranchStatistic entities for JSON response using eager loading to avoid N+1 queries
+    List<BranchStatistic> branchStatistics = branchStatisticRepository.findByIdIn(
```
garionpin, line 101: Won't this cause the branchStatisticsWithoutTextUnits entities to get their properties using N+1 queries? 
They still have data that needs to be serialized and sent down in JSON form (outside of the TextUnit information that we don't want for them) - except now they won't be retrieved in one go.

garionpin (changes requested):

webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticWS.java
```diff
@@ -1,14 +1,6 @@
...
-import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.branchEquals;
-import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.branchNameEquals;
-import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.createdAfter;
-import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.createdBefore;
-import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.createdByUserNameEquals;
-import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.deletedEquals;
-import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.empty;
-import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.search;
-import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.totalCountLessThanOrEqualsTo;
+import static com.box.l10n.mojito.rest.repository.BranchStatisticSpecification.*;
```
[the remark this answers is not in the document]
baktayPin: Yes, indeed looks like my IntelliJ automatically did this change

webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticWS.java
baktayPin: Yes, I tried that unfortunately the branchStatisticIds doesn't give me the all the list of items as it's paged, instead it would give me only the first page of Ids.

webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticWS.java
baktayPin: This line is using the branchStatisticIdsWithLessThan, so not sure why the branchStatisticsWithoutTextUnits would be retrieved here?

webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticWS.java
garionpin: How many branches do you expect to return from this API request? More than what one page (the current one) contains?

baktayPin: Yes in my test DB I had 16 branches returned, in production if user doesn't select the 'only my branches' it could even return 57 branches (6 pages).

baktayPin: When I was testing I was expecting 16 (>30K branches) + 8 (<=30K) branches, so 3 pages was the number I was expecting from the method.

garionpin: If the size of the page passed in is 10 - only 10 results should be returned though - there's no value then to ever query data for more than 10 branches at a time.

webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticWS.java
garionpin: The problem is that the branchStatisticsWithoutTextUnits entities still have linked entities - if they're not retrieved explicitly through an optimized entity graph - how many queries will they generate to get their related Repository, User, Authority etc. entities?  

Tracking the SQL queries generated should confirm how many queries would result as a result of the response being serialized.

garionpin (approved):

garionpin (approved):

webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticWS.java
```diff
@@ -73,15 +74,38 @@ public Page<BranchStatistic> getBranchesOfRepository(
...
+            .collect(Collectors.toList());
+
     // 2. Hydrate BranchStatistic entities for JSON response using eager loading to avoid N+1
     // queries
     List<BranchStatistic> branchStatistics =
-        branchStatisticRepository.findByIdIn(branchStatisticIds.getContent(), pageable.getSort());
+        branchStatisticRepository.findByIdIn(branchStatisticIdsWithLessThan, pageable.getSort());
 
+    // Merge the two lists:
+    branchStatistics.addAll(branchStatisticsWithoutTextUnits);
```
aurambaj, line 108: it seems that logic is changing the order of the results by putting them at the end of the page. Would that be an issue. I guess it is inside a page so it is not a big difference.

aurambaj (approved):

webapp/src/main/java/com/box/l10n/mojito/rest/repository/BranchStatisticWS.java
baktayPin: Yes, it will be adding to the end of the page's end. I suppose we can order them as well, but not sure if it's necessary. Let me know if you want me to update it.

aurambaj: yeah not sure it is super important, my initial thought is that we should keep the same order as what `Page<Long> branchStatisticIds =
        sparseBranchStatisticRepository.findAllWithIdOnly(` returns but probably not a big deal. Let's keep it like this

aurambaj (approved):

PR #873, opened 2022-10-12
Support list of file types in push/pull/extract/... commands and misc. changes

aurambaj:


cli/src/main/java/com/box/l10n/mojito/cli/command/param/Param.java
```diff
@@ -94,12 +94,12 @@ public class Param {
...
   public static final String FILE_TYPE_SHORT = "-ft";
   public static final String FILE_TYPE_DESCRIPTION =
-      "File type (if none provided it will scan for default formats: XLIFF, XCODE_XLIFF, MAC_STRING, MAC_STRINGSDICT, ANDROID_STRINGS, PROPERTIES, PROPERTIES_NOBASENAME, PROPERTIES_JAVA, RESW, RESX, PO, XTB)";
+      "File type(s), can be a list. If none provided it will scan for default formats: XLIFF, XCODE_XLIFF, MAC_STRING, MAC_STRINGSDICT, ANDROID_STRINGS, PROPERTIES, PROPERTIES_NOBASENAME, PROPERTIES_JAVA, RESW, RESX, PO, XTB";
 
   public static final String FILTER_OPTIONS_LONG = "--filter-options";
   public static final String FILTER_OPTIONS_SHORT = "-fo";
   public static final String FILTER_OPTIONS_DESCRIPTION =
-      "Filter options. Format: option1=value1&option2=value2";
+      "Filter options, can be a list. Format: option1=value1";
```
garionpin, line 102: What is the syntax for multiple values? Is it space or comma separated? (An example would still be beneficial there)

garionpin (approved):

cli/src/main/java/com/box/l10n/mojito/cli/command/param/Param.java
aurambaj: that should work: `-fo "option1=value1" "option2=value2"`

cli/src/main/java/com/box/l10n/mojito/cli/filefinder/FileFinder.java
```diff
@@ -193,7 +217,12 @@ private List<FileMatch> getSourceMatcheCandidates(FileType fileType) throws File
     }
 
     SourceFileVisitor sourceFileVisitor =
-        new SourceFileVisitor(fileType, sourceDirectory, sourcePathFilterRegex);
+        new SourceFileVisitor(
+            fileType,
```
garionpin, line 221: Do we walk the tree for each and every different fileType? vs.: instead of doing a single walk of the tree - and checking every file against one or more file type extensions.

cli/src/main/java/com/box/l10n/mojito/cli/filefinder/DirectoryScanUtils.java
```diff
@@ -0,0 +1,124 @@
...
+   * shouldScanDirectory([ "webapp", "locale" ], [ "webapp" ]) --> true, we need to recurse to find
+   * directories that potentially match shouldScanDirectory([ "*", "locale" ], [ "webapp" ]) -->
+   * true shouldScanDirectory([ "*", "locale" ], [ "webapp", "locale" ]) --> true
+   *
+   * <p>Edge cases: - if the directory is an empty array, which is equivalent to ask should we scan
+   * the root, then this will return true since we need to scan the root regardless of the pattern
+   * to get started - if an empty array is provided as pattern it is equivalent to matching the root
+   * and it will always be true regardless of the directory
+   */
+  static boolean shouldScanDirectory(String[] directoryPattern, String[] aDirectory) {
```
garionpin, line 96: Should the pattern and directory be passed as one object that has a Pattern and Directory value in a list of similar objects?

I was confused initially that any of the patterns could apply to any of the directories passed in.

garionpin: Also - this method name should be pluralized.

cli/src/main/java/com/box/l10n/mojito/cli/command/CommandHelper.java
```diff
@@ -86,68 +87,94 @@ public List<Repository> getAllRepositories() {
    * Get list of {@link com.box.l10n.mojito.cli.filefinder.FileMatch} from source directory
    *
    * @param commandDirectories
-   * @param fileType
+   * @param fileTypes
```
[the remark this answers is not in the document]
aurambaj: i'm planning to keep the commits unsquashed, each commit can be reviewed as 1 unit

cli/src/main/java/com/box/l10n/mojito/cli/filefinder/FileFinder.java
aurambaj: We walk the tree once (in fact we walk it twice: one for the source file and one for target files) and check all filetypes registered or provided as command options. Trying to remove the target scan could make sense since the limiting factor is IO. That said checking for all file types become expansive too and it doesn't seem to be a really valid use case after all. We use a file type in all of our usages. It feels like it could/should be made a mandatory option

cli/src/main/java/com/box/l10n/mojito/cli/filefinder/DirectoryScanUtils.java
aurambaj: Not 100% sure to get your point, more context... This is working on a single directory with a single pattern. Args are string arrays because the pattern and path have been split already. I didn't want to keep splitting the pattern every single time (see splitPatterns() called in the ctor). Does it make sense or you mean something different?

PR #883, opened 2022-11-03
Not able to build on Mac M1 because Cannot download node - 404 | Slack notification not working because using a depreciated api endpoint

kdcyberdude:
Why the project is not building on Mac M1 - https://github.com/eirslett/frontend-maven-plugin/issues/952#issuecomment-762163294
Why slack notification is not working - https://github.com/howdyai/botkit/issues/1988

### DIFF
--- a/common/src/main/java/com/box/l10n/mojito/slack/SlackClient.java
+++ b/common/src/main/java/com/box/l10n/mojito/slack/SlackClient.java
@@ -25,7 +25,7 @@ public class SlackClient {
 
   static final String BASE_API_URL = "https://slack.com/api/";
   static final String API_USER_LOOKUP_BY_EMAIL = "users.lookupByEmail";
-  static final String API_IM_OPEN = "im.open";
+  static final String API_IM_OPEN = "conversation.open";
   static final String API_CHAT_POST_MESSAGE = "chat.postMessage";
 
   public static final String COLOR_GOOD = "good";
--- a/common/src/main/java/com/box/l10n/mojito/slack/SlackClient.java
+++ b/common/src/main/java/com/box/l10n/mojito/slack/SlackClient.java
@@ -25,7 +25,7 @@ public class SlackClient {
 
   static final String BASE_API_URL = "https://slack.com/api/";
   static final String API_USER_LOOKUP_BY_EMAIL = "users.lookupByEmail";
-  static final String API_IM_OPEN = "conversation.open";
+  static final String API_CONVERSATION_OPEN = "conversations.open";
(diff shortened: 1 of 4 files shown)


aurambaj:
@kdcyberdude can you accept the license/cla, so i can merge your changes?

common/src/main/java/com/box/l10n/mojito/slack/SlackClient.java
```diff
@@ -25,7 +25,7 @@ public class SlackClient {
 
   static final String BASE_API_URL = "https://slack.com/api/";
   static final String API_USER_LOOKUP_BY_EMAIL = "users.lookupByEmail";
-  static final String API_IM_OPEN = "im.open";
+  static final String API_IM_OPEN = "conversation.open";
```
aurambaj, line 28: would you mind renaming the constant? 

Does the new entry point take the exact same input, I check the doc quickly and it seems to now use "users" field in the payload and not "user"? Have you tried it with that exact code?

kdcyberdude: Hi @aurambaj, I've already signed a CLA agreement months ago but it's still pending with the following reason.

> kdcyberdude seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).
You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/box/mojito?pullRequest=883) it.

I suggest you to push the changes yourself. As you mentioned it is needs "conversations" in the API and "users" in the parameters. I had made the changes month ago and tested it already. 

One more thing, can you update the docs mentioning the slack notification trigger points. Still I am not able to fully understand when the notification get's triggered if multiple developers are pushing the changes in their respective branch on mojito.

aurambaj: 🤔 that's unfortunate, not sure what's happening. I went ahead and commit: https://github.com/box/mojito/commit/4d205e45405fb6deb2d87508c97f3d38b0b83b9c for the slack client fix. 

Thanks for following up!

The way we use it is that each developer as it own feature branch. they are independent branches (with their own new/deleted strings) so each developer have their own notifications as described in that doc: https://www.mojito.global//docs/guides/branching/. The string set in Mojito is the union of the string on all branches. Do you use it the same way? What's unclear? are you missing notifications?

aurambaj:
could not merge at the time CLA, but the content of this PR was eventually added

PR #888, opened 2022-11-21
Send timeout value with third party sync command

maallen:
Allows a user to set a custom job timeout value with the third party sync command

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/pollableTask/Pollable.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/pollableTask/Pollable.java
@@ -60,4 +60,13 @@ public @interface Pollable {
    * @return the delay in seconds
    */
   long timeout() default 3600;
+
+  /**
+   * Indicates if the timeout value set in the annotation should take precedence over the parent's
+   * timeout value.
+   *
+   * @return {@code true} if a parent tasks timeout value should be overridden in the child else
+   *     {@code false} (default)
+   */
+  boolean overrideParentTimeout() default false;
 }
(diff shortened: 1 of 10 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSInMemory.java
```diff
@@ -88,7 +88,13 @@ public void pull(
       Map<String, String> localeMapping,
       String skipTextUnitsWithPattern,
       String skipAssetsWithPathPattern,
-      List<String> optionList) {}
+      List<String> optionList) {
+    try {
+      Thread.sleep(60000);
+    } catch (InterruptedException e) {
+      throw new RuntimeException(e);
```
garionpin, line 95: Is this intended or is it test code that remained in the PR?

garionpin (approved):

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSInMemory.java
maallen: That was test code used to delay my local third party call, will remove now

webapp/src/main/java/com/box/l10n/mojito/service/pollableTask/Pollable.java
```diff
@@ -60,4 +60,13 @@
...
   long timeout() default 3600;
+
+  /**
+   * Indicates if the timeout value set in the annotation should take precedence over the parent's
+   * timeout value.
+   *
+   * @return {@code true} if a parent tasks timeout value should be overridden in the child else
+   *     {@code false} (default)
+   */
+  boolean overrideParentTimeout() default false;
```
aurambaj, line 71: Can you update the `timeout()` doc to reflect this addition as well something like 

> If the pollable task has a parent task, it will use the parent's timeout unless {@link #overrideParentTimeout()} is to true

aurambaj (approved):

aurambaj (approved):

PR #891, opened 2022-12-02
Add web service and CLI to machine translate untranslated strings in …

aurambaj:
…a repository

a quick but working implementation: one locale at the time, get untranslated strings, send them to the machine translation engine and then import the translations in the repository


webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/RepositoryMachineTranslationService.java
```diff
@@ -0,0 +1,127 @@
...
+                    } else {
+                      logger.error("There must be a translation available but if not, just skip");
+                      textUnitDTOStreamFor1Translation = Stream.empty();
+                    }
+
+                    return textUnitDTOStreamFor1Translation;
+                  })
+              .collect(Collectors.toList());
+
+      textUnitBatchImporterService.importTextUnits(machineTranslatedTextUnitDTOs, false, true);
```
garionpin, line 122: Shouldn't these translations be marked (imported) with a different status - so that someone looking at the repo can easily tell the difference between MT and human translated strings?

cli/src/main/java/com/box/l10n/mojito/cli/command/RepositoryMachineTranslationCommand.java
```diff
@@ -0,0 +1,84 @@
...
+ * Command to machine translate untranslated strings in a repository.
+ *
+ * @author jaurambault
+ */
+@Component
+@Scope("prototype")
+@Parameters(
+    commandNames = {"repository-machine-translate"},
+    commandDescription = "Machine translate untranslated strings in a repository")
+public class RepositoryMachineTranslationCommand extends Command {
```
garionpin, line 29: Rather than needing to run this command to translate the repo every time - would it not be better to have it implemented as part of the pull command (as an additional option) - so that every time someone wants to see a fully translated app - they can pass in the extra parameter? 
In that case - importing the MT translations into the repo and saving them explicitly against the text units can be avoided.

webapp/src/main/java/com/box/l10n/mojito/service/machinetranslation/RepositoryMachineTranslationService.java
garionpin: Also - if we leverage across repos (I don't remember what happens for MT leveraging when repo names aren't provided - that might be the case) - MT translations could be returned back as actual translations - they poison our TMs essentially.

PR #898, opened 2022-12-12
Fix remove screenshot NullPointer test

montedonio-pinterest:


webapp/src/test/java/com/box/l10n/mojito/service/screenshot/ScreenshotServiceTest.java
```diff
@@ -105,7 +106,9 @@ public void testDeleteScreenshot() throws Exception {
     ScreenshotTextUnit createdScreenshotTextUnit =
         screenshotTextUnitRepository.findById(screenshotTextUnit.getId()).orElse(null);
     assertEquals(createdScreenshotTextUnit.getScreenshot().getId(), screenshot.getId());
+    screenshot.setScreenshotTextUnits(new HashSet(Arrays.asList(createdScreenshotTextUnit)));
```
aurambaj, line 109: can use `Sets.newHashSet()` from Guava

aurambaj (commented):
what was the issue? why is it transient?

PR #902, opened 2023-01-06
Allow configuration of multiple notifiers (slack, phabricator, github) per branch

aurambaj:


webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/BranchNotificationMessageSendersConfigurationProperties.java
```diff
@@ -277,4 +287,19 @@ public void setNoMoreStrings(String noMoreStrings) {
       }
     }
   }
+
+  static class GithubConfigurationProperties {
+
+    // TODO(jean) it migth make more to look up by "id" vs "owner" to have all implementation
+    // with the same logic, instead of having logical keys that are implementation specific
+    String owner;
```
aurambaj, line 295: @maallen thoughts?

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/BranchNotificationMessageSendersConfigurationProperties.java
```diff
@@ -31,4 +41,104 @@ public void setEnabled(boolean enabled) {
...
+      return useDirectMessage;
+    }
+
+    public void setUseDirectMessage(boolean useDirectMessage) {
+      this.useDirectMessage = useDirectMessage;
+    }
+
+    public static class MessageBuilderConfigurationProperties {
+
+      String newStrings =
```
aurambaj, line 108: @maallen I tried to refactor names so that they are consistent. We'll need to be careful when updating the configuration to use the new names

webapp/src/main/java/com/box/l10n/mojito/entity/Branch.java
```diff
@@ -63,6 +63,13 @@ public class Branch extends SettableAuditableEntity {
   @OneToOne(mappedBy = "branch", fetch = FetchType.LAZY)
   BranchStatistic branchStatistic;
 
+  @ElementCollection(fetch = FetchType.EAGER)
```
garionpin, line 66: This implies that each and every time we want to work with a Branch entity - we will always want to retrieve its notifiers - even if the task at hand isn't to notify anything (e.g.: listing branches per repo, asset extraction etc.).

A HQL @Query might be better in this case to limit the retrieval of this data to solely the scenario that is needed in.

webapp/src/main/resources/db/migration/V62__branch_notifiers.sql
```diff
@@ -0,0 +1,3 @@
+alter table branch_notification add column notifier_id varchar(255);
+create table branch_notifiers (branch_id bigint not null, notifiers varchar(255)) engine=InnoDB;
```
garionpin, line 2: Was the "engine=InnoDB" part left in by accident?
We don't specify the underlying storage engine anywhere else - outside of some Quartz tables and spring session table. (Do you have any information if they're required there?)

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionDiffCommand.java
```diff
@@ -147,6 +149,13 @@ public class ExtractionDiffCommand extends Command {
       description = "Optional username who owns the branch when pusing to a repository")
   String pushToBranchCreatedBy;
 
+  @Parameter(
+      names = {"--push-to-branch-notifiers", "-pbn"},
+      variableArity = true,
+      required = false,
+      description = "Optional list of notifiers when pusing to a repository")
```
garionpin, line 156: Could we add a bit more information for CLI users as to what these notifiers are for and when they would get called/invoked in the localization process?

common/src/main/java/com/box/l10n/mojito/slack/SlackClients.java
```diff
@@ -0,0 +1,40 @@
+package com.box.l10n.mojito.slack;
+
+import java.util.AbstractMap.SimpleEntry;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.springframework.stereotype.Component;
+
+/**
+ * Create Slack clients from configuration and expose them by client ids. This allows to configure
+ * and communicate with different Slack instances.
```
garionpin, line 10: Could you remind me please for which scenario we now require the ability to communicate with different Slack instances (vs. just assuming one instance) ?

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/BranchNotificationMessageSenders.java
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  private Map<String, BranchNotificationMessageSender> createInstancesFromConfiguration(
+      BranchNotificationMessageSendersConfigurationProperties
+          branchNotificationMessageSendersConfigurationProperties,
+      GithubClients githubClients,
+      SlackClients slackClients,
+      BranchUrlBuilder branchUrlBuilder) {
+
+    checkIdsStartWithPrefix(
```
garionpin, line 59: Why is it important to validate the name of the keys?
Do we functionally rely on the naming pattern/prefix latter on to drive some logic/instantiate the right class? (that wouldn't be ideal if that's the case)

Or is this purely so as to enforce consistent naming in the config files?

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/BranchNotificationMessageSendersConfigurationProperties.java
garionpin: (As the IDs in other cases are strings - does it matter what we put in the IDs in general for the values as long as they're unique?)

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/BranchNotificationService.java
```diff
@@ -0,0 +1,384 @@
...
+  }
+
+  /**
+   * Schedule to check/send notification for missing screenshot in 30 minutes from now.
+   *
+   * @param branch
+   * @param notifierId
+   */
+  void scheduleMissingScreenshotNotificationsForBranch(Branch branch, String notifierId) {
+    Date date = DateTime.now().plusMinutes(30).toDate();
```
garionpin, line 171: Should the time (30m) be configurable?

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/job/BranchNotificationMissingScreenshotsJob.java
```diff
@@ -21,14 +22,27 @@
...
-        "execute for branchId: {} and sender type: {}", input.getBranchId(), input.getSenderType());
-    branchNotificationService.sendMissingScreenshotNotificationForBranch(
-        input.getBranchId(), input.getSenderType());
+    if (input.getSenderType() == null) {
+      logger.debug(
+          "execute for branchId: {} and notifier id: {}",
+          input.getBranchId(),
+          input.getNotifierId());
+      branchNotificationService.sendMissingScreenshotNotificationForBranch(
+          input.getBranchId(), input.getSenderType());
```
aurambaj, line 37: typo!! fix

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/BranchNotificationMessageSendersConfigurationProperties.java
maallen: I think that should be fine, presumably in that case we'll need to store the notifier id in the database for each relevant branch (and specify the notifier id rather than owner type when creating the branch itself). I don't see an issue here as long as we're consistent in the approach between what's used in the cli command and the server backend

webapp/src/main/java/com/box/l10n/mojito/entity/Branch.java
aurambaj: yes that was kind of my goal, in fact I didn't really want to have a table for that but I didn't want to hack the list into a single column. Do you think of a use case where this will cause performance issue?

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/BranchNotificationService.java
aurambaj: could be, never discuss changing it though so not urgent

cli/src/main/java/com/box/l10n/mojito/cli/command/ExtractionDiffCommand.java
aurambaj: adding some basic info, but this would require a full independent documentation ideally

PR #905, opened 2023-01-12
Add a GithubClient method to add a comment to a commit

aurambaj:


maallen (approved):

common/src/main/java/com/box/l10n/mojito/github/GithubClient.java
```diff
@@ -86,6 +86,22 @@ public void addCommentToPR(String repository, int prNumber, String comment) {
     }
   }
 
+  public void addCommentToCommit(String repository, String commit, String comment) {
```
maallen, line 89: Nitpick but maybe update the `commit` variable name to `commitSha` for clarity?

aurambaj: good one thanks!

PR #908, opened 2023-01-12
add YAML filter and fileType support

montedonio-pinterest:


common/src/main/java/com/box/l10n/mojito/okapi/filters/YAMLFilter.java
```diff
@@ -0,0 +1,30 @@
...
+  public String getName() {
+    return FILTER_CONFIG_ID;
+  }
+
+  @Override
+  public List<FilterConfiguration> getConfigurations() {
+    List<FilterConfiguration> list = new ArrayList();
+    list.add(
+        new FilterConfiguration(
+            getName() + "-yaml",
```
aurambaj, line 23: this should be getName() only I think

cli/src/main/java/com/box/l10n/mojito/cli/filefinder/file/YAMLFileType.java
```diff
@@ -0,0 +1,10 @@
+package com.box.l10n.mojito.cli.filefinder.file;
+
+/** @author loliveiramontedonio */
+public class YAMLFileType
```
aurambaj, line 4: to be consistent with the filter name, can you adjust the case "YAML" --> "Yaml"

webapp/package-lock.json
```diff
@@ -6047,9 +6047,9 @@
       },
       "dependencies": {
         "node-fetch": {
-          "version": "2.6.7",
```
aurambaj, line 6050: can you remove updates to this file since it is unrelated?

aurambaj (approved):

PR #917, opened 2023-02-03
Added GithubGetInstallationTokenCommand

maallen:
Adds a command to retrieve a Github installation token for a configured github client

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/GithubGetInstallationTokenCommand.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/GithubGetInstallationTokenCommand.java
@@ -4,14 +4,13 @@ import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
 import com.box.l10n.mojito.cli.console.ConsoleWriter;
 import com.box.l10n.mojito.github.GithubClients;
+import java.io.IOException;
+import java.security.NoSuchAlgorithmException;
+import java.security.spec.InvalidKeySpecException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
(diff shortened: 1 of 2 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/GithubGetInstallationTokenCommand.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+@Scope("prototype")
+@Parameters(
+    commandNames = {"github-install-token"},
+    commandDescription = "Retrieves a github installation token for the specified installation")
+public class GithubGetInstallationTokenCommand extends Command {
+
+  @Autowired(required = false)
```
aurambaj, line 22: `required = false` is not necessary, there should always be an instance, potentially with no clients in it.

cli/src/main/java/com/box/l10n/mojito/cli/command/GithubGetInstallationTokenCommand.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Override
+  public boolean shouldShowInCommandList() {
+    return false;
+  }
+
+  @Override
+  protected void execute() throws CommandException {
+    try {
+      if (githubClients == null) {
```
aurambaj, line 51: having a doubt now, but see ExtractionDiffNotificationCommand. that test seems more relevant on `githubClients.getClient(owner)`

maallen: Turns out `githubClients.getClient(owner)` will throw an exception if there's no client configured for the supplied owner so I can remove the check here

aurambaj (approved):

PR #922, opened 2023-02-10
Allow customization of Github API endpoint

maallen:
Allows a custom Github endpoint to be set from configuration, if no configuration is set it defaults to `https://api.github.com`

### DIFF
--- a/common/src/main/java/com/box/l10n/mojito/github/GithubClient.java
+++ b/common/src/main/java/com/box/l10n/mojito/github/GithubClient.java
@@ -44,11 +44,13 @@ public class GithubClient {
 
   private PrivateKey signingKey;
 
+  private final String endpoint;
+
   protected GHAppInstallationToken githubAppInstallationToken;
 
   protected GitHub gitHubClient;
 
-  public GithubClient(String appId, String key, String owner, long tokenTTL) {
+  public GithubClient(String appId, String key, String owner, long tokenTTL, String endpoint) {
     this.appId = appId;
     this.key = key;
     if (owner == null || owner.isEmpty()) {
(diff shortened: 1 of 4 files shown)


garionpin (approved):

ehoogerbeets (commented):
Thanks for putting that in!

common/src/main/java/com/box/l10n/mojito/github/GithubClient.java
```diff
@@ -57,10 +59,12 @@ public GithubClient(String appId, String key, String owner, long tokenTTL) {
     }
     this.owner = owner;
     this.tokenTTL = tokenTTL;
+    this.endpoint =
+        endpoint.endsWith("/") ? endpoint.substring(0, endpoint.length() - 1) : endpoint;
   }
 
   public GithubClient(String appId, String key, String owner) {
-    this(appId, key, owner, 60000L);
+    this(appId, key, owner, 60000L, "https://api.github.com");
```
aurambaj, line 67: can we grab that from their constant "GITHUB_URL"?

common/src/main/java/com/box/l10n/mojito/github/GithubClientConfiguration.java
```diff
@@ -10,6 +10,8 @@ public class GithubClientConfiguration {
 
   Long tokenTTL = 60000L;
 
+  String endpoint = "https://api.github.com";
```
aurambaj, line 13: same?

common/src/main/java/com/box/l10n/mojito/github/GithubClient.java
maallen: No, it's not a public constant so can't be accessed outside of the `org.kohsuke.github` package

aurambaj (approved):

PR #923, opened 2023-02-10
Remove 'translations-required' label from Github PR 

maallen:
Adds logic to `GithubExtractionDiffNotificationCommand` to remove the `translations-required` label from a PR if it exists and there is no new strings introduced.


cli/src/main/java/com/box/l10n/mojito/cli/command/GithubExtractionDiffNotificationCommand.java
```diff
@@ -162,6 +167,22 @@ private String getBaseMessage(ImmutableMap<String, Object> arguments, String mes
     return messageFormat.format(arguments);
   }
 
+  private void removeTranslationsRequiredLabelIfPresent(GithubClient github) {
```
aurambaj, line 170: we should actually stop using that command, and use the generic one instead: `ExtractionDiffNotificationCommand`. Can you refactor accordingly? this one was meant to be removed, the Phabricator one we can keep for a little while more

maallen: Refactored, also added logic to add the `translations-required` label to a PR if it hasn't already been applied to avoid the initial lag of the branch being created on the server and then labelling occurring.

aurambaj (approved):

PR #929, opened 2023-03-07
Cleanup old manual test command

montedonioluiz:


aurambaj (approved):
Thanks for catching that!

PR #940, opened 2023-05-18
GetWordsInString Method Updates

hylstonnb:
Fixes issues when text containing HTML tags are provided, causing failures under the spell check.

- Html tags should now be removed from the list of words (result of the method getWordsInString), as they shouldn't be considered as words for spell checks.
- The method getWordsInString returns an empty array if a null or empty string is provided as input. 
- Added unit tests for the getWordsInString method

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/CheckerUtils.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/CheckerUtils.java
@@ -11,30 +11,30 @@ import org.apache.commons.lang3.StringUtils;
 
 public class CheckerUtils {
 
-    /**
-     * Pattern matching for html tags.
-     *
-     * <p>"<" matches an opening angle bracket, "[^>]+" matches one or more characters that are not a
-     * closing angle bracket and ">" matches a closing angle bracket
-     */
-    private static final Pattern HTML_TAGS_PATTERN = Pattern.compile("<[^>]+>");
+  /**
+   * Pattern matching for html tags.
+   *
+   * <p>"<" matches an opening angle bracket, "[^>]+" matches one or more characters that are not a
+   * closing angle bracket and ">" matches a closing angle bracket
+   */
+  private static final Pattern HTML_TAGS_PATTERN = Pattern.compile("<[^>]+>");
 
-    public static List<String> getWordsInString(String str) {
-        if (StringUtils.isEmpty(str)) {
-            return new ArrayList<>();
-        }
-        List<String> words = new ArrayList<>();
-        BreakIterator wordBreakIterator = BreakIterator.getWordInstance(Locale.ENGLISH);
-        str = str.replaceAll(HTML_TAGS_PATTERN.pattern(), "");
-        wordBreakIterator.setText(str);
-        int start = wordBreakIterator.first();
-        for (int end = wordBreakIterator.next();
-             end != BreakIterator.DONE;
-             start = end, end = wordBreakIterator.next()) {
-            if (wordBreakIterator.getRuleStatus() != BreakIterator.WORD_NONE) {
-                words.add(str.substring(start, end));
-            }
-        }
-        return words;
+  public static List<String> getWordsInString(String str) {
+    if (StringUtils.isEmpty(str)) {
+      return new ArrayList<>();
     }
+    List<String> words = new ArrayList<>();
+    BreakIterator wordBreakIterator = BreakIterator.getWordInstance(Locale.ENGLISH);
+    str = str.replaceAll(HTML_TAGS_PATTERN.pattern(), "");
+    wordBreakIterator.setText(str);
+    int start = wordBreakIterator.first();
+    for (int end = wordBreakIterator.next();
+        end != BreakIterator.DONE;
+        start = end, end = wordBreakIterator.next()) {
+      if (wordBreakIterator.getRuleStatus() != BreakIterator.WORD_NONE) {
+        words.add(str.substring(start, end));
+      }
+    }
+    return words;
+  }
 }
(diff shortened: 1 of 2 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/checks/CheckerUtils.java
```diff
@@ -1,15 +1,31 @@
...
 
 public class CheckerUtils {
 
+  /**
+   * Pattern matching for html tags.
+   *
+   * <p>"<" matches an opening angle bracket, "[^>]+" matches one or more characters that are not a
+   * closing angle bracket and ">" matches a closing angle bracket
+   */
+  private static final Pattern HTML_TAGS_PATTERN = Pattern.compile("<[^>]+>");
```
geoarion, line 20: To note, this won't work for html attributes that are localizable but could still be spell-checked, e.g.: the "title=" attribute on links or the "alt=" attribute on images, "value=" or "placeholder=" for input tags, etc.

At this point in time, reducing false positives (<br> tags reported as misspelled) - is likely more important than not spell checking the attributes mentioned as they are much rarer - so this might be Ok for now.

@aurambaj if you have any thoughts here.

hylstonnb: Thanks for your feedback George. It might not be easy to properly identify when strings inside html tags should be considered for spell checking or not. Since this rarely happens, let me know if any of you think this is worth spending time on.

aurambaj: @geoarion didn't you look into using a library to extract words from html text like jsoup? worth pushing more on that you think? We could also restrict to just simple tags (no attribute) which I think were the most reported. Not sure how many reports we had on ones with attributes.

geoarion: @aurambaj The default behaviour from JSoup's text() method on a parsed html Document to extract text from html is equivalent to this - though they do provide the list of elements and attributes that could be iterated on to also get the text from the attributes (or from a pre-approved list of attributes).

That said, the number of instances of misspelled text in html attributes is likely to be small / might not be worth effort to cover.

cli/src/test/java/com/box/l10n/mojito/cli/command/checks/CheckerUtilsTest.java
```diff
@@ -0,0 +1,96 @@
...
+  void weShouldBeAbleTo_getWordsInString_providingASingleWordAsInput() {
+    List<String> result = CheckerUtils.getWordsInString("Hello");
+    assertEquals(1, result.size());
+    assertTrue(result.contains("Hello"));
+  }
+
+  @Test
+  void weShouldBeAbleTo_getWordsInString_providingMultipleWordsAsInput() {
+    List<String> result = CheckerUtils.getWordsInString("Hello, how can I help you today?");
+    assertEquals(7, result.size());
```
aurambaj, line 21: don't we want to check the order here vs only contains?  Should we assert directly on a list of words instead of those many asserts?

cli/src/test/java/com/box/l10n/mojito/cli/command/checks/CheckerUtilsTest.java
```diff
@@ -0,0 +1,96 @@
...
+    assertTrue(result.contains("today"));
+  }
+
+  @Test
+  void weShouldBeAbleTo_getWordsInString_ignoringHtmlTag() {
+    List<String> result =
+        CheckerUtils.getWordsInString("This is a string containing html tag <br>");
+    assertEquals(7, result.size());
+    assertFalse(result.contains("<br>"));
+    assertFalse(result.contains("br"));
```
aurambaj, line 37: same, can we check the exact list?

aurambaj (changes requested):

cli/src/test/java/com/box/l10n/mojito/cli/command/checks/CheckerUtilsTest.java
hylstonnb: Makes totally sense. Addressed, thank you 💯

cli/src/test/java/com/box/l10n/mojito/cli/command/checks/CheckerUtilsTest.java
hylstonnb: also addressed, thank you 💯

cli/src/test/java/com/box/l10n/mojito/cli/command/checks/CheckerUtilsTest.java
```diff
@@ -0,0 +1,76 @@
...
+  void weShouldBeAbleTo_getWordsInString_providingASingleWordAsInput() {
+    List<String> result = CheckerUtils.getWordsInString("Hello");
+    List<String> expectedResult = Collections.singletonList("Hello");
+    assertEquals(expectedResult, result);
+  }
+
+  @Test
+  void weShouldBeAbleTo_getWordsInString_providingMultipleWordsAsInput() {
+    List<String> result = CheckerUtils.getWordsInString("Hello, how can I help you today?");
+    List<String> expectedResult = Arrays.asList("Hello", "how", "can", "I", "help", "you", "today");
```
aurambaj, line 22: I just think about that now, but we also have assertJ library that makes the test a bit better, https://www.javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/AbstractIterableAssert.html#containsExactly(ELEMENT...)

hylstonnb: it really reads better. Done, thank you for the suggestion

hylstonnb: and thank you for providing an example @tedyoung :)

aurambaj (approved):

PR #944, opened 2023-05-19
GetWordsInString Method Changes

hylstonnb:
Fixes issues when text containing HTML tags are provided, causing failures under the spell check.

- Html tags should now be removed from the list of words (result of the method getWordsInString), as they shouldn't be considered as words for spell checks.
- The method getWordsInString returns an empty array if a null or empty string is provided as input.
- Added unit tests for the getWordsInString method


aurambaj (approved):
Thanks Hylston!

PR #945, opened 2023-05-24
GetWordsInString method updates related to wished attributes inside html tags

hylstonnb:
- Creates removeHtmlTagsFromStringButKeepingWishedHtmlAttValues method which is used by the GetWordsInString method. The reason is that we still would like to have spell checks for certain html attributes, but most of the content inside the html tags can be ignored by spell check.
- Added unit tests covering the changes
- We now have a list (WISHED_HTML_ATTRIBUTES_FOR_SPELL_CHECK) where html attributes can be added, so if you think there are more attributes we should insert in that list, please let me know.
- At this time the WISHED_HTML_ATTRIBUTES_FOR_SPELL_CHECK list contains 4 attributes: title, alt, value, and placeholder

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/CheckerUtils.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/CheckerUtils.java
@@ -29,14 +29,10 @@ public class CheckerUtils {
    * tags which contains one of the attributes from the WISHED_HTML_ATTRIBUTES_FOR_SPELL_CHECK list.
    */
   static {
-    StringBuilder pattern = new StringBuilder("<[^>]+(?:");
-    for (String att : WISHED_HTML_ATTRIBUTES_FOR_SPELL_CHECK) {
-      pattern.append(att).append("|");
-    }
-    pattern.deleteCharAt(pattern.length() - 1);
-    pattern.append(")\"(.*?)\".*?>");
+    String pattern =
+        "<[^>]+(?:" + String.join("|", WISHED_HTML_ATTRIBUTES_FOR_SPELL_CHECK) + ")\"(.*?)\".*?>";
 
-    WISHED_HTML_ATTRIBUTES_VALUE_PATTERN = Pattern.compile(pattern.toString());
+    WISHED_HTML_ATTRIBUTES_VALUE_PATTERN = Pattern.compile(pattern);
   }
 
   public static List<String> getWordsInString(String str) {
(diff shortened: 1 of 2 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/checks/CheckerUtils.java
```diff
@@ -17,13 +19,33 @@ public class CheckerUtils {
...
+  private static final List<String> WISHED_HTML_ATTRIBUTES_FOR_SPELL_CHECK =
+      Arrays.asList("title=", "alt=", "value=", "placeholder=");
+
+  /**
+   * Initializes regex pattern for the wished_html_attributes_value_pattern. The regex matches html
+   * tags which contains one of the attributes from the WISHED_HTML_ATTRIBUTES_FOR_SPELL_CHECK list.
+   */
+  static {
+    StringBuilder pattern = new StringBuilder("<[^>]+(?:");
+    for (String att : WISHED_HTML_ATTRIBUTES_FOR_SPELL_CHECK) {
```
aurambaj, line 33: can probably be replaced by something like `stream().collect(joining('|'))`

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/CheckerUtils.java
```diff
@@ -35,4 +57,29 @@ public static List<String> getWordsInString(String str) {
...
+   * Removes all the html tags present in a string, but keeping a wished attribute value (attached
+   * at the end of the string) when applicable.
+   *
+   * @param str A string which may or may not contain html tags.
+   * @return 1 - The same input string if there are no html tags present; or 2 - The input string
+   *     without any html tag content; or 3 - The input string removing the html tags, but attaching
+   *     the value of the attribute (at the end) if matches any of the attributes from the
+   *     WISHED_HTML_ATTRIBUTES_FOR_SPELL_CHECK list.
+   */
+  protected static String removeHtmlTagsFromStringButKeepingWishedHtmlAttValues(String str) {
```
aurambaj, line 71: maybe shorter name `removeHtmlTagsAndKeepWishedAttribute` ?

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/CheckerUtils.java
```diff
@@ -35,4 +57,29 @@ public static List<String> getWordsInString(String str) {
...
+    if (StringUtils.isEmpty(str)) {
+      return str;
+    }
+
+    Matcher matcher = WISHED_HTML_ATTRIBUTES_VALUE_PATTERN.matcher(str);
+    String wishedHtmlAttValue = null;
+    if (matcher.find()) {
```
aurambaj, line 78: does this assume a single translatable attribute per tag?

aurambaj (approved):
Thanks!

PR #946, opened 2023-05-25
CLI checks should not fail when strings are removed from code base (gettext edge case)

aurambaj:
In the case of gettext extractor and PO files there is an edge case that can lead to check existing strings even though they are not changed directly. If the string does not pass the check it will raise a failure which becomes pretty confusing. No error should be raised when removing old strings, or moving them around (include changing files, etc).

See the specific example in the code documentation

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractCliChecker.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractCliChecker.java
@@ -1,8 +1,17 @@
 package com.box.l10n.mojito.cli.command.checks;
 
+import static java.util.function.Function.identity;
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toMap;
+
 import com.box.l10n.mojito.cli.command.extraction.AssetExtractionDiff;
 import com.box.l10n.mojito.okapi.extractor.AssetExtractorTextUnit;
+import com.google.common.collect.Sets;
+import com.google.common.collect.Sets.SetView;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.regex.Pattern;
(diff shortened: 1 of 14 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractCliChecker.java
```diff
@@ -31,14 +37,133 @@ public void setCliCheckerOptions(CliCheckerOptions options) {
...
+   * miss running the check when string were reused somewhere else with yet another comment.
+   *
+   * <p>This checks for any added text unit if there is a matching text unit in "removed" based on
+   * the source and content only. Then check that the "added" comments are a subset of the "removed"
+   * comments. The set is built on the assemption that we can split by '\n'
+   *
+   * @param assetExtractionDiffs
+   * @return
+   */
+  protected List<AssetExtractorTextUnit> getAddedTextUnitsExcludingInconsitentComments(
```
geoarion, line 116: Nit: typo
getAddedTextUnitsExcludingInconsitentComments ->
getAddedTextUnitsExcludingInconsistentComments

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractCliChecker.java
```diff
@@ -31,14 +37,133 @@ public void setCliCheckerOptions(CliCheckerOptions options) {
...
       List<AssetExtractionDiff> assetExtractionDiffs) {
     return assetExtractionDiffs.stream()
-        .map(AssetExtractionDiff::getAddedTextunits)
+        .map(
+            assetExtractionDiff -> {
+              Map<String, AssetExtractorTextUnit> md5ToRemoved =
```
geoarion, line 121: Nit: md5ToBeRemoved or md5Removed?

geoarion (approved):

PR #948, opened 2023-06-07
Fixing Spell Check Notification Format

hylstonnb:
Spell check notifications were being displayed such as:
image

- Changed to display the `The string ... contains misspelled words` just once instead of repeating the phrase when more than one misspelled word was found. Please note that unit tests have good output sample to look at.

- Changed * to ● because the spell check slack notification was not being rendered as bullet point.

- Updated unit tests accordingly


aurambaj:
Nice, thanks for the improvements!

aurambaj (approved):

PR #949, opened 2023-06-09
Fixing Notification Messages Format

hylstonnb:
Follow-up work of [PR948](https://github.com/box/mojito/pull/948)

This PR contains the following changes:

- Changed all the notification message `*` symbols to bullet point symbol to be compatible with other platforms such as slack
- Updated the code to use a same bullet point constant (if changes are needed we can now change it in a single place).
- Added an empty line after the baseMessage. This is helpful when there are more some custom text to show after the base message, so this text will be append after an empty line.
- Unit tests updated accordingly

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractCliChecker.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractCliChecker.java
@@ -19,6 +19,8 @@ public abstract class AbstractCliChecker {
 
   protected CliCheckerOptions cliCheckerOptions;
 
+  public static final String BULLET_POINT = "\t• ";
+
   public static AbstractCliChecker createInstanceForClassName(String className)
       throws CliCheckerInstantiationException {
     try {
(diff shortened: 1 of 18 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractCliChecker.java
```diff
@@ -19,6 +19,8 @@ public abstract class AbstractCliChecker {
 
   protected CliCheckerOptions cliCheckerOptions;
 
+  public static final String BULLET_POINT = "\t● ";
```
aurambaj, line 22: we use "tab" now? any thoughts if it is better than spaces? is the behavior consistent across platforms?

aurambaj:
when you're done with your change, would you squashing/rebasing (git rebase -i HEAD~4 or the applicable number) to just have meaningful commit (here a single commit is probably enough)

cli/src/main/java/com/box/l10n/mojito/cli/command/checks/AbstractCliChecker.java
hylstonnb: The problem is that spaces at the beginning are being ignored/removed on Github. I am using tab now to see if it works, but I am not sure if will work as expected. 

image

hylstonnb:
closing this one as it is going to be replaced by https://github.com/box/mojito/pull/950

PR #953, opened 2023-06-22
Support Adobe Magento CSV language packages 

aurambaj:
The CSV has 2 columns: source, target. The filename is made of the locale only with "_" and is inside the "i18n" directory

See reference documentation: https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/translations/xlate.html#m2devgde-xlate-languagepack

### DIFF
--- a/common/src/main/resources/com/box/l10n/mojito/okapi/filters/okf_table_csv_mojito_adobe_magento.fprm
+++ b/common/src/main/resources/com/box/l10n/mojito/okapi/filters/okf_table_csv_mojito_adobe_magento.fprm
@@ -1,0 +1,9 @@
+#v1
+fieldDelimiter=,
+textQualifier="
+sourceIdColumns=1
+sourceColumns=1
+targetColumns=2
+sourceIdSourceRefs=1
+targetSourceRefs=1
+parametersClass=net.sf.okapi.filters.table.csv.Parameters
--- a/cli/src/main/java/com/box/l10n/mojito/cli/filefinder/file/CSVAdobeMagentoFileType.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/filefinder/file/CSVAdobeMagentoFileType.java
@@ -1,0 +1,37 @@
+package com.box.l10n.mojito.cli.filefinder.file;
+
+import static com.box.l10n.mojito.cli.filefinder.FilePattern.DOT;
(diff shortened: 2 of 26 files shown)


common/src/main/resources/com/box/l10n/mojito/okapi/filters/okf_table_csv_mojito_adobe_magento.fprm
```diff
@@ -0,0 +1,9 @@
+#v1
+fieldDelimiter=,
+textQualifier="
+sourceIdColumns=1
+sourceColumns=1
+targetColumns=2
```
hylstonnb, line 6: Could you let me know why targetColumns is 2 and not 1?

webapp/package-lock.json
```diff
@@ -1044,9 +1044,9 @@
       },
       "dependencies": {
         "minimist": {
-          "version": "0.2.2",
-          "resolved": "https://registry.npmjs.org/minimist/-/minimist-0.2.2.tgz",
-          "integrity": "sha512-g92kDfAOAszDRtHNagjZPPI/9lfOFaRBL/Ud6Z0RKZua/x+49awTydZLh5Gkhb80Xy5hmcvZNLGzscW5n5yd0g=="
+          "version": "0.2.4",
```
hylstonnb, line 1047: is `package-lock.json` file commit intentional?

hylstonnb (commented):
In general LGTM, just have a couple comments

webapp/package-lock.json
aurambaj: nope, thanks for flagging.

common/src/main/resources/com/box/l10n/mojito/okapi/filters/okf_table_csv_mojito_adobe_magento.fprm
aurambaj: it is the index of the column where the target string will be. If you look at the `fr_FR.csv`, you can see the first column is the source (english) and the second column is the translation.

hylstonnb: ok that is the column index, makes sense, thanks

PR #954, opened 2023-06-22
Updating Translations Ready Message

hylstonnb:
Replaces `branchName` and `githubRepository` placeholders accordingly, when they are present in the `Translations ready` message.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/github/BranchNotificationMessageSenderGithub.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/github/BranchNotificationMessageSenderGithub.java
@@ -86,7 +86,7 @@ public class BranchNotificationMessageSenderGithub implements BranchNotification
       githubClient.addCommentToPR(
           branchDetails.getRepository(),
           branchDetails.getPrNumber(),
-          branchNotificationMessageBuilderGithub.getTranslatedMessage());
+          branchNotificationMessageBuilderGithub.getTranslatedMessage(branchName, branchDetails));
       updatePRLabel(
           githubClient,
           branchDetails.getRepository(),
--- a/webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/slack/BranchNotificationMessageBuilderSlack.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/slack/BranchNotificationMessageBuilderSlack.java
@@ -8,10 +8,13 @@ import static com.box.l10n.mojito.slack.request.Attachment.MRKDOWNIN_PRETEXT;
 import static com.box.l10n.mojito.slack.request.Attachment.MRKDWNIN_TEXT;
 
 import com.box.l10n.mojito.service.branch.BranchUrlBuilder;
+import com.box.l10n.mojito.service.branch.notification.github.GithubBranchDetails;
 import com.box.l10n.mojito.slack.request.Action;
 import com.box.l10n.mojito.slack.request.Attachment;
 import com.box.l10n.mojito.slack.request.Field;
 import com.box.l10n.mojito.slack.request.Message;
+import com.google.common.collect.ImmutableMap;
+import com.ibm.icu.text.MessageFormat;
 import java.util.List;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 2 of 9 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/slack/BranchNotificationMessageBuilderSlack.java
```diff
@@ -132,4 +135,15 @@ String getSummaryString(List<String> strings) {
         .map(s -> StringUtils.abbreviate(s, STRING_IN_SUMMARY_ABRREVIATE_LENGHT))
         .collect(Collectors.joining(", "));
   }
+
+  protected String getTranslationsReadyMsg(String branchName) {
+    GithubBranchDetails branchDetails = new GithubBranchDetails(branchName);
```
aurambaj, line 140: That notifier is not Github specific, so using `GithubBranchDetails` would be a problem when it is used for example with a change in Phabricator.

aurambaj: let's make it an option to use or not `GithubBranchDetails` to process the branchName. The  config has to be added to `SlackConfigurationProperties` and then passed down as other current options

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/github/BranchNotificationMessageSenderGithub.java
```diff
@@ -86,7 +86,7 @@ public void sendTranslatedMessage(String branchName, String username, String mes
       githubClient.addCommentToPR(
           branchDetails.getRepository(),
           branchDetails.getPrNumber(),
-          branchNotificationMessageBuilderGithub.getTranslatedMessage());
+          branchNotificationMessageBuilderGithub.getTranslatedMessage(branchName));
```
aurambaj, line 89: should we pass the branch detail here, vs re-splitting?

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/slack/BranchNotificationMessageBuilderSlack.java
```diff
@@ -132,4 +135,15 @@ String getSummaryString(List<String> strings) {
...
+    MessageFormat messageFormat = new MessageFormat(translationsReadyMsg);
+    ImmutableMap<String, Object> messageParamMap =
+        ImmutableMap.<String, Object>builder()
+            .put("branchName", branchName)
+            .put("repository", branchDetails.getRepository())
```
aurambaj, line 145: can we prefix those message params with github eg. `github.repository` , `github.prNumber`, `github.owner` ?

`repository` could/would be the "mojito" repository

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/slack/BranchNotificationMessageBuilderSlack.java
aurambaj: @hylstonnb something like that, need to finish passing the option around but that's the idea

``` 
protected String getTranslationsReadyMsg(String branchName) {
    final Builder<String, Object> messageParamMap = ImmutableMap.<String, Object>builder();
    messageParamMap.put("branchName", branchName)
    if (isGithubPR) {
      GithubBranchDetails branchDetails = new GithubBranchDetails(branchName);
      messageParamMap.put("repository", branchDetails.getRepository());
    }
    MessageFormat messageFormat = new MessageFormat(translationsReadyMsg);
    return messageFormat.format(messageParamMap.build());
  }
```

Need to specify a new config, 
```
static class SlackConfigurationProperties {

    String slackClientId;

    MessageBuilderConfigurationProperties messages = new MessageBuilderConfigurationProperties();

    String userEmailPattern;

    boolean useDirectMessage = false;
    
    boolean isGithubPR = false;
```

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/slack/BranchNotificationMessageBuilderSlack.java
hylstonnb: I tried to rename it using github.repository but seems like `.` cannot be used there, as it caused an error. I replaced to githubRepository.

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/slack/BranchNotificationMessageBuilderSlack.java
hylstonnb: makes sense, addressed, thank you

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/github/BranchNotificationMessageSenderGithub.java
hylstonnb: yes, will change, thanks

webapp/src/main/java/com/box/l10n/mojito/service/branch/notification/slack/BranchNotificationMessageBuilderSlack.java
```diff
@@ -132,4 +139,15 @@ String getSummaryString(List<String> strings) {
         .map(s -> StringUtils.abbreviate(s, STRING_IN_SUMMARY_ABRREVIATE_LENGHT))
         .collect(Collectors.joining(", "));
   }
+
+  protected String getTranslationsReadyMsg(String branchName) {
+    final ImmutableMap.Builder<String, Object> messageParamMap = ImmutableMap.builder();
+    messageParamMap.put("branchName", branchName);
+    if (isGithubPR) {
+      GithubBranchDetails branchDetails = new GithubBranchDetails(branchName);
+      messageParamMap.put("githubRepository", branchDetails.getRepository());
```
aurambaj, line 148: In my original comment, I suggested `github.repository`. This uses `githubRepository`. Is this due to implementation limitation with the former or a stylistic choice? 

@garionpin any preference on the format to use?

hylstonnb: Yes, I have answered that above.
image

aurambaj (approved):

PR #955, opened 2023-06-26
PR should be unblocked when I18N skip checks is applied

hylstonnb:
Changing Github commit status to success when I18N checks are skipped

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/extractioncheck/ExtractionCheckNotificationSenderGithub.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/extractioncheck/ExtractionCheckNotificationSenderGithub.java
@@ -114,6 +114,17 @@ public class ExtractionCheckNotificationSenderGithub extends ExtractionCheckNoti
 
   @Override
   public void sendChecksSkippedNotification() {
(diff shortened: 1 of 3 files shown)


cli/src/main/java/com/box/l10n/mojito/cli/command/extractioncheck/ExtractionCheckNotificationSenderGithub.java
```diff
@@ -118,6 +118,15 @@ public void sendChecksSkippedNotification() {
       githubClients
           .getClient(githubOwner)
           .addCommentToPR(githubRepo, prNumber, GithubIcon.WARNING + " " + checksSkippedMessage);
+      githubClients
```
aurambaj, line 121: is it relevant for this action to be in the `if` statement? This should be done regardless of the message, no?

hylstonnb: Thank you for pointing that out. Looking at that closely, I moved the statement to a higher level method.

aurambaj (approved):

aurambaj (approved):

PR #960, opened 2023-07-12
Translation history modal to show leveraging and integrity check messages

aurambaj:


aurambaj:
new modal: 

Screenshot 2023-07-13 at 1 14 21 AM

Screenshot 2023-07-13 at 1 13 00 AM

old modal: Screenshot 2023-07-13 at 1 11 18 AM

webapp/src/main/java/com/box/l10n/mojito/service/tm/importer/TextUnitBatchImporterService.java
```diff
@@ -279,12 +304,17 @@ void applyIntegrityChecks(
             textUnitForBatchImport.setStatus(currentTextUnit.getStatus());
           } else {
             logger.info(
-                "Integrity check failed for string with source {}, content {}: {}",
+                "Integrity check failed for string with source:\n{}\n\nand content:\n{}",
```
hylstonnb, line 307: there is a warning in intellij saying "fewer arguments provided than placeholders specified." Wouldn't be better just use String.format instead, just to make sure?

hylstonnb (commented):
Thank you for these chances. Adding a few comments

webapp/src/main/java/com/box/l10n/mojito/rest/textunit/TextUnitWS.java
```diff
@@ -230,21 +230,21 @@ public TextUnitAndWordCount getTextUnitsCount(
...
    * @return the translations that matches the search parameters
    * @throws InvalidTextUnitSearchParameterException
    */
-  @RequestMapping(method = RequestMethod.GET, value = "/api/textunits/{textUnitId}/history")
+  @RequestMapping(method = RequestMethod.GET, value = "/api/textunits/{tmTextUnitId}/history")
   @ResponseStatus(HttpStatus.OK)
   @JsonView(View.TranslationHistorySummary.class)
   public List<TMTextUnitVariant> getTextUnitHistory(
-      @PathVariable Long textUnitId,
+      @PathVariable Long tmTextUnitId,
```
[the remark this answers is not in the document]
aurambaj: exact, no other change beside making the code more readable. With the previous name it was not clear at first sight which id exactly it was.

webapp/src/main/java/com/box/l10n/mojito/service/tm/importer/TextUnitBatchImporterService.java
aurambaj: that's the right code though. If the last element is an exception, it will show the stack independently and not interpolating in the template. 

```
logger.info("template {}, {}, {}", 1, 2, exception)
```
will show

```
template 1, 2, {}
stack trace...
```
and not:
```
template 1, 2, message/stack trace...
```

https://www.slf4j.org/faq.html#paramException

hylstonnb: yes, my bad sorry, the warning was in the previous implementation. You actually fixed it now. Thank you

webapp/package-lock.json
```diff
@@ -1044,9 +1044,9 @@
       },
       "dependencies": {
         "minimist": {
-          "version": "0.2.2",
```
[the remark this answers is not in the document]
aurambaj: it is fine since the frontend is changed

PR #966, opened 2023-07-24
Added metrics to Smartling TMS sync

maallen:
Adds latency metrics to the Smartling Third Party Sync to enable a better understanding of where performance bottlenecks may occur.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
@@ -20,12 +20,14 @@ import com.box.l10n.mojito.smartling.request.Binding;
 import com.box.l10n.mojito.smartling.request.Bindings;
 import com.box.l10n.mojito.smartling.response.File;
 import com.box.l10n.mojito.smartling.response.StringInfo;
+import com.google.common.base.Stopwatch;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 import com.google.common.io.Files;
 import io.micrometer.core.annotation.Timed;
+import io.micrometer.core.instrument.MeterRegistry;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
```diff
@@ -407,6 +417,9 @@ public void push(
     List<SmartlingFile> result =
         Stream.concat(singularFiles, pluralFiles).collect(Collectors.toList());
     resultProcessor.processPush(result, options);
+    meterRegistry
+        .timer("ThirdPartyTMSSmartling.push", "repository", repository.getName())
+        .record(totalStopwatch.stop().elapsed());
```
aurambaj, line 422: Not sure if you know about this pattern where the "instrumented code" is wrapped in lambda that is passed to `record()`? 

``` meterRegistry
        .timer(
            "PullRunAssetService.saveTextUnitVariantsMultiRow",
            Tags.of("repositoryId", Objects.toString(repository.getId())))
        .record(
            () -> {
              Instant now = Instant.now();
              jdbcTemplate.update(
                  "delete from pull_run_text_unit_variant where pull_run_asset_id = ? and locale_id = ?",
                  pullRunAsset.getId(),
                  localeId);
              Lists.partition(uniqueTmTextUnitVariantIds, BATCH_SIZE)
                  .forEach(
                      tuvIdsBatch ->
                          saveTextUnitVariantsMultiRowBatch(
                              pullRunAsset, localeId, tuvIdsBatch, now));
            });```

maallen: I'd seen the option but when I tried to use it in some cases I ran into issues around some of the existing implementation variables needing to be `final or effectively final` to be able to be used within a lamda. I then switched to this approach to avoid needing to refactor lots of the existing code to be lamda friendly.

maallen:
@aurambaj Just to note that there is some existing instrumentation on the overall push/pull methods using the `@Timed` annotation 

`  @Timed("SmartlingSync.push")`

however we can't use tags with this approach as the specific runtime repo/locale isn't available at compile time. With this in mind should I remove the above annotation and rename the new stats to match the old stat or leave them as they are in the PR at present?

webapp/src/main/java/com/box/l10n/mojito/service/thirdparty/ThirdPartyTMSSmartling.java
aurambaj: With this pattern the behavior will be a bit different if exception are raise. We'd need a try catch to have the same behavior. 

I also see we started to use "retry with lambda" in that class. So i'M wondering if the re-factoring is warranted. Is it hard to do elegantly? 

Let me know your last thoughts, I could merge like that if you feel it is the best approach.

maallen: I've updated to the lamda metric reporting approach now, all unit tests are passing

PR #971, opened 2023-07-28
Make Quartz jobs optionally recoverable

aurambaj:
Updates the QuartzJobInfo to include "requestRecovery" and use it when instantiating the Quartz job instances.  This allows to define each job to be recoverable independently.


hylstonnb (approved):
LGTM, thanks for fixing

PR #973, opened 2023-08-01
Fixing UserDetailsServiceImpl Injection

hylstonnb:
UserDetailsServiceImpl injection was failing when using 2.6.15 spring boot version.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/security/WebSecurityConfig.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/security/WebSecurityConfig.java
@@ -11,9 +11,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.AdviceMode;
-import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.context.annotation.Primary;
 import org.springframework.http.HttpStatus;
 import org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder;
 import org.springframework.security.config.annotation.authentication.configurers.ldap.LdapAuthenticationProviderConfigurer;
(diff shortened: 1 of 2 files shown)


webapp/src/main/java/com/box/l10n/mojito/security/WebSecurityConfig.java
```diff
@@ -63,6 +61,8 @@ public class WebSecurityConfig extends WebSecurityConfigurerAdapter {
 
   @Autowired UserService userService;
 
+  @Autowired UserDetailsServiceImpl userDetailsService;
```
aurambaj, line 64: This is not equivalent to previous code. 

Previous code creates a `UserDetailsServiceImpl` and register as Primary bean for in the container. It means that if we need such a bean in other class, it will use that instance, regardless if another of the same was defined in springboot. 

The new code is just injecting an existing instance of UserDetailsServiceImpl into the service. That assumes that there is one already registered in the container. 

Did the old code register a bean on top of an existing bean in the container, hence needing the @Primary or was there no bean at all? Does the new code have a bean in the container, and if so is it of the type needed? 

What issue is this trying to solve? What error?

hylstonnb: Yes, I changed the control of the instantiation and configuration of the UserDetailsServiceImpl to Spring (instead of the custom one), by using the @Autowired annotation, and I did it because of the error below (which appeared in spring boot 2.6) "The dependencies of some of the beans in the application context form a cycle:
webSecurityConfig (field com.box.l10n.mojito.security.UserDetailsServiceImpl com.box.l10n.mojito.security.WebSecurityConfig.userDetailsService)"

Also, after examining the code, i couldn't find reasons to have that bean instantiation under our control (instead of being managed by spring)

Do you see any reasons for that?

aurambaj: I missed the @service annotation earlier. So that moves where it is instantiated. Should we add/keep `@Primary` on the @Service then? 

> Also, after examining the code, i couldn't find reasons to have that bean instantiation under our control (instead of being managed by spring) Do you see any reasons for that?

Nothing obvious... maybe it was from some documentation/example.

hylstonnb: Sure, shouldn't hurt to keep @Primary annotation. I will push the change, thanks.

aurambaj (approved):

aurambaj (approved):

PR #974, opened 2023-08-01
Add parameter to allow skipping of plurals in comment checker

maallen:
Allows skipping of plurals in the context and comment checker if an option is passed to the extraction cli checks command


garionpin (approved):

aurambaj (approved):
Thanks!

PR #990, opened 2024-01-04
Fix empty drop creation when empty trans-units are pending translation

wadimw:
## Scenario

An empty trans-unit gets pushed to Mojito.
```xml
<trans-unit id="" resname="100_character_description_" datatype="php">
    <source/>
</trans-unit>
```
It shows up with _Needs Translation_ status but has 0 words.

A drop export is triggered for the repository where all other trans-units have already been translated.

Drop will still get created, because Mojito sees that some trans-unit still needs translation - however, this drop will be empty because a 0-word trans-unit is not included in the drop.

## Fix

Updated drop-export command to check if any *words* need translation, rather than trans-units.

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/DropExportCommand.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/DropExportCommand.java
@@ -168,7 +168,7 @@ public class DropExportCommand extends Command {
         RepositoryStatistic repoStat = repository.getRepositoryStatistic();
         if (repoStat != null) {
             for (RepositoryLocaleStatistic repoLocaleStat : repoStat.getRepositoryLocaleStatistics()) {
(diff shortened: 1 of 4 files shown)


wadimw:
This is probably the simplest solution for this case of empty drop creation problem, though it's not complete.

One could argue that:
a. an empty trans-unit should never have a _Needs Translation_ status
b. an empty trans-unit should not be checked in to Mojito at all
c. an empty trans-unit with _Needs Translation_ status *should* in fact be exported in a drop; a drop import would then accept an empty `<target/>` (just like a push accepts the empty `<source/>`) and just carry on as usual (i.e. advance trans-unit status to _Pending Review_)

Additionally, it's worth noting that there is no way to manually change status of an empty trans-unit in Webapp UI, so if we decide on c. we should fix that as well.

PR #991, opened 2024-01-09
Allow disabling server rotation health check and rotation toggle API

wadimw:
In a single-pod setup, server rotation is not useful. Having it be down by default requires an additional step of issuing an API request upon pod start just to make health check useable. In order not to change the default behaviour, this PR instead adds support to disable rotation support altogether (i.e. disable rotation health indicator and the toggle API endpoint).

Rotation support can be disabled through properties `rotation.enabled=false`.


wadimw:
Note that I refrained from using `ConditionalOnEnabledHealthIndicator` because not only the health indicator has to be disabled, but also API endpoint.

Issue #994, opened 2024-02-02
Feature Request: website user management


mensinda:
What we want is the ability to manage the mojito users directly in the web GUI:

- creating new users
- editing existing users
- deleting users

These operations are currently only possible with the provided CLI.

PR #995, opened 2024-02-02
Add a user management tab (implements #994)

mensinda:
This PR adds a new `User Management` tab where ADMINs and PMs can:

- create new users
- edit existing users
- delete users


testUsers.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+ROLES=(ADMIN PM TRANSLATOR USER)
```
aurambaj, line 3: Can this be moved into a Java integration test. If not, this probably should be still moved to the "test" directory

webapp/src/main/java/com/box/l10n/mojito/react/ReactAppController.java
```diff
@@ -103,6 +110,13 @@ ReactUser getReactUser() {
               reactUser.setGivenName(currentAuditor.getGivenName());
               reactUser.setSurname(currentAuditor.getSurname());
               reactUser.setCommonName(currentAuditor.getCommonName());
+
+              Role role = Role.USER;
+              Authority authority = authorityRepository.findByUser(currentAuditor);
+              if (authority != null) {
+                role = Role.valueOf(authority.getAuthority().replace("ROLE_", ""));
+              }
```
aurambaj, line 118: introduce a function?

What's the advantage to removing the "ROLE_" prefix?

webapp/src/main/resources/public/js/components/users/UserDeleteModal.js
```diff
@@ -0,0 +1,70 @@
...
+
+    mixins: [FluxyMixin],
+
+    propTypes: {
+        intl: intlShape.isRequired,
+    },
+
+    statics: {
+        storeListeners: {
+            "onUserStoreChanged": UserStore,
```
aurambaj, line 21: In recent code, there was an attempt to move to stateless react component and the state passed down from the alt store. See BranchesPage/BranchesSearchResults/etc or ScreenshotsPage/ScreenshotsGrid/etc

You may want to consider using it for those new pages too. It might be easier to swap out "alt" later which is an outdated library (considering redux).

mensinda: I can look into it.

Do you also want completely stateless components (so no `getInitialState` and no `this.setState({...})`), or just how the existing alt stores are used? Especially looking at the `UserModal`, moving the state into a dedicated store would increase the code size quite a bit (dedicated actions, a dedicated store).

webapp/src/main/java/com/box/l10n/mojito/service/security/user/UserService.java
```diff
@@ -24,12 +29,43 @@ public class UserService {
...
+    final Optional<User> currentUser = auditorAwareImpl.getCurrentAuditor();
+    if (currentUser.isEmpty()) {
+      // Can't happen in the webapp because only authenticated users may use
+      // the API endpoints. However, allow this for tests
+      return;
+    }
+    final String currentAuthority = currentUser.get().getAuthorities().iterator().next().getAuthority();
+    final Role currentRole = createRoleFromAuthority(currentAuthority);
+
+    if (ROLE_HIERARCHY.indexOf(currentRole) > ROLE_HIERARCHY.indexOf(role)) {
```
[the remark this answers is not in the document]
aurambaj: > Also I am a bit curious why the Roles / Authorities are present in Mojito in the first place? Is this a Spring requirement or is there something already planed for the future?

For context, the original idea (7+ years ago) was to have authorization but we never got time to actually work on it on the server side. So, you can consider this as starting from scratch and if you find new ways in Spring to implement this, you should certainly consider them and make a proposal.

> Additionally, a [User can technically have multiple Authorities](https://github.com/box/mojito/blob/master/webapp/src/main/java/com/box/l10n/mojito/entity/security/user/User.java#L69-L71). However, practically each user has exactly one as far as I can tell

I think, we mapped to the Spring data model at the time in term of relationship cardinality. We implemented `UserDetails` interface backed by hibernate. Implementation is `UserDetailsImpl`. I see I eventually added `OAuth2UserDetailsImpl` (no handling of roles) but this points to `UserDetail` being a must have to have the overall spring security function properly. 

> Wouldn't it be possible to simplyfy this by using a @OneToOne relationship, or embedding the [Authority](https://github.com/box/mojito/blob/master/webapp/src/main/java/com/box/l10n/mojito/entity/security/user/Authority.java) directly into the user?

I think we need to keep the 1:N relationship. This was standard and probably still is. 

It would probably just replace the role hierarchy added in this PR. Eg. only the ROLE_ADMIN can create/modify any user user. ROLE_EDIT_USER can only CRUD on regular user/translator. 

I need to re-read all docs, that feel more like what I remember, vs having homegrown code to handle fine grain authorization.

webapp/src/main/java/com/box/l10n/mojito/react/ReactAppController.java
mensinda: I have refactored it to use the existing methods in the `UserService`. Additionally, I went ahead with the `ROLE_` enum refactoring to simplify the logic.

webapp/src/main/java/com/box/l10n/mojito/service/security/user/UserService.java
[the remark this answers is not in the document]
mensinda: I have simplified the permission check.

restclient/src/main/java/com/box/l10n/mojito/rest/client/UserClient.java
```diff
@@ -43,8 +47,30 @@ public List<User> getUsers(String username) {
...
+              authenticatedRestTemplate.getForEntityWithQueryParams(
+                      getBasePathForEntity(), new ParameterizedTypeReference<Page<User>>() {}, filterParams
+              );
+
+      Page<User> page = responseEntity.getBody();
+      if (page == null) {
+        break;
+      }
+
+      result.addAll(page.getContent());
```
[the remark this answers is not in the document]
mensinda: I went ahead and changed the code to fetch `Integer.MAX_VALUE`.

testUsers.sh
mensinda: I have removed the script.

webapp/src/main/resources/public/js/components/users/UserDeleteModal.js
aurambaj: Right it requires adding a store and having callbacks passed as props in the components. I found it relatively clean in the end but I haven't touch the frontend in a while, I can't say how much work it is to convert this. 

Since it works as it, I'm fine with the current code if it is too much to migrate.

aurambaj (commented):
@mensinda This is a great change. Thank you for contributing.

I've tested locally a little bit and it is working pretty well. The UI looks good too! 

Small ask for consistency with other pages can we move the create button up? and maybe add some text to say "Add user" ?

Screenshot 2024-02-21 at 3 32 45 PM

Screenshot 2024-02-21 at 3 29 52 PM

Happy to merge soon, though.

webapp/src/main/resources/public/js/components/header/Header.js
```diff
@@ -103,6 +104,14 @@ class Header extends React.Component {
                             ScreenshotsRepositoryActions.getAllRepositories();
                         }
                     }}><NavItem><FormattedMessage id="header.screenshots"/></NavItem></LinkContainer>
+                    {
```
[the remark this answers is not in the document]
mensinda: I have moved the User Management to there

webapp/src/main/resources/public/js/components/users/UserDeleteModal.js
mensinda: I have refactored the frontend to now use `AltContainer`.

restclient/src/main/java/com/box/l10n/mojito/rest/client/UserClient.java
```diff
@@ -43,8 +47,20 @@ public List<User> getUsers(String username) {
...
+    filterParams.put("page", "0");
+    filterParams.put("size", String.valueOf(Integer.MAX_VALUE));
+
+    ResponseEntity<Page<User>> responseEntity =
+        authenticatedRestTemplate.getForEntityWithQueryParams(
+            getBasePathForEntity(), new ParameterizedTypeReference<Page<User>>() {}, filterParams
+        );
+
+    Page<User> page = responseEntity.getBody();
+    if (page == null) {
```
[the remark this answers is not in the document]
mensinda: The null check is now gone.

mensinda:
I have also moved the new user button to the requested location.

---

I have tried formatting with `./mvnw spotless:apply`, however I am getting:

```
[ERROR] Step 'google-java-format' found problem in 'src/main/java/com/box/l10n/mojito/service/security/user/UserService.java':
You are not using latest version on JVM 11+.
Try to upgrade to google-java-format 1.15.0, which may have fixed this problem.
```

If I bump the version, formatting works, but the entire project has changes...

webapp/src/main/resources/public/js/components/users/UserModal.js
```diff
@@ -0,0 +1,271 @@
+import React from "react";
+import {FormattedMessage, injectIntl} from "react-intl";
+import {Button, FormControl, Modal, Form, FormGroup, ControlLabel, Alert, Collapse, Glyphicon} from "react-bootstrap";
+import UserStatics from "../../utils/UserStatics";
+import UserActions from "../../actions/users/UserActions";
```
aurambaj, line 5: Big thanks for doing the refactoring for using "alt"! 

One last suggestion, in other pages, the actions are also passed through props eg.

```
 <AltContainer store={ScreenshotsLocaleStore}>
                                <LocalesDropdown
                                    onSelectedBcp47TagsChanged={(selectedBcp47Tags) => {
                                        ScreenshotsHistoryActions.disableHistoryUpdate();
                                        ScreenshotsPaginatorActions.changeCurrentPageNumber(1);
                                        ScreenshotsLocaleActions.changeSelectedBcp47Tags(selectedBcp47Tags);
                                        ScreenshotsHistoryActions.enableHistoryUpdate();
                                        ScreenshotsPageActions.performSearch();
                                        }}
                                    onDropdownToggle={ScreenshotsLocaleActions.changeDropdownOpen}/>
                            </AltContainer>
```

This way the UI component is truly decoupled from the store (that should help when we want to move say to redux) and can potentially be reused by different page

webapp/src/main/resources/public/js/components/users/UserDeleteModal.js
aurambaj: Thanks a lot, it looks great! I added a minor comment related to passing the actions down via props.

webapp/src/main/resources/public/js/components/header/Header.js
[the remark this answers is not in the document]
aurambaj: Thanks!

webapp/src/main/resources/public/js/components/users/UserModal.js
mensinda: > This way the UI component is truly decoupled from the store

Only theoretically. In practice you likely won't be able to swap them out and will only use the `UserStore`.

> and can potentially be reused by different page

I highly doubt this will be possible, since the components are pretty specific.

> that should help when we want to move say to redux

From what I can tell from https://react-redux.js.org/tutorials/quick-start, this shouldn't be required. As far as I understand the tutorial, you use the slice directly in the component without the props.

aurambaj: agree for the user components it makes little sense. 

That's applicable for components like the LocalesDropdown which should be the same for all pages but if actions are hardcoded in it then it is problematic. Then I guess it boils down to wanting to apply the same pattern everywhere. 

This is mostly suggestions at this points, i'm happy to merge as it if you think that's the way to go

mensinda: I have changed it for the `UserErrorModal` and the `UserDeleteModal`, since it does make sense here.

However, I would be against also changing it for the other components. Especially in the `UserEditModal`, where I would have to use a prop for updating every input field. Moving these calls to props would make the code harder to understand IMHO.

aurambaj:
> I have also moved the new user button to the requested location.
> 
> I have tried formatting with `./mvnw spotless:apply`, however I am getting:
> 
> ```
> [ERROR] Step 'google-java-format' found problem in 'src/main/java/com/box/l10n/mojito/service/security/user/UserService.java':
> You are not using latest version on JVM 11+.
> Try to upgrade to google-java-format 1.15.0, which may have fixed this problem.
> ```
> 
> If I bump the version, formatting works, but the entire project has changes...

I suspect that this is an issue with the Java version that is used by the maven wrapper.  We're currently using Java 17 for the project but maybe the wrapper load another version

It should look like that

```
➜  mojito git:(ja/lib-spring3-0-13) ✗ ./mvnw -v            
Apache Maven 3.8.1 (05c21c65bdfed0f71a2f2ada8b84da59348c4c5d)
Maven home: /Users/jeanaurambault/.m2/wrapper/dists/apache-maven-3.8.1-bin/5jr7905sp869ck60ofuug4vpod/apache-maven-3.8.1
Java version: 17.0.9, vendor: Homebrew, runtime: /usr/local/Cellar/openjdk@17/17.0.9/libexec/openjdk.jdk/Contents/Home
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "14.2.1", arch: "x86_64", family: "mac"
```

To make the maven wrapper use a specific version, you may need to export the JAVA_HOME, something like

```
export JAVA_HOME=/usr/local/Cellar/openjdk@17/17.0.9/libexec/openjdk.jdk/Contents/Home
```

For example, I see this error when my wrapper uses Java 21. It is not exactly the same error as you but similar type of problem

```
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.24.1:apply (default-cli) on project mojito-test-common: Execution default-cli of goal com.diffplug.spotless:spotless-maven-plugin:2.24.1:apply failed: java.lang.reflect.InvocationTargetException: 'com.sun.tools.javac.tree.JCTree com.sun.tools.javac.tree.JCTree$JCImport.getQualifiedIdentifier()' -> [Help 1]
``` 

with 

```
➜  mojito git:(ja/lib-spring3-0-13) ✗ ./mvnw -v
Apache Maven 3.8.1 (05c21c65bdfed0f71a2f2ada8b84da59348c4c5d)
Maven home: /Users/jeanaurambault/.m2/wrapper/dists/apache-maven-3.8.1-bin/5jr7905sp869ck60ofuug4vpod/apache-maven-3.8.1
Java version: 21.0.2, vendor: Eclipse Adoptium, runtime: /Library/Java/JavaVirtualMachines/temurin-21.jdk/Contents/Home
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "14.2.1", arch: "x86_64", family: "mac"
```

aurambaj:
I'm also updating the Java install instruction here: https://github.com/pinterest/mojito/pull/99

mensinda:
I have managed to get the formatter to run on a different machine and the CI is green now.

aurambaj:
Thanks a lot @mensinda

PR #998, opened 2024-02-23
Always show the asset path in the workbench

mensinda:
This moves the resource button next to the locale. Additionally, the resource path of a text is now always shown and not hidden behind a tooltip:

tmp

### DIFF
--- a/webapp/src/main/resources/properties/en.properties
+++ b/webapp/src/main/resources/properties/en.properties
@@ -225,6 +225,13 @@ search.result.empty=There are no results. Update your search parameter to find w
 # Message displayed when the user has not selected repository and locales
 search.result.selectRepoAndLocale=Select at least one repository and locale before you can search.
 
+search.viewMode.title=View mode: {mode}
+
+search.viewMode.FULL=Full
+search.viewMode.REDUCED=Reduced
+search.viewMode.STANDARD=Standard
+search.viewMode.COMPACT=Compact
+
 # Label used in the header loggedInUser dropdown to show user's profile
 header.loggedInUser.profile=My Profile
 
@@ -361,6 +368,9 @@ textUnit.saveVirtualAssetTextUnit.failed=Couldn't not set translate attribute, c
 
 # Confirmation message
 textUnit.cancel.confirmation=Are you sure you want to cancel editing? You will lose the changes you've made.
+
+textUnit.tag.repo=repo
+textUnit.tag.asset=asset
 
 # Label displayed on the repositories dropdown in the workbench
 search.repository.btn.text={numberOfSelectedRepositories, plural, =0{Repositories} =1{1 repository} other{# repositories}}
--- a/webapp/src/main/resources/public/js/components/workbench/TextUnit.js
+++ b/webapp/src/main/resources/public/js/components/workbench/TextUnit.js
@@ -32,6 +32,7 @@ import {
     OverlayTrigger,
     Tooltip
 } from "react-bootstrap";
+import ViewModeStore from "../../stores/workbench/ViewModeStore";
 
 
 let TextUnit = createReactClass({
@@ -218,7 +219,7 @@ let TextUnit = createReactClass({
         textUnit.setStatus(TextUnitSDK.STATUS.APPROVED);
         textUnit.setTarget(this.state.translation);
         textUnit.setTranslated(true);
-        
+
         return textUnit;
     },
 
@@ -382,7 +383,7 @@ let TextUnit = createReactClass({
         let rendered = '';
         if (!this.props.textUnit.isUsed()) {
             rendered = (
-                <Label bsStyle="default" className="mrxs"><FormattedMessage id="textUnit.unused"/></Label>
+                <Label bsStyle="default"><FormattedMessage id="textUnit.unused"/></Label>
             );
         }
         return rendered;
@@ -397,7 +398,7 @@ let TextUnit = createReactClass({
         let rendered = '';
         if (this.props.textUnit.getDoNotTranslate()) {
             rendered = (
-                <Label bsStyle="default" className="mrxs"><FormattedMessage id="textUnit.doNotTranslate"/></Label>
+                <Label bsStyle="default"><FormattedMessage id="textUnit.doNotTranslate"/></Label>
             );
         }
         return rendered;
@@ -412,7 +413,7 @@ let TextUnit = createReactClass({
         let rendered = '';
         if (this.props.textUnit.getPluralForm() != null) {
             rendered = (
-                <Label bsStyle="default" className="mrxs clickable" onClick={this.onPluralFormLabelClick}>
+                <Label bsStyle="default" className="clickable" onClick={this.onPluralFormLabelClick}>
                     {this.props.textUnit.getPluralForm()}
                 </Label>
             );
@@ -495,7 +496,7 @@ let TextUnit = createReactClass({
                                  dir={dir}
                                  onClick={this.onClickTextArea}/>
 
-                    <ButtonToolbar className="mtxs mbxs">
+                    <ButtonToolbar className="mtxs">
                         <Button bsStyle='primary' bsSize="small" disabled={saveDisabled}
                                 onClick={!saveDisabled ? this.saveTextUnitIfNeeded : null}>
                             <FormattedMessage id='label.save'/>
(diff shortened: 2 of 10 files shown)


aurambaj:
@mensinda  I like the concept in fact I always wanted to have the asset somehow directly shown. 

One problem though is that it is pretty common for it to be somewhat long which 
1) will make the UI cluttered with repetition 
2) overflow badly...  

for the overflow maybe we can restrict it to max size, truncating the start and just show the end of the path. Then when hovering on the text, we'd see the whole path.

for the cluttered part, I've been thinking to have a selector to choose different "view mode" next to the pagination. It would show more or less information based on the mode. 

One other info pretty useful to have is the "repository name" when doing cross repository searches.

So we could have a condensed view, which the current display and and expanded view which would show the repository name and repository path. To avoid having to deal with length of the path, I think it would make sense to just have the path on its own line.

Any thoughts on that? Wondering if you have other information besides the path that you feel should be directly accessible.

mensinda:
> for the overflow maybe we can restrict it to max size, truncating the start and just show the end of the path

Implemented by setting the max width to `256px` and `overflow: hidden`.

> for the cluttered part, I've been thinking to have a selector to choose different "view mode" next to the pagination

Implemented.

> One other info pretty useful to have is the "repository name" when doing cross repository searches.

I have also added this.

> Any thoughts on that? Wondering if you have other information besides the path that you feel should be directly accessible.

Not that I can currently think of.

aurambaj:
Thanks @mensinda ! Merging the PR as it but wondering about the use of color, ideally I'd avoid adding to many also it is not super obvious what the color conveys in that case. I'm thinking that may in that expended view it could be more explicit

Potentially, on its own line before the locale/string id, Have 2 labels prefixed with "repo" eg. "repo: demo" and other "asset: demo.properties". I think "grey" would be fine for those

webapp/src/main/resources/public/js/components/workbench/ViewModeDropdown.js
```diff
@@ -0,0 +1,40 @@
...
+                <FormattedMessage id={"search.viewMode." + mode} />
+            </MenuItem>
+        );
+    },
+
+    render() {  
+        return (
+            <DropdownButton  
+                id="WorkbenchStatusDropdown" 
+                title={this.props.intl.formatMessage({id: "search.viewMode.title"})}
```
aurambaj, line 29: Maybe update the dropdown title with the selected mode?

webapp/src/main/resources/public/js/components/workbench/TextUnit.js
```diff
@@ -841,17 +868,66 @@ let TextUnit = createReactClass({
         );
     },
 
+    renderRepository() {
+        if (this.props.viewMode.viewMode !== ViewModeStore.VIEW_MODE.FULL) {
+            return;
+        }
+
+        return (
+            <Label bsStyle='danger' bsSize='large' className="clickable" onClick={this.onRepositoryClick}>
```
aurambaj, line 877: I don't think we should be using the "danger" style here since it is just regular information. see the comment on the whole PR. Maybe having a prefix on the label to say "repository" or "repo" and put it on its own line?

webapp/src/main/resources/public/js/components/workbench/TextUnit.js
```diff
@@ -841,17 +868,66 @@ let TextUnit = createReactClass({
...
+        let displayAssetPath = assetPathWithZeroWidthSpace;
+
+        if (this.props.viewMode.viewMode === ViewModeStore.VIEW_MODE.REDUCED) {
+            // Remove file extension
+            displayAssetPath = displayAssetPath.replace(/\.[^.]+/, '');
+        }
+
+        return (
+            <OverlayTrigger placement="top" overlay={assetPathTooltip}>
+                <Label style={{maxWidth: "256px", overflow: "hidden"}} bsStyle='info' bsSize='large' className="clickable" onClick={this.onAssetPathClick}>
```
aurambaj, line 901: same comment on the style/color and the prefix/be on its own line.

webapp/src/main/resources/properties/en.properties
```diff
@@ -222,6 +222,12 @@ search.result.empty=There are no results. Update your search parameter to find w
 # Message displayed when the user has not selected repository and locales
 search.result.selectRepoAndLocale=Select at least one repository and locale before you can search.
 
+search.viewMode.title=View mode
+
+search.viewMode.FULL=Full
+search.viewMode.REDUCED=Reduced
```
aurambaj, line 228: Reduced -> Standard?

webapp/src/main/resources/public/js/stores/workbench/ViewModeStore.js
```diff
@@ -0,0 +1,28 @@
+import alt from "../../alt";
+import ViewModeActions from "../../actions/workbench/ViewModeActions";
+import { readCookie, writeCookie } from "../../utils/Cookie";
```
aurambaj, line 3: can we store this in local storage instead since it is client side only configuration?

webapp/src/main/resources/properties/en.properties
mensinda: Maybe I just add a 4th standard state (and make it the default) which has the same information as the current layout?

webapp/src/main/resources/public/js/stores/workbench/ViewModeStore.js
mensinda: Do you mean with `local storage` inside the DB in the server?

If you mean just saved in JS (it is gone if you refresh the page), then I would like to keep it a little more persistent so that I don't have to change it back every time.

webapp/src/main/resources/public/js/components/workbench/TextUnit.js
mensinda: Also, adding a new line is not that easy, since most of the space is already used. So adding a new line in the label, would make each text line even bigger.

Would you be fine with using the `secondary` color for the repository?

webapp/src/main/resources/public/js/stores/workbench/ViewModeStore.js
aurambaj: browser [local storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage), that should support browser restart, etc

webapp/src/main/resources/public/js/components/workbench/TextUnit.js
aurambaj: btw, I just remembered that the asset info is already accessible by hovering on the text unit (and clicking on the icon drills down to search in that asset). I guess that's the approach i had to avoid cluttering everything at the time. 

we already have a few label, eg. "unused" can be added to the right of the locale.

I think we should keep the danger color for things that are actually problematic and not for regular display. 

In the full view, we can use more line for clarify, I think. Maybe something like. 

Screenshot 2024-02-28 at 2 20 00 PM

webapp/src/main/resources/public/js/components/workbench/ViewModeDropdown.js
mensinda: Done.

webapp/src/main/resources/properties/en.properties
mensinda: This is now implemented

webapp/src/main/resources/public/js/stores/workbench/ViewModeStore.js
mensinda: Ah, I wasn't aware of that. I have changed it to the local storage now.

webapp/src/main/resources/public/js/components/workbench/TextUnit.js
mensinda: Full:

tmp

Reduced:

tmp

webapp/src/main/resources/sass/mojito.scss
```diff
@@ -462,15 +462,51 @@ div.textunit label {
...
+.text-unit-root .left {
+    flex-grow: 1;
+    height: 100%;
+    display: grid;
+    gap: 7px;
+    align-items: center;
+
+    grid-template-columns: min-content min-content min-content fit-content(100%) 1fr min-content min-content;
+    grid-template-rows: min-content 1fr min-content min-content 1fr;
+    grid-template-areas:
```
aurambaj, line 488: nice

aurambaj (approved):

PR #999, opened 2024-02-28
Enforce serving content as UTF-8

mensinda:
Without this change, the content is sometimes served as `ISO-8859-1` instead of `UTF-8`.

Unfortunately, I couldn't find a solution that works with the `ModelAndView` (setting `produces` and / or `headers` in the `@RequestMapping`, `ResponseEntity<ModelAndView>` is not supported / generates an error, etc.). Thus, I had to refactor the code to use mustache directly, so that the header could be set via the `ResponseEntity<String>`.


aurambaj:
> Without this change, the content is sometimes served as `ISO-8859-1` instead of `UTF-8`.
> 

isn't client sensitive? under which circumstance did it happen? what's the result/issue caused? 

I haven't noticed similar issue so trying to understand a bit better the issue

> Unfortunately, I couldn't find a solution that works with the `ModelAndView` (setting `produces` and / or `headers` in the `@RequestMapping`, `ResponseEntity<ModelAndView>` is not supported / generates an error, etc.). Thus, I had to refactor the code to use mustache directly, so that the header could be set via the `ResponseEntity<String>`.

mensinda:
> isn't client sensitive? under which circumstance did it happen?

Yes, If i change my browser language to German / Deutsch, spring will set the content type automagially to `text/html;charset=ISO-8859-1`. This problem doesn't occur when the browser language is english, then spring correctly uses `text/html;charset=UTF-8`

tmp

`charset=ISO-8859-1` is wrong in this case, because the encoding of the `index.html` is still `UTF-8` and the JS scripts also insert `UTF-8` characters (zero with whitespace / special UTF-8 whitespace characters, for instance, which are rendered wrong in the screenshot above). Thus, mojito should *always* serve content as `UTF-8` regardless of the browser language.

aurambaj:
I was able to reproduce in German. Also tried in French but this does not happen. There must something configurable with spring 🤔 

Screenshot 2024-03-01 at 11 18 13 AM


Screenshot 2024-03-01 at 11 16 19 AM

mensinda:
Regardless of what spring thinks it's right, it is clearly wrong in this case. I *think* I saw a property that controls it somewhere that controls this behavior, but I am unable to find it again. Regardless, IMHO it would be better to explicitly set the locale at the location where it is required, than configuring spring somewhere else and it potentially silently breaking with future spring releases.

PR #1000, opened 2024-02-28
Restrict access to endpoints for some roles

mensinda:
Restrict access to endpoints for some roles:

- `USER`s can only view and not edit
- `TRANSLATOR`s can translate
- `ADMIN`s and `PM`s can use more management functions

### DIFF
--- a/webapp/src/main/resources/public/js/components/drops/Drops.js
+++ b/webapp/src/main/resources/public/js/components/drops/Drops.js
@@ -26,6 +26,7 @@ import ImportDropConfig from "../../sdk/drop/ImportDropConfig";
 import NewDropModal from "./NewDropModal";
 import RepositoryActions from "../../actions/RepositoryActions";
 import ConfirmationModal from "../widgets/ConfirmationModal";
+import AuthorityService from "../../utils/AuthorityService";
 
 let Drops = createReactClass({
     displayName: 'Drops',
@@ -276,7 +277,7 @@ let Drops = createReactClass({
 
         return (
             <tr key={`Drops.repositoryRow.${drop.name}.${drop.repository.name}`} className={rowClass}>
-                <td>{drop.name}{this.getButtonControlBar(drop)}</td>
+                <td>{drop.name}{AuthorityService.canEditProjectRequests() && this.getButtonControlBar(drop)}</td>
                 <td>{drop.repository.name}</td>
                 <td><FormattedNumber value={wordCount}/></td>
                 <td><FormattedDate value={drop.createdDate} day="numeric" month="long" year="numeric"/></td>
@@ -463,7 +464,7 @@ let Drops = createReactClass({
         return (
             <div className="pull-right mrm mlm">
                 <Button bsStyle="primary" bsSize="small" className="new-request-button"
-                        onClick={this.onClickNewRequest}>
+                        onClick={this.onClickNewRequest} disabled={!AuthorityService.canEditProjectRequests()}>
                     <FormattedMessage id="drops.newRequest.btn"/>
                 </Button>
             </div>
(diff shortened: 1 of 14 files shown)


webapp/src/main/resources/public/js/components/screenshots/ScreenshotViewerModal.js
```diff
@@ -96,7 +97,7 @@ class ScreenshotViewerModal extends React.Component {
                         <div className="branches-screenshotviewer-modal-delete-container">
                             {this.props.isDeleting ? <span className="glyphicon glyphicon-refresh spinning"/> :
                                 <OverlayTrigger trigger="click" placement="top" overlay={this.popover()}>
-                                    <Button bsStyle="danger" style={{fontSize: 11}}>
+                                    <Button bsStyle="danger" style={{fontSize: 11}} disabled={!AuthorityService.canEditScreenshots()}>
```
aurambaj, line 100: for UI components that are stateless/exclusively use "props" I'd rather keep them like this instead of having a dependency on AuthorityService.

For others that state, I guess it is fine to have that AuthorityService depenedency

webapp/src/main/resources/public/js/components/drops/Drops.js
```diff
@@ -276,7 +277,7 @@ let Drops = createReactClass({
 
         return (
             <tr key={`Drops.repositoryRow.${drop.name}.${drop.repository.name}`} className={rowClass}>
-                <td>{drop.name}{this.getButtonControlBar(drop)}</td>
+                <td>{drop.name}{AuthorityService.canEditProjectRequests() && this.getButtonControlBar(drop)}</td>
```
aurambaj, line 280: also see later comment. Here i'm fine calling the AuthorityService directly because there is already some state in the component.

For stateless component, I'd rather avoid though

webapp/src/main/resources/public/js/components/screenshots/Screenshot.js
```diff
@@ -97,6 +98,7 @@ class Screenshot extends React.Component {
                     <span className="screenshot-glyph">
                         <StatusGlyph status={this.props.screenshot.status} 
                                      onClick={this.props.onStatusGlyphClick}
+                                     disabled={!AuthorityService.canEditScreenshots()}
```
aurambaj, line 101: see other comments too - can we keep that stateless and with no dependency on AuthorityService

webapp/src/main/resources/public/js/components/screenshots/ScreenshotViewerModal.js
mensinda: OK, I have refactored it.

webapp/src/main/resources/public/js/components/screenshots/Screenshot.js
mensinda: Also refactored

aurambaj (approved):

aurambaj:
Thank you!

aurambaj:
@mensinda there are conflicts with master, since I merged the other PR.

mensinda:
I have rebased the PR

ehoogerbeets:
Daniel, this is very exciting! I have been wanting this feature for some time now. Q for you: will these types of restrictions by role also be enforced on the backend as well?

mensinda:
@ehoogerbeets Yes, with the changes in the [WebSecurityConfig](https://github.com/box/mojito/pull/1000/files#diff-dc3400c53361654b474a9d2ce88de15b2725d8257254b8892432572be592240d). I can't guarantee that I caught all cases, but from my limited understanding this should cover most of it.

PR #1001, opened 2024-02-28
Add password change modal

mensinda:
This allows users, regardless of there role, to change their password.

I didn't use a dedicated store here, because that would be way to overkill IMHO. The Component is pretty much self contained, so there is no real benefit for using a store, besides using the same pattern everywhere (eventually maybe).

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/rest/security/PasswordChangeRequest.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/rest/security/PasswordChangeRequest.java
@@ -1,0 +1,3 @@
+package com.box.l10n.mojito.rest.security;
+
+public record PasswordChangeRequest(String currentPassword, String newPassword) {}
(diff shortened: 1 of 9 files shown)


webapp/src/main/java/com/box/l10n/mojito/rest/security/PasswordChangeRequest.java
```diff
@@ -0,0 +1,24 @@
+package com.box.l10n.mojito.rest.security;
+
+import java.io.Serializable;
+
+public class PasswordChangeRequest implements Serializable {
```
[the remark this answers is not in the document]
mensinda: I have switched it to a record.

webapp/src/main/java/com/box/l10n/mojito/rest/security/PasswordChangeRequest.java
```diff
@@ -0,0 +1,6 @@
+package com.box.l10n.mojito.rest.security;
+
+import java.io.Serializable;
+
+public record PasswordChangeRequest(String currentPassword, String newPassword)
+    implements Serializable {}
```
aurambaj, line 6: For simplicity sake I'd remove "implements Serializable", record do implement it anyway.

mensinda: changed it.

PR #1002, opened 2024-02-29
Add a dropdown for selecting the page size on the workbench

mensinda:
tmp

### DIFF
--- a/webapp/src/main/resources/properties/de.properties
+++ b/webapp/src/main/resources/properties/de.properties
@@ -226,6 +226,8 @@ search.viewMode.REDUCED=Reduziert
 search.viewMode.STANDARD=Standard
 search.viewMode.COMPACT=Kompakt
 
+search.unitsPerPage=Texte pro Seite: {pageSize}
+
 # Label used in the header loggedInUser dropdown to show user's profile
 header.loggedInUser.profile=Mein Profil
 
--- a/webapp/src/main/resources/properties/en.properties
+++ b/webapp/src/main/resources/properties/en.properties
@@ -235,6 +235,8 @@ search.viewMode.REDUCED=Reduced
 search.viewMode.STANDARD=Standard
 search.viewMode.COMPACT=Compact
 
+search.unitsPerPage=Page size: {pageSize}
+
 # Label used in the header loggedInUser dropdown to show user's profile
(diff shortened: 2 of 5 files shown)


aurambaj:
brainstorming on the label...

I feel like it should be either "Text Units per page" or "Per Page: 100". Leaning toward "Per Page: 100" since it is shorter. It is a little less explicit but given the context it is still understandable. Thoughts? 

> tmp

mensinda:
@aurambaj How about "Page size: 100"?

mensinda:
Rebased, and I switched the english text to "Page size". However, I would like to keep the german translation, because "Seitengröße" sounds weird.

ehoogerbeets:
Daniel, Box has all the translations for mojito in (surprise!) an internal instance of mojito. I'll add this translation manually.

@aurambaj What do you think of this idea? We can make a drop from this instance and share it publically on Box and then solicit volunteers to help us translate it?

mensinda:
Since this is a JS only change, I would say that the CI error seems unrelated

aurambaj:
> Daniel, Box has all the translations for mojito in (surprise!) an internal instance of mojito. I'll add this translation manually.
> 
> @aurambaj What do you think of this idea? We can make a drop from this instance and share it publically on Box and then solicit volunteers to help us translate it?

@mensinda did contribute translations manually to the German file.

Typically, we should not edit the properties file by hand but we don't have a public instance of Mojito where people could collaborate / update strings.

We can always re-import the current properties files into a Mojito instance, there is a command for that.

@ehoogerbeets when you say add "this manually" you mean generating a PR from the Box Mojito instance that contains update properties files? 

Is Box still translating new incoming strings or are the translation available only for older strings? There used to be a bot pushing new translation but last commit if from 5 years ago. 

To have volunteer help translate, we'd need some public instance running. Drop could work or we just re-create the project from the properties files.

ehoogerbeets:
The "bot" is still there but hasn't been operating for a number of years as we are not sending the strings to our vendor. It could be resurrected again though, if for nothing else to make sure that the properties files for the other languages have the same entries as the English one does. As you say, we would have to re-import the current properties files into mojito in order to preserve the translations already contributed by the open-source community.

Issue #1026, opened 2025-06-23
Project request tab is not loading after doing any actions project is requested/ doing import


Dharshini31:
https://github.com/user-attachments/assets/aa7abbef-95be-489f-b797-b08788fea65d

PR #1045, opened 2026-01-27
Fix paginated Drop data fetching

wadimw:
## Lazy loading of `Drop#createdByUser`

API response for `GET /api/drops` sometimes returns `createdByUser: null`, even when corresponding database field (column `created_by_user_id` in table `drops`) has correct data. This causes `/project-requests` page to crash with `TypeError` when accessing `createdByUser.getDisplayName()` here: https://github.com/box/mojito/blob/2e3b87fa2b409438ee73643e6f131f7485d0c8f7/webapp/src/main/resources/public/js/components/drops/Drops.js#L284

Retrieval of drops is done through JPA Hibernate - specifically using `DropRepository#findAll` (https://github.com/box/mojito/blob/2e3b87fa2b409438ee73643e6f131f7485d0c8f7/webapp/src/main/java/com/box/l10n/mojito/rest/drop/DropWS.java#L80). In legacy Mojito (v0.110) this was a builtin Hibernate method, but it was overriden to use a NamedEntityGraph in https://github.com/box/mojito/commit/1d95b1049391d8ef57bab54738764a8df45ccc34 (as a followup/bugfix to changes introduced in https://github.com/box/mojito/commit/33e88bc52c0a9f3207e75ba94d8ac9330ec58f7d).

Because this graph does not include `createdByUser`, `DropWS#getDrops` response would sometimes not include User data due to lazy loading. We should update the code to ensure that User data is always included in the response.

This PR updates NamedEntityGraph `Drop.legacy` to include `createdByUser` to ensure eager User loading for Drops.

## Inefficient Drop pagination

In addition to the above, we've discovered that calling `DropRepository#findAll` results in the following warning log:

```log
HHH90003004: firstResult/maxResults specified with collection fetch; applying in memory
```

This is a limitation of Hibernate, which can't apply pagination at the database level when fetching collections. Due to that, it falls back to fetching ALL matching drops from the database (along with other data specified in the EntityGraph) and then applying pagination in memory. This is inefficient and can even cause OOM issues.

This PR reverts the EntityGraph addition from paginated `DropRepository#findAll`. Instead, it implements manual data initialization _after_ the paginated query result is retrieved. This behaviour is encapsulated in `DropService#findAll`.

Note that this pattern of manual data initialization has previously been implemented in the codebase for UserRepository (as described in https://github.com/box/mojito/blob/33e88bc52c0a9f3207e75ba94d8ac9330ec58f7d/webapp/src/main/java/com/box/l10n/mojito/service/security/user/UserRepository.java#L29-L33).

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/drop/DropService.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/drop/DropService.java
@@ -34,9 +34,13 @@ import java.util.List;
 import java.util.NoSuchElementException;
 import net.sf.okapi.common.exceptions.OkapiBadFilterInputException;
 import net.sf.okapi.common.exceptions.OkapiIOException;
+import org.hibernate.Hibernate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.jpa.domain.Specification;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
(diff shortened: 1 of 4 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/drop/DropService.java
```diff
@@ -495,4 +499,36 @@ public void completeDrop(Drop drop) {
...
+          if (drop.getImportPollableTask() != null) {
+            Hibernate.initialize(drop.getImportPollableTask());
+            Hibernate.initialize(drop.getImportPollableTask().getSubTasks());
+          }
+          if (drop.getExportPollableTask() != null) {
+            Hibernate.initialize(drop.getExportPollableTask());
+            Hibernate.initialize(drop.getExportPollableTask().getSubTasks());
+          }
+          Hibernate.initialize(drop.getTranslationKits());
+          if (drop.getTranslationKits() != null) {
```
ehoogerbeets:
```suggestion
          if (drop.getTranslationKits() != null) {
              Hibernate.initialize(drop.getTranslationKits());
```

ehoogerbeets: Don't know if that is correct, but that is the pattern for the previous lines in this method...

wadimw: `Hibernate.initialize` accepts nullable values so it's fine, but you're right in that it's inconsistent this way. Added 9471a9c which _streamlines_ (pun intended) optional graph relations handling.

ehoogerbeets (approved):

wadimw:
Possibly the same issue as https://github.com/box/mojito/issues/1026

Issue #1021, opened 2025-03-19
Partial Importing in 'Project Request' string units


azhagusundaram-t-6139:
I followed the steps provided in the documentation.
[https://www.mojito.global//docs/guides/project-request/](https://www.mojito.global//docs/guides/project-request/)

1. Created the Project Request
2. Mojito successfully created the '.xliff' files
3. I shared them to translation vendors
4. Placed the translated '.xliff' files in the 'Localised Files' folder
5. Now, I try to import those strings by clicking the 'Import Request' button
<img width="435" alt="Image" src="https://github.com/user-attachments/assets/1027334a-2ee9-4527-a662-e4488bb945da" />

6. I have 60 units of translated strings. **Out of 60 first 23 strings have imported and not others!** 
7. Status has been marked as 'Import Failed' 
<img width="440" alt="Image" src="https://github.com/user-attachments/assets/57e211da-1816-450f-a98a-f5d8a0bfcea0" />

8. Now, if I manually removed those 23 strings from the translated xliff file, again some set of strings got imported and some are not imported!
9. I'm repeating this manually for each language translation files! Tedious!!!

**Note: I'm currently using the latest 'master' branch**

Please fix this or guide me where I can debug this 🙏

ehoogerbeets:
Also, if you are using Box as the way to send and receive files to/from your translation vendor, then you can look at the xliffs in the Localized Files folder on Box to see what the problem might be. If the import fails due to something in a file, it puts some metadata on that file which gives a clue as to what the problem might be. Sometimes it is character set, sometimes it is the state attribute, sometimes something else?

PR #1049, opened 2026-02-03
Fix unexpected document end when importing drops with large XLIFF files

wadimw:
This PR fixes the following error:

```log
com.box.l10n.mojito.service.drop.importdropexception: com.ctc.wstx.exc.wstxeofexception: unexpected eof; was expecting a close tag for element <note>
 at [row,col,system-id]: [128,206,"/some/file/path/to/be/read/from/db"]
```

which would occur when importing a drop containing XLIFF files larger than 8 KiB. On the UI side, this would appear as `Import Failed` in the `Project Requests` page, and would result in a partial import (translations for strings before the 8KiB mark would be imported correctly).

---

<details>

<summary>Click here to see the full stack trace</summary>

```log
2026-01-29t07:42:26.609-08:00 debug 97 --- [ pollabletask-5] c.b.l.mojito.service.drop.dropservice    : error when importing file, keep importing other files

com.box.l10n.mojito.service.drop.importdropexception: com.ctc.wstx.exc.wstxeofexception: unexpected eof; was expecting a close tag for element <note>
 at [row,col,system-id]: [128,206,"/some/file/path/to/be/read/from/db"]
	at com.box.l10n.mojito.service.drop.dropservice.updatetmwithlocalizedxliff_aroundbody10(dropservice.java:341)
	at com.box.l10n.mojito.service.drop.dropservice$ajcclosure11.run(dropservice.java:1)
	at org.aspectj.runtime.reflect.joinpointimpl.proceed(joinpointimpl.java:270)
	at com.box.l10n.mojito.service.pollabletask.pollablecallable.call(pollablecallable.java:50)
	at java.base/java.util.concurrent.futuretask.run(futuretask.java:317)
	at com.box.l10n.mojito.service.pollabletask.pollableaspect.syncexecute(pollableaspect.java:111)
	at com.box.l10n.mojito.service.pollabletask.pollableaspect.ajc$inlineaccessmethod$com_box_l10n_mojito_service_pollabletask_pollableaspect$com_box_l10n_mojito_service_pollabletask_pollableaspect$syncexecute(pollableaspect.java:1)
	at com.box.l10n.mojito.service.pollabletask.pollableaspect.createpollablewrapper(pollableaspect.java:79)
	at com.box.l10n.mojito.service.drop.dropservice.updatetmwithlocalizedxliff(dropservice.java:331)
	at com.box.l10n.mojito.service.drop.dropservice.importfile_aroundbody8(dropservice.java:313)
	at com.box.l10n.mojito.service.drop.dropservice$ajcclosure9.run(dropservice.java:1)
	at org.aspectj.runtime.reflect.joinpointimpl.proceed(joinpointimpl.java:270)
	at com.box.l10n.mojito.service.pollabletask.pollablecallable.call(pollablecallable.java:50)
	at java.base/java.util.concurrent.futuretask.run(futuretask.java:317)
	at com.box.l10n.mojito.service.pollabletask.pollableaspect.syncexecute(pollableaspect.java:111)
	at com.box.l10n.mojito.service.pollabletask.pollableaspect.ajc$inlineaccessmethod$com_box_l10n_mojito_service_pollabletask_pollableaspect$com_box_l10n_mojito_service_pollabletask_pollableaspect$syncexecute(pollableaspect.java:1)
	at com.box.l10n.mojito.service.pollabletask.pollableaspect.createpollablewrapper(pollableaspect.java:79)
	at com.box.l10n.mojito.service.drop.dropservice.importfile(dropservice.java:301)
	at com.box.l10n.mojito.service.drop.dropservice.importdrop_aroundbody6(dropservice.java:256)
	at com.box.l10n.mojito.service.drop.dropservice$ajcclosure7.run(dropservice.java:1)
	at org.aspectj.runtime.reflect.joinpointimpl.proceed(joinpointimpl.java:270)
	at com.box.l10n.mojito.service.pollabletask.pollablecallable.call(pollablecallable.java:50)
	at java.base/java.util.concurrent.futuretask.run(futuretask.java:317)
	at org.springframework.security.concurrent.delegatingsecuritycontextrunnable.run(delegatingsecuritycontextrunnable.java:94)
	at java.base/java.util.concurrent.executors$runnableadapter.call(executors.java:572)
	at java.base/java.util.concurrent.futuretask.run(futuretask.java:317)
	at java.base/java.util.concurrent.threadpoolexecutor.runworker(threadpoolexecutor.java:1144)
	at java.base/java.util.concurrent.threadpoolexecutor$worker.run(threadpoolexecutor.java:642)
	at java.base/java.lang.thread.run(thread.java:1583)
caused by: net.sf.okapi.common.exceptions.okapiioexception: com.ctc.wstx.exc.wstxeofexception: unexpected eof; was expecting a close tag for element <note>
 at [row,col,system-id]: [128,206,"/some/file/path/to/be/read/from/db"]
	at net.sf.okapi.filters.xliff.xlifffilter.processnote(xlifffilter.java:3671)
	at net.sf.okapi.filters.xliff.xlifffilter.processtransunit(xlifffilter.java:1688)
	at net.sf.okapi.filters.xliff.xlifffilter.read(xlifffilter.java:643)
	at net.sf.okapi.filters.xliff.xlifffilter.next(xlifffilter.java:360)
	at net.sf.okapi.steps.common.rawdocumenttofiltereventsstep.handleevent(rawdocumenttofiltereventsstep.java:166)
	at net.sf.okapi.common.pipeline.pipeline.execute(pipeline.java:117)
	at net.sf.okapi.common.pipeline.pipeline.process(pipeline.java:227)
	at net.sf.okapi.common.pipeline.pipeline.process(pipeline.java:199)
	at net.sf.okapi.common.pipelinedriver.pipelinedriver.processbatch(pipelinedriver.java:182)
	at com.box.l10n.mojito.service.tm.tmservice.updatetmwithxliff(tmservice.java:921)
	at com.box.l10n.mojito.service.tm.tmservice.updatetmwithtranslationkitxliff(tmservice.java:840)
	at com.box.l10n.mojito.service.drop.dropservice.updatetmwithlocalizedxliff_aroundbody10(dropservice.java:338)
	... 28 common frames omitted
caused by: com.ctc.wstx.exc.wstxeofexception: unexpected eof; was expecting a close tag for element <note>
 at [row,col,system-id]: [128,206,"/some/file/path/to/be/read/from/db"]
	at com.ctc.wstx.sr.streamscanner.throwunexpectedeof(streamscanner.java:701)
	at com.ctc.wstx.sr.basicstreamreader.throwunexpectedeof(basicstreamreader.java:5612)
	at com.ctc.wstx.sr.basicstreamreader.nextfromtree(basicstreamreader.java:2811)
	at com.ctc.wstx.sr.basicstreamreader.next(basicstreamreader.java:1122)
	at net.sf.okapi.filters.xliff.xlifffilter.processnote(xlifffilter.java:3642)
	... 39 common frames omitted

2026-01-29t07:42:26.610-08:00 debug 97 --- [ pollabletask-5] c.b.l.m.boxsdk.boxapiconnectionprovider  : getting box api connection
2026-01-29t07:42:27.110-08:00 debug 97 --- [ pollabletask-5] c.b.l.m.s.p.pollabletaskexceptionutils   : error happened during task execution

com.box.l10n.mojito.service.drop.importdropexception: number of files not imported: 1, check sub task for more information
	at com.box.l10n.mojito.service.drop.dropservice.importdrop_aroundbody6(dropservice.java:271)
	at com.box.l10n.mojito.service.drop.dropservice$ajcclosure7.run(dropservice.java:1)
	at org.aspectj.runtime.reflect.joinpointimpl.proceed(joinpointimpl.java:270)
	at com.box.l10n.mojito.service.pollabletask.pollablecallable.call(pollablecallable.java:50)
	at java.base/java.util.concurrent.futuretask.run(futuretask.java:317)
	at org.springframework.security.concurrent.delegatingsecuritycontextrunnable.run(delegatingsecuritycontextrunnable.java:94)
	at java.base/java.util.concurrent.executors$runnableadapter.call(executors.java:572)
	at java.base/java.util.concurrent.futuretask.run(futuretask.java:317)
	at java.base/java.util.concurrent.threadpoolexecutor.runworker(threadpoolexecutor.java:1144)
	at java.base/java.util.concurrent.threadpoolexecutor$worker.run(threadpoolexecutor.java:642)
	at java.base/java.lang.thread.run(thread.java:1583)
```

</details>

---

Note that the logged error position [128,206] was exacly 8192 characters. Additionally, this happens regardless of selected DropExporter (i.e. this is not caused by Box SDK failing to provide document content).

The root cause was one step in the Okapi pipeline (`IntegrityCheckStep`) advanding the underlying document stream to the end, while another step (`RawDocumentToFilterEventsStep` with `XLIFFFIlter`) was in the middle of parsing it. This issue was introduced in #731 which changed the method to retrieve content of the whole document from `RawDocument#getCharSequence` to `RawDocument#getReader`. According to the findings described in https://github.com/box/mojito/pull/1049#issuecomment-3854615251, it seems like it's only allowed to access the reader/stream within `BasePipelineStep#handleRawDocument`, not later (in any Filter Events handlers).

The fix is tested through new `DropServiceTest#forTranslationLargeXliffFile`.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/assetintegritychecker/integritychecker/DocumentIntegrityCheckStep.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/assetintegritychecker/integritychecker/DocumentIntegrityCheckStep.java
@@ -1,0 +1,63 @@
+package com.box.l10n.mojito.service.assetintegritychecker.integritychecker;
+
+import com.google.common.io.CharStreams;
+import java.io.IOException;
+import java.io.Reader;
+import java.util.List;
+import net.sf.okapi.common.Event;
+import net.sf.okapi.common.pipeline.BasePipelineStep;
+import net.sf.okapi.common.resource.RawDocument;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Configurable;
+
+/**
+ * @author aloison
+ */
+@Configurable
+public class DocumentIntegrityCheckStep extends BasePipelineStep {
+
+  /** Logger */
+  static Logger logger = LoggerFactory.getLogger(DocumentIntegrityCheckStep.class);
+
+  @Autowired IntegrityCheckerFactory integrityCheckerFactory;
+
+  @Override
+  public String getName() {
+    return "Document Integrity Check";
+  }
+
+  @Override
+  public String getDescription() {
+    return "Runs document-level integrity checks to make sure a document is valid."
+        + " When an issue is detected, it will throw an IntegrityCheckException."
+        + " Expects: raw document. Sends back: raw document.";
+  }
+
+  @Override
+  protected Event handleRawDocument(Event event) {
+    logger.debug("Check integrity of document");
+    RawDocument document = event.getRawDocument();
+
+    String documentContent;
+    try {
+      Reader reader = document.getReader();
+      documentContent = CharStreams.toString(reader);
+      // CharStreams#toString does not close the readable implicitly
+      reader.close();
+    } catch (IOException e) {
+      logger.error("Error reading document content", e);
+      throw new RuntimeException("Error reading document content", e);
+    }
+
+    // TODO(P1): do not hardcode the type here
+    List<DocumentIntegrityChecker> documentIntegrityCheckers =
+        integrityCheckerFactory.getDocumentCheckers("xliff");
+    for (DocumentIntegrityChecker checker : documentIntegrityCheckers) {
+      checker.check(documentContent);
+    }
+
+    return super.handleRawDocument(event);
+  }
+}
--- a/webapp/src/main/java/com/box/l10n/mojito/service/assetintegritychecker/integritychecker/TextUnitIntegrityCheckStep.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/assetintegritychecker/integritychecker/TextUnitIntegrityCheckStep.java
@@ -4,10 +4,7 @@ import com.box.l10n.mojito.entity.Asset;
 import com.box.l10n.mojito.entity.TMTextUnit;
 import com.box.l10n.mojito.entity.TMTextUnitVariantComment;
 import com.box.l10n.mojito.service.tm.TMTextUnitRepository;
-import com.google.common.io.CharStreams;
-import java.io.IOException;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import net.sf.okapi.common.Event;
@@ -16,7 +13,6 @@ import net.sf.okapi.common.pipeline.BasePipelineStep;
 import net.sf.okapi.common.pipeline.annotations.StepParameterMapping;
 import net.sf.okapi.common.pipeline.annotations.StepParameterType;
 import net.sf.okapi.common.resource.ITextUnit;
-import net.sf.okapi.common.resource.RawDocument;
 import net.sf.okapi.common.resource.TextContainer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -27,10 +23,10 @@ import org.springframework.beans.factory.annotation.Configurable;
  * @author aloison
  */
 @Configurable
-public class IntegrityCheckStep extends BasePipelineStep {
+public class TextUnitIntegrityCheckStep extends BasePipelineStep {
 
   /** Logger */
-  static Logger logger = LoggerFactory.getLogger(IntegrityCheckStep.class);
+  static Logger logger = LoggerFactory.getLogger(TextUnitIntegrityCheckStep.class);
 
   @Autowired TMTextUnitRepository tmTextUnitRepository;
 
@@ -39,7 +35,6 @@ public class IntegrityCheckStep extends BasePipelineStep {
   Map<Long, Set<TextUnitIntegrityChecker>> textUnitIntegrityCheckerMap = new HashMap<>();
 
   private LocaleId targetLocale;
-  private RawDocument rawDocument;
 
   @SuppressWarnings("deprecation")
   @StepParameterMapping(parameterType = StepParameterType.TARGET_LOCALE)
@@ -47,42 +42,15 @@ public class IntegrityCheckStep extends BasePipelineStep {
     this.targetLocale = targetLocale;
   }
 
-  @StepParameterMapping(parameterType = StepParameterType.INPUT_RAWDOC)
-  public void setInputDocument(RawDocument rawDocument) {
-    this.rawDocument = rawDocument;
-  }
-
   @Override
   public String getName() {
-    return "Integrity Check";
+    return "Text Unit Integrity Check";
   }
 
   @Override
   public String getDescription() {
     return "Updates the TM with the extracted new/changed variants."
-        + " Expects: raw document. Sends back: original events.";
-  }
-
-  @Override
-  protected Event handleStartDocument(Event event) {
-    logger.debug("Check integrity of document");
-
-    String documentContent = null;
-    try {
-      documentContent = CharStreams.toString(rawDocument.getReader());
-    } catch (IOException e) {
-      logger.error("Error reading document content", e);
-      throw new RuntimeException("Error reading document content", e);
-    }
-
-    // TODO(P1): do not hardcode the type here
-    List<DocumentIntegrityChecker> documentIntegrityCheckers =
-        integrityCheckerFactory.getDocumentCheckers("xliff");
-    for (DocumentIntegrityChecker checker : documentIntegrityCheckers) {
-      checker.check(documentContent);
-    }
-
-    return super.handleStartDocument(event);
+        + " Expects: Text unit events. Sends back: original events.";
   }
 
   @Override
--- a/webapp/src/main/java/com/box/l10n/mojito/service/tm/TMService.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/tm/TMService.java
@@ -47,7 +47,8 @@ import com.box.l10n.mojito.retry.DataIntegrityViolationExceptionRetryTemplate;
 import com.box.l10n.mojito.security.AuditorAwareImpl;
 import com.box.l10n.mojito.service.WordCountService;
 import com.box.l10n.mojito.service.asset.AssetRepository;
-import com.box.l10n.mojito.service.assetintegritychecker.integritychecker.IntegrityCheckStep;
+import com.box.l10n.mojito.service.assetintegritychecker.integritychecker.DocumentIntegrityCheckStep;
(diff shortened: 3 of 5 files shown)


wadimw:
Notes on the go: During direct isolated testing of the flow of `TMService#updateTMWithXliff`, removal of the `IntegrityCheckStep` allows the test to handle large documents correctly. That step breaks things, because it calls `CharStreams.toString(rawDocument#getReader)`, which somehow sideeffects the underlying stream that is created on the fly in `RawDocument` ctor. This was introduced in https://github.com/box/mojito/pull/731 and probably wasn't ever noticed, because as per Guava documentation, `CharStreams#toString` does not close the stream when it's done with the reader. If we revert to using CharSequence-based document, then it works fine for large files - but then it breaks the encoding-related tests. In theory we could try to use `RawDocument#setEncoding` to change it, but in practice it does not work, because this method no-ops for documents created from CharSequence and keeps the default `UTF-16`, i.e. it's impossible to force encoding to `UTF-8` that we want. So, a hacky way to fix this is to update the encoding through reflection - subsequent processing produces `UTF-8` output streams. This might be fine, since we're already using reflection to force set the URI on the RawDocument anyway - but this actually opposes Okapi RawDocument contract, which says you can have only ONE of: a charsequence, a stream or an URI and possibly introduces some unexpected runtime behaviour, so we should probably avoid introducing _more_ hacky reflection stuff.

The funny part is that this solution (i.e. reverting to CharSequence-backed document, but then force changing the encoding) actually works https://github.com/box/mojito/actions/runs/21634652420/job/62356567792

common/src/main/java/com/box/l10n/mojito/okapi/RawDocument.java
```diff
@@ -24,18 +23,19 @@ public RawDocument(CharSequence inputCharSequence, LocaleId sourceLocale) {
   }
 
   public RawDocument(CharSequence inputCharSequence, LocaleId sourceLocale, LocaleId targetLocale) {
-    super(
-        new ByteArrayInputStream(inputCharSequence.toString().getBytes(StandardCharsets.UTF_8)),
```
ehoogerbeets, line 28: Did you look to see if net.sf.okapi.common.resource.RawDocument (the parent class) properly closes the stream in its first parameter?

wadimw:
Possibly the same issue as https://github.com/box/mojito/issues/1021

wadimw:
New findings:

Seems like we can get rid of the `RawDocument` URI injection through reflection workaround, if we instead manually set the document name on the `START_DOCUMENT` event to avoid NPE from `QualityChecker`. This can be done by overriding `QualityCheckStep#handleStartDocument`, because this event actually exposes a public method `StartDocument#setName` - so it's a bit less hacky then reflection. Additionally, this means that this workaroud is more contained, since it's implemented on that particular class which requires it. I previously thought this URI change might even be the root cause of this issue, but it seems that it's not the case - more on that below. This however means, that the URI injection removal is merely a refactor, and so it's outside of the scope of this issue. We may still want to include it here, or not. (Note: after discussing with @ehoogerbeets we've decided to move the URI refactor into a separate PR #1052)

Now, onto the actual issue. The way stream-based `RawDocument` works is that it provides the same cached InputStream instance every time `RawDocument#getStream` or `RawDocumen#getReader` is called - it's just that these methods implicitly reopen that stream (if it has been closed elsewhere) and reset it to the beginning. So, if we retrieve this stream anywhere, we must be aware that this introduces side effects which may impact any other piece of code that also refers to it. Given that the `RawDocumentToFilterEventsStep` with `XLIFFFilter` uses that stream within an XML parser of the `XLIFFFilter`, that means that if that parser requests the stream before our `IntegrityCheckStep#handleStartDocument` does, our request (and further reading using `CharStreams.toString`), puts the `XLIFFFilter` in a weird state (because it might have been in the middle of the document, and we've suddenly advanced the stream to the end). So basically, it looks like we shouldn't have two steps in a single pipeline that actually use the stream at the same time.

Note that this behaviour didn't surface earlier, because `CharStreams#toString` does not implicitly close the stream - so there was no obvious way to notice that this stream has been tampered with. Additionally, it seems that the XML parser of `XLIFFFilter` caches some of the read data in memory, so this may explain why _nothing_ wrong happened for small files. ~I also now realize that this must mean that the `IntegrityCheckStep` never really got the WHOLE document, since the `XLIFFFilter` would probably read ahead beforehand~ <- not true, because `RawDocument#getReader` implicitly resets the stream to the beginning.

To confirm these findings, I first tried adding `reader.close()` and then another `rawDocument.getReader()` immediately after closing to verify the side effects actually impact `XLIFFFilter` behaviour. The findings match - closing the stream causes `XLIFFFilter` parser to throw when it tries to read more data than what it has cached beforehand, while re-requesting the reader causes the `XLIFFFilter` parser to throw due to unexpected character `<` (i.e. the underlying shared stream has been reset to the beginning of the file). The latter also confirms behavior that we've run into originally, i.e. the unclosed reader would report EOF (due to the read ahead) even though `XLIFFFilter` parser buffered only 8k of the input document content.

Using a debugger I discovered that the `XLIFFFilter` actually calls `RawDocument#getStream` _multiple_ times, so it seems that the reopen-reset behaviour is at least somewhat expected in the whole design of Okapi pipeline. More importantly though, I noticed that this is **only** called during `RawDocumentToFilterEventsStep#handleRawDocument` as opposed to ours `IntegrityCheckStep#handleStartDocument`. This suggests that it's actually fine to retrieve (and implicitly reopen-reset) the document stream multiple times - but it can only happen in the beginning of the pipeline, before `RawDocumentToFilterEventsStep` with `XLIFFFilter` actually starts reading step by step and emitting text unit events.

The idea then is to move the `CharStreams#toString` call to an eariler moment within the pipeline, where apparently it is fine to mess with the Stream. So, I moved this code to `IntegrityCheckStep#handleRawDocument`. This caused it to never run - because it turned out that the`RawDocumentToFilterEventsStep` would suppress the `RAW_DOCUMENT` event - which I assume is exacly why this code was initially placed in the `START_DOCUMENT` handler. Note again, that this placement was not a problem initially, because before #731 the `RawDocument` was based on `CharSequence`, so the integrity check utilized `RawDocument#getCharSequence` rather than the reader - thus it didn't touch the stream at all. To get the `IntegrityCheckStep#handleRawDocument` to execute, I had to move `IntegrityCheckStep` _before_ the `RawDocumentToFilterEventsStep`. But then, I ran into the other side of the same issue, i.e. that now the checks implemented in `IntegrityCheckStep#handleTextUnit` didn't run, because the `TEXT_UNIT` events were not produced yet.

So, eventually I decided to split the integrity checks into two pipeline steps - one for document-level checks that read the whole file content from its stream while it's still allowed (i.e. before `RawDocumentToFilterEventsStep`), other for text-unit-level cheks (after `RawDocumentToFilterEventsStep`, because it relies on the events emmited by it).

EDIT: IT WOOOOOORKS WOOHOOOOOOOO https://github.com/box/mojito/actions/runs/21718444824/job/62640989530?pr=1049

ehoogerbeets (approved):
So simple!

PR #1051, opened 2026-02-05
Update documentation, including for some new features

ehoogerbeets:
- Fix many typos
- Add AI documentation and documentation of other new features
- Added information about roles and who can do what


zmoskala:
@ehoogerbeets please update again now that we merged to master

ehoogerbeets:
Rebased and the documentation updated now that the upstream-patched code is in master. Ready for review now.

zmoskala (commented):
Copy/grammar pass only. These are all wording nits on lines this PR already touches (or their immediate context) — no behavior or accuracy concerns in this batch.

docs/_docs/guides/002-install_springboot3.md
```diff
@@ -159,7 +159,7 @@ max_allowed_packet = 256M
 If using a older version of MySQL, there is a [known issue](https://github.com/box/mojito/issues/120) when creating the schema. One workaround is to use `utf8`
 instead `utf8mb4` but it has its limitation in term of character support.
 
-We recommand to run both MySQL and the Java service using `UTC` timezone (or a least make sure they both the same timezone). To set
+We recommend running both MySQL and the Java service using `UTC` timezone (or a least make sure they both the same timezone). To set
```
zmoskala: Leftover grammar in the sentence being fixed here: "or a least make sure they both the same timezone" → "or at least make sure they both use the same timezone".

Same paragraph, in the context just above (line 159): "If using a older version" → "an older version"; "One workaround is to use `utf8` instead `utf8mb4`" → "instead of `utf8mb4`"; "it has its limitation in term of character support" → "it has its limitations in terms of character support".

Further down the file, outside the diff (line 173): "it is also common to use the console to enter credential" → "to enter credentials".

docs/_docs/guides/002-install_springboot3.md
```diff
@@ -37,7 +37,7 @@ One simple solution is to add an `application.properties` next to the `jar`. To
 The server provides an entry point to fetch a `bash` script that downloads the latest CLI from the server and create a bash 
 wrapper to easily run the CLI.
 
-It can be called with a one liner to make the bash command available rigth away in the current console. Replace 
+It can be called with a one liner to make the bash command available right away in the current console. Replace
```
zmoskala: Two copy nits around this fix: "a one liner" → "a one-liner", and on line 37 the subject/verb disagree — "a `bash` script that downloads the latest CLI from the server and create a bash wrapper" → "and creates a bash wrapper".

docs/_docs/refs/configurations_springboot2.md
```diff
@@ -96,7 +96,7 @@ Depending on the file size that will be processed, it might be required to incre
 If using a older version of MySQL, there is a [known issue](https://github.com/box/mojito/issues/120) when creating the schema. One workaround is to use `utf8`
 instead `utf8mb4` but it has its limitation in term of character support.
 
-We recommand to run both MySQL and the Java service using `UTC` timezone (or a least make sure they both the same timezone). To set
+We recommend running both MySQL and the Java service using `UTC` timezone (or a least make sure they both the same timezone). To set
```
zmoskala: Same leftover as in the Spring Boot 3 install guide: "or a least make sure they both the same timezone" → "or at least make sure they both use the same timezone".

Context just above (lines 96-97): "If using a older version" → "an older version"; "instead `utf8mb4`" → "instead of `utf8mb4`"; "limitation in term of character support" → "limitations in terms of character support".

docs/_docs/guides/008-managing-locales.md
```diff
@@ -23,15 +23,14 @@ We use `mojito-cli` to configure locales in a repository.  Locales can be config
 
 By default, all locales configured in the repository are required to be fully translated.  These locales get automatically included in the translation requests.
 
-You can configure the locales to be partially translated.  English in United Kingdom (en-GB) is a good example because most of the strings do not need to be "translated".  Source strings can be used as-is in most cases and only some strings that are specific to English in United Kingdom need to be overriden.
+You can configure the locales to be partially translated.  English in United Kingdom (en-GB) is a good example because most of the strings do not need to be "translated".  Source strings can be used as-is in most cases and only some strings that are specific to English in United Kingdom need to be overriden. In this sense, en-GB is an inherited locale. That is, it inherits
```
zmoskala: Typo carried over on this line: "need to be overriden" → "overridden".

docs/_docs/guides/008-managing-locales.md
```diff
@@ -44,7 +43,8 @@ mojito repo-update -n MyRepo -l "(fr-CH)->fr-FR" fr-FR ja-JP zh-CN zh-TW
 The above example makes French in Switzerland (fr-CH) the child locale of French in France (fr-FR).
 
 
-Note that the partially translated locales are displayed in grey color in locales list.
+Note that the partially translated locales are displayed in grey color in locales list. The parent locale for partially
```
zmoskala: Wording: "displayed in grey color in locales list" → "displayed in grey in the locales list".

docs/_docs/guides/006-workbench.md
```diff
@@ -10,6 +10,8 @@ Workbench in {{ site.mojito_green }} is a place where you can search and edit al
 
 Use Workbench to fix bugs, communicate context to translation teams and manage global terminology changes. Translators and reviewers can also work directly in Workbench.
 
+> **Who can use the Workbench:** Everyone can search and view. **Translators**, **Project Managers**, and **Admins** can edit translations (Translators only for their assigned locales). Export, Import, and Share are available to Translators, PMs, and Admins. See [User Roles & Permissions]({{ site.url }}/docs/guides/user-roles-overview/) for details.
+
 ##  Searching
```
zmoskala: Double space in the heading: `##  Searching` → `## Searching`. Same on `##  Managing text units` further down. The new headings this PR adds (`## Exporting search results`, `## Importing translations`, `## Sharing searches`) use a single space, so the file is now inconsistent — worth normalizing while you are here.

docs/_docs/guides/authentication_springboot3.md
```diff
@@ -17,7 +17,7 @@ Change or add an authentication mechanisms by updating the configuration. Eg. to
 
     l10n.security.authenticationType=DATABASE,OAUTH2
         
-You can chosse to either show the {{ site.mojito_green }}'s login page or to automatically redirect to another page.
+You can choose to either show the {{ site.mojito_green }}'s login page or to automatically redirect to another page.
 Eg. to redirect to Github OAuth when the not authenticated
```
zmoskala: Grammar in this line: "Eg. to redirect to Github OAuth when the not authenticated" → "E.g. to redirect to GitHub OAuth when the user is not authenticated." (also `Github` → `GitHub`).

A few more in the same file, outside the diff: line 12 "This enables to have a dual authentication scheme" → "This enables a dual authentication scheme"; line 14 the parenthesis opened at "(potentially `OAuth` for regular users…" is never closed; line 16 "Change or add an authentication mechanisms" → "an authentication mechanism"; line 29 "{{ site.mojito_green }} support `OAuth 2`" → "supports".

docs/_docs/guides/014-ai-translate.md
```diff
@@ -0,0 +1,93 @@
...
+| Option | Description | Default |
+|--------|-------------|---------|
+| **Source texts per locale** | Maximum number of source strings to translate per locale | 100 |
+| **Text unit IDs** | Optional comma/whitespace-separated list of TM text unit IDs to limit the scope | — |
+| **Model override** | AI model identifier (e.g. `gpt-4.1`) | gpt-4.1 |
+| **Prompt suffix** | Optional text appended to the base prompt | — |
+| **Related strings** | Extra context sent to the AI. See [Related strings](#related-strings) below. | NONE |
+| **Translate type** | `TARGET_ONLY_NEW` = only untranslated; `TARGET_ONLY` = overwrite existing; `WITH_REVIEW` = translate all and mark for review | TARGET_ONLY_NEW |
+| **Status filter** | `FOR_TRANSLATION` = only strings needing translation; `ALL` = every string | FOR_TRANSLATION |
+| **Import status** | Status applied to imported translations: `REVIEW_NEEDED`, `ACCEPTED`, or `TRANSLATION_NEEDED` | REVIEW_NEEDED |
```
zmoskala, line 43: The three values here match the current AI Translate dropdown, so the docs are accurate to the UI. One heads-up for readers: ACCEPTED is the Workbench label for a finished translation, but the import path does TMTextUnitVariant.Status.valueOf(...), and that enum is only TRANSLATION_NEEDED, REVIEW_NEEDED, APPROVED. There is no mapping from ACCEPTED → APPROVED, so choosing ACCEPTED fails the import. Not something this docs PR introduced — worth a short warning here, and a product follow-up so the form/CLI send APPROVED. Default REVIEW_NEEDED is fine.

zmoskala: For reference, sharing a step by step explanation when this can currently fail:
### What happens if you select ACCEPTED
Assume a repo that actually has strings to translate, Import status = **ACCEPTED**, click **Start translation**. The string `"ACCEPTED"` is never rewritten to `APPROVED`.
1. **Form submit** (`RepositoryAiTranslatePage.submitJob` → `buildRequestPayload`)  
   The dropdown value is copied into the JSON as-is: `"importStatus": "ACCEPTED"`.
2. **HTTP POST** (`RepositoryAiTranslateClient.translateRepository`)  
   POST `/api/proto-ai-translate`. The browser is only starting a job, not applying status yet.
3. **REST controller** (`AiTranslateWS.aiTranslate`)  
   `importStatus` is a plain `String`. Spring does not check it against the enum. It is passed into `AiTranslateInput` unchanged.
4. **Schedule background work** (`AiTranslateService.aiTranslateAsync`)  
   This does **not** call `valueOf`. It hands the input to Quartz (`AiTranslateJob`) and returns a pollable task id. HTTP is **200**. The page starts waiting on that task. From the user’s point of view, Start succeeded.
   Quartz is just Mojito’s “run this later” engine, so a long translation does not block the HTTP request.
5. **Quartz runs the job** (`AiTranslateJob.call`)  
   Calls `aiTranslateService.aiTranslate(...)`. The web form sends `useBatch: false`, so this goes to `aiTranslateNoBatch`.
6. **Per locale** (`aiTranslateNoBatch`)  
   Strings are loaded for the locale. If the list is **empty**, it `continue`s and never parses import status.  
   If there is **at least one** string, it hits:
   `Status importStatus = Status.valueOf(aiTranslateInput.importStatus());`
   (`AiTranslateService.aiTranslateNoBatch`, the `Status.valueOf` line)
   `Status` is `TMTextUnitVariant.Status`: only `TRANSLATION_NEEDED`, `REVIEW_NEEDED`, `APPROVED`.  
   `valueOf("ACCEPTED")` looks for a constant named `ACCEPTED`. There isn’t one. Java throws `IllegalArgumentException: No enum constant ...Status.ACCEPTED`. Nothing in this method catches it.
   This is **before** the OpenAI loop, so it typically fails without importing and without sending those strings to the model.
7. **Job marked failed** (`QuartzPollableJob`)  
   The throw is caught, the pollable task is finished with an error message.
8. **UI** (`RepositoryAiTranslatePage` wait loop)  
   Polling sees `pollableTask.errorMessage` and shows it on the page. Translations were not saved.
If every locale had nothing to translate, step 6 never calls `valueOf` and the job can finish without applying any status. The failure is specifically: **ACCEPTED + at least one string to translate**.

In general: My suggestion is to add a one-line warning on this row that `ACCEPTED` currently fails, and a product follow-up so the form/CLI send `APPROVED`.

docs/_docs/guides/014-ai-translate.md
```diff
@@ -0,0 +1,93 @@
...
+## Options
+
+| Option | Description | Default |
+|--------|-------------|---------|
+| **Source texts per locale** | Maximum number of source strings to translate per locale | 100 |
+| **Text unit IDs** | Optional comma/whitespace-separated list of TM text unit IDs to limit the scope | — |
+| **Model override** | AI model identifier (e.g. `gpt-4.1`) | gpt-4.1 |
+| **Prompt suffix** | Optional text appended to the base prompt | — |
+| **Related strings** | Extra context sent to the AI. See [Related strings](#related-strings) below. | NONE |
+| **Translate type** | `TARGET_ONLY_NEW` = only untranslated; `TARGET_ONLY` = overwrite existing; `WITH_REVIEW` = translate all and mark for review | TARGET_ONLY_NEW |
```
zmoskala: Can you please verify if this is correct? It seems to me there is an issue with these definitions:

**What the table says today**
- `TARGET_ONLY_NEW` = only untranslated
- `TARGET_ONLY` = overwrite existing
- `WITH_REVIEW` = translate all and mark for review

**What the code actually uses**
- **Which strings are sent** is **Status filter** (`FOR_TRANSLATION` vs `ALL`). `aiTranslateNoBatch` does `textUnitSearcherParameters.setStatusFilter(statusFilter)`. `translateType` is not used there. You can combine any type with either filter.
- **Status written on import** is **Import status**. After the model returns, the server does `textUnitDTO.setStatus(importStatus)`. `WITH_REVIEW` does not itself set `REVIEW_NEEDED`. Extra review JSON (explanation, confidence, review flag) is stored as a comment.

**What Translate type actually is**  
`AiTranslateType.fromString(translateType)` (`AiTranslateType.java`) picks the prompt and the JSON the model must return.

- `TARGET_ONLY_NEW` and `TARGET_ONLY` are the same kind of “just translate” prompt. They are **not** new-vs-overwrite. The difference is output shape: `TARGET_ONLY_NEW` expects a list `{ targets: [{ tmTextUnitId, target }, …] }` (lookup by id; fits the web path sending several strings per request). `TARGET_ONLY` expects a single `{ content: "…" }` (id is ignored).
- `WITH_REVIEW` is a richer prompt/output, not “translate everything.”

The names are easy to misread (`TARGET_ONLY_NEW` sounds like a scope filter), which is likely how this row got written. If you agree with this finding, please describe them as prompt/output modes and leave scope and saved status to the Status filter and Import status rows. The later note that CLI defaults to `WITH_REVIEW` and the web form to `TARGET_ONLY_NEW` is still right — that is a different default prompt, not different overwrite behavior.

docs/_docs/guides/000-user-roles-overview.md
```diff
@@ -0,0 +1,93 @@
...
+permalink: /docs/guides/user-roles-overview/
+---
+
+{{ site.mojito_green }} uses roles to control what you can see and do. This guide explains each role and what applies to you.
+
+## Roles at a glance
+
+| Role | Who you are | What you can do |
+|------|-------------|-----------------|
+| **User** | Basic access | View repositories, workbench, branches, screenshots. Search and browse. No edits. |
```
zmoskala: Please fix who can export / import / share — this new “User = browse only” story does not match the Workbench or the API.

The toolbar always renders Share, Export, and Import (`Workbench.js`). `AuthorityService` does not gate those actions (it only gates edit-translations, project-requests, user-management, edit-screenshots).

Server (`WebSecurityConfig`, first match wins):
- Export uses search: `POST /api/textunits/search` → any authenticated user
- Share saves the search: `POST /api/clobstorage` → any authenticated user
- Import writes translations: other `/api/textunits/**` → TRANSLATOR, PM, ADMIN only (`// USERs are not allowed to change translations`)

So a User **can** export and share. They **see** Import but the save is 403.

Maybe you can mark export and share as available to User, and say import is visible to everyone but only Translator/PM/Admin can succeed? Same wording can be used on the feature table (line 78) and the Workbench “Who can export/import/share” callouts.

docs/_docs/guides/000-user-roles-overview.md
```diff
@@ -0,0 +1,93 @@
...
+---
+
+## For Project Managers
+
+**You can do everything Translators can, plus:**
+- Create translation and review [project requests]({{ site.url }}/docs/guides/project-request/)
+- Import and re-import completed projects
+- Cancel project requests
+- Add and manage screenshots in the [legacy screenshot dashboard]({{ site.url }}/docs/guides/branching/#collecting-screenshots)
+- Manage users (add, update roles) via **Settings → User Management**
```
zmoskala: Maybe we can tighten this to indicate that PMs can manage users, but they cannot create or edit Admins?

`UserService.checkPermissionsForRole` throws `AccessDeniedException`: “PMs are not allowed to edit / create ADMINs” if the current user is a PM and the target role is `ROLE_ADMIN`. That runs on create and on save, including when the existing user is already an Admin (so a PM cannot edit an Admin’s name either). Translators/Users cannot manage users at all, which you already have right.

docs/_docs/guides/008-managing-locales.md
zmoskala: I'm afraid this sentence is easy to misread. Maybe we can make the parent language explicit?

**What `(en-GB)` does**  
Parentheses mean: this locale does not need a full translation. Missing strings are filled in from a **parent** locale.

**What this page says**  
It says that parent is en-US, and that `(en-GB)` is the same as writing `(en-GB)->en-US`.

**What the code does**  
If you do **not** write `->…`, Mojito does not assume en-US. It uses this repository’s **source locale** (`RepositoryService.addRepositoryLocale`: `repository.getSourceLocale()`). In these docs that source is usually `en`, not `en-US`.

So:
- `(en-GB)` → parent = whatever this repo’s source is (often `en`)
- `(en-GB)->en-US` → parent = `en-US` (only works if `en-US` is already a locale in that repo)

If someone copies `(en-GB)->en-US` and the repo has no `en-US`, the command fails (“parent locale doesn’t exist”).

Please see if this would be better in your opinion:
lines 26-27: You can configure the locales to be partially translated. English in United Kingdom (en-GB) is a good example because most of the strings do not need to be "translated". Source strings can be used as-is in most cases and only some strings that are specific to English in United Kingdom need to be overridden. In this sense, en-GB is an inherited locale: missing strings are filled in from the repository source locale (often en).

lines 46-47: Note that the partially translated locales are displayed in grey in the locales list. If you omit ->…, the parent is the repository source locale. For a repo whose source is en, (en-GB) is equivalent to (en-GB)->en. Use -> only when the parent is a different locale that is already in the repo, as in (fr-CH)->fr-FR above. Do not write (en-GB)->en-US unless en-US is actually a locale in that repository.

docs/_docs/guides/authentication_springboot3.md
```diff
@@ -17,7 +17,7 @@ Change or add an authentication mechanisms by updating the configuration. Eg. to
...
     
     l10n.security.unauth-redirect-to==/login/oauth2/authorization/github
```
zmoskala: Typo in this example: `l10n.security.unauth-redirect-to==/login/...` has two `=` signs. 

The in-repo example is a single `=` (`application.properties`: `l10n.security.unauth-redirect-to=/login/oauth2/authorization/github`). Please drop the extra `=`. (Pre-existing in this file; you are already editing it.)

docs/_docs/guides/manage-users.md
```diff
@@ -5,17 +5,19 @@ categories: guides
...
 
+> **Who can manage users:** **Project Managers** and **Admins** can access User Management (Settings → User Management). Translators and Users cannot. See [User Roles & Permissions]({{ site.url }}/docs/guides/user-roles-overview/).
+
 The default authentication used by {{ site.mojito_green }} relies on the database
 
     l10n.security.authenticationType=DATABASE
     
 User information is stored in database and all the authentication process
-is handle by the server. Managing users is done 
+is handled by the server. Managing users is done
```
zmoskala: The new blurb points to Settings → User Management, then this sentence still says users are managed via the CLI. Can we say the web UI and CLI are both valid (CLI remains useful for bootstrap/password), so it is clear?

docs/_docs/guides/000-user-roles-overview.md
```diff
@@ -0,0 +1,93 @@
...
+- Configure [Box integration]({{ site.url }}/docs/guides/integrating-with-box/) for project requests via **Settings → Box Integration**
+
+**You cannot:**
+- Access [AI Translate]({{ site.url }}/docs/guides/ai-translate/) (Admin only)
```
zmoskala: UI-only Admin gating is accurate (Header.js, onEnterAiTranslate). The API is not Admin-only: WebSecurityConfig allows PM+ADMIN on remaining /api/** writes. If this page is about the product UI, please say so. If it is about enforcement, PMs can still call the AI Translate endpoint / repository-ai-translate CLI.

zmoskala (changes requested):

docs/_docs/guides/002-install_springboot3.md
ehoogerbeets: Done — this wording is updated on the branch.

docs/_docs/guides/002-install_springboot3.md
ehoogerbeets: Done — this wording is updated on the branch.

docs/_docs/refs/configurations_springboot2.md
ehoogerbeets: Done — this wording is updated on the branch.

docs/_docs/guides/008-managing-locales.md
ehoogerbeets: Done — this wording is updated on the branch.

docs/_docs/guides/008-managing-locales.md
ehoogerbeets: Done — this wording is updated on the branch.

docs/_docs/guides/006-workbench.md
ehoogerbeets: Done — this wording is updated on the branch.

docs/_docs/guides/authentication_springboot3.md
ehoogerbeets: Done — this wording is updated on the branch.

docs/_docs/guides/authentication_springboot3.md
ehoogerbeets: Done — this wording is updated on the branch.

docs/_docs/guides/000-user-roles-overview.md
ehoogerbeets: Done — PMs are documented as able to manage users, but not create or edit Admins.

docs/_docs/guides/manage-users.md
ehoogerbeets: Done — the manage-users guide now says the web UI and CLI are both valid, with CLI still useful for bootstrap and password changes.

docs/_docs/guides/000-user-roles-overview.md
ehoogerbeets: Done — the Admin-only restriction is described as UI gating; PMs can still use the AI Translate API and repository-ai-translate CLI.

docs/_docs/guides/014-ai-translate.md
ehoogerbeets: Done — Translate type is documented as prompt/output modes. Status filter owns which strings are sent; Import status owns the saved status. The CLI vs web default note now says that is a different default prompt, not overwrite behavior.

docs/_docs/guides/000-user-roles-overview.md
ehoogerbeets: Done — User can export and share. Import is visible to everyone, but only Translator/PM/Admin can save (User gets 403). Same wording on the roles table, the feature table, and the Workbench callouts.

docs/_docs/guides/008-managing-locales.md
ehoogerbeets: Done — `(en-GB)` now inherits from the repository source locale (often `en`), not `en-US`. The page says to use `->` only when the parent is already in the repo, and not to copy `(en-GB)->en-US` unless `en-US` is actually configured.

docs/_docs/guides/014-ai-translate.md
ehoogerbeets: Leaving the docs as-is for this PR.

`ACCEPTED` is the Workbench label for a finished translation; the stored enum is `APPROVED`. Workbench already maps accept → `APPROVED`, but AI Translate passes `ACCEPTED` straight into `Status.valueOf(...)`, so the job fails. That is a product bug, not a docs mismatch with the current UI/CLI options.

Tracked as LXP-2324: https://jira.inside-box.net/browse/LXP-2324

Once that lands (send or map `ACCEPTED` → `APPROVED`), this Import status row will match behaviour. No docs warning here so it does not go stale after the fix.

ehoogerbeets: Agreed on the failure path — thanks for the walkthrough.

This is a software bug (form/CLI send `ACCEPTED`; import only knows `APPROVED`). Docs will keep listing the current options so they match the UI; the behaviour will be fixed in LXP-2324: https://jira.inside-box.net/browse/LXP-2324

zmoskala: Sounds good, thanks!

zmoskala (approved):
Approved

ehoogerbeets:
The spotless check is fixed in a different PR

PR #1052, opened 2026-02-06
Refactor workaround for Okapi in-memory document handling

wadimw:
Split off from #1049

This PR removes the URI injection using reflection on RawDocument which is potentially fragile. Instead, it replaces it with an override on `QualityCheckStep` `START_DOCUMENT` event handler that injects the expected value directly into the `StartDocument` resource through its public method `StartDocument#setName`.

Additionally, it slightly refactors the empty `RawDocument` approach to make it more explicit that the _content_ of this document is not relevant - it's only used to "drive" the Okapi pipeline and provide Locale configuration.

---

Rationale for the URI refactor:

Force setting URI on a `RawDocument` backed by InputStream or a CharSequence breaks its contract as specified [here](https://gitlab.com/okapiframework/Okapi/-/blob/v1.45.0/okapi/core/src/main/java/net/sf/okapi/common/resource/RawDocument.java#L52-L53):

> The RawDocument object has one (and only one) of three input objects: a CharSequence, a URI, or an InputStream.

Even the private runtime of RawDocument itself depends on whether the URI field is set (e.g. [here](https://gitlab.com/okapiframework/Okapi/-/blob/v1.45.0/okapi/core/src/main/java/net/sf/okapi/common/resource/RawDocument.java#L324)), meaning this workaround might introduce unexpected behaviour and is potentially fragile.

After removal of the URI injection from Mojito code, the only thing that breaks seems to be `QualityCheckStep` - specifically, it throws a NullPointerException on the following [line](https://gitlab.com/okapiframework/Okapi/-/blob/v1.45.0/okapi/libraries/lib-verification/src/main/java/net/sf/okapi/lib/verification/QualityChecker.java#L114):

```java
@Override
public void processStartDocument (StartDocument sd,
	List<String> sigList)
{
	currentDocId = (new File(sd.getName())).toURI();
// ...
```

This can be avoided by setting an override on our existing `QualityCheckStep` subclass which will set this name if it's null. This way, the workaround is localized only to the class that actually needs it, rather than affecting all steps of Okapi pipelines that would rely on `RawDocument`.

### DIFF
--- a/common/src/main/java/com/box/l10n/mojito/okapi/RawDocument.java
+++ b/common/src/main/java/com/box/l10n/mojito/okapi/RawDocument.java
@@ -1,14 +1,8 @@
 package com.box.l10n.mojito.okapi;
 
 import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.lang.reflect.Field;
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
-import java.util.UUID;
 import net.sf.okapi.common.LocaleId;
-import org.springframework.util.ReflectionUtils;
 
 /**
  * When creating {@link RawDocument} from a string the URI is not set and setter is not available.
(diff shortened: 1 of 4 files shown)


wadimw:
TODO should probably also get rid of https://github.com/box/mojito/blob/f7a5979605c83bd84cce277d4fa1a1652ebb9391/common/src/main/java/com/box/l10n/mojito/okapi/steps/FilterEventsToInMemoryRawDocumentStep.java#L31-L32 in this PR

wadimw:
Closing for now, we can revisit in future

PR #1053, opened 2026-02-09
fix(po): escape backslash characters in SimpleEncoder and fix unescape roundtrip

jaknas:
## Problem

PO files use C-style string escaping where backslash must be escaped as `\\` to represent a literal backslash. Two issues existed:

1. **`SimpleEncoder` did not escape backslashes or tabs** — strings containing `\` (e.g. Windows paths like `C:\Users\test`) or tabs were written to PO files without proper escaping.

2. **`UnescapeUtils.unescape()` used sequential regex replacements** — this corrupted strings containing ambiguous sequences like `\\n` (escaped-backslash followed by literal `n`), which was decoded as a newline character instead of `\n`.

### Encoding (SimpleEncoder.java)
Added missing cases to `SimpleEncoder.encode()` for backslash (`\\`) and tab (`\t`), following the [GNU PO file format](https://www.gnu.org/software/gettext/manual/html_node/PO-File-Entries.html) which uses C-style string escaping. `POEncoder` extends `SimpleEncoder`, so this fix applies to PO file writing.

### Decoding (UnescapeUtils.java)
Rewrote `unescape()` to use a **single-pass** `Matcher.find()` approach with a switch statement, matching Okapi's own `POFilter.unescape()` pattern `\\[abfnrtv"'\\]` exactly. Handles all C-style escape sequences: `\\`, `\a`, `\b`, `\f`, `\n`, `\r`, `\t`, `\v`, `\"`, `'`.

The single-pass approach is critical because `\\n` (4 chars) must match `\\` first (→ `\`), leaving `n` as a literal — not `\n` (newline).

### PO Filter integration (POFilter.java)
- **`unescapeSource/Target`**: Made no-ops — Okapi's built-in single-pass unescape in `toAbstract()` already handles all C-style escape sequences in TextUnit source/target. Any additional unescaping would double-process and corrupt strings (e.g. `C:\Users\test` where `\t` becomes TAB).
- **`adaptTextUnitToCLDRForm`**: Apply `unescape()` to raw `msgID`/`msgIDPlural` so plural sources are fully decoded (previously only quotes were handled).
- **`renameTextUnitWithSourceAndContent`**: Unchanged — keeps using raw `msgID` for text unit names to preserve backward compatibility with existing TM data.

### No breaking changes
Text unit names are preserved as-is (raw PO escapes). This avoids orphaning existing translations in the TM. The fix is limited to correct encoding/decoding of source and target content.

## Test coverage

| Test class | # Tests | What it covers |
|---|---|---|
| `SimpleEncoderTest` | 21 | Encoding each char type (incl. tab), string combos, 7 roundtrip tests including edge cases (`\n`, `\r`, `\t`, `\"` as literal chars) |
| `UnescapeUtilsTest` | 16 | Individual helpers, single-pass edge cases (`\\n`→backslash+n not newline, `\\r`, `\\\"`, tab, empty, no-op, triple backslash) |
| `POFilterTest` | 13 | Encoder wiring, backslash/special char escaping, roundtrip through POEncoder + UnescapeUtils |
| `AssetExtractorTest` | 13 (6 new) | Integration: backslash, newline, quote, complex multi-escape, msgctxt with backslash, realistic user-facing message. Verifies name preserves raw escapes while source is properly decoded. |
| `TMServiceTest` | 1 new + 2 updated | New `testLocalizePoBackslashEscaping` (full roundtrip: extract → search → localize → import → re-localize); existing tests verify name/source/target consistency |

### Cross-filter validation
The `unescape()` method is also called by `AndroidFilter`, `JSFilter`, `RegexEscapeDoubleQuoteFilter`, and `MacStringsdictFilterKey`. All existing filter tests pass with the new behavior (41 tests: Android 16, MacStringsdict 1, MacStringsdictKey 6, MacStrings 18).

### DIFF
--- a/common/src/main/java/com/box/l10n/mojito/okapi/filters/POFilter.java
+++ b/common/src/main/java/com/box/l10n/mojito/okapi/filters/POFilter.java
@@ -167,19 +167,16 @@ public class POFilter extends net.sf.okapi.filters.po.POFilter {
   }
 
   void unescapeSource(TextUnit textUnit) {
-    String sourceString = textUnitUtils.getSourceAsString(textUnit);
-    String unescapedSourceString = unescapeUtils.replaceEscapedQuotes(sourceString);
-    textUnitUtils.replaceSourceString(textUnit, unescapedSourceString);
+    // No-op: Okapi's POFilter already fully handles all C-style escape sequences
+    // (\\, \n, \r, \t, \", etc.) via its own single-pass unescape in toAbstract().
+    // Any additional unescaping here would double-process and corrupt strings.
+    // For example, PO \\\" (escaped-backslash + escaped-quote) becomes \" (literal
+    // backslash + quote) after Okapi's unescape. Applying replaceEscapedQuotes()
+    // would then strip the literal backslash.
   }
 
   void unescapeTarget(TextUnit textUnit) {
-    TextContainer target = textUnit.getTarget(targetLocale);
-    if (target != null) {
-      String targetString = target.toString();
-      String unescapedTargetString = unescapeUtils.replaceEscapedQuotes(targetString);
-      TextContainer newTarget = new TextContainer(unescapedTargetString);
-      textUnit.setTarget(targetLocale, newTarget);
-    }
+    // No-op: same reasoning as unescapeSource.
   }
 
   boolean isPluralGroupStarting(Event event) {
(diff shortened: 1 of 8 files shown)


ehoogerbeets (dismissed):
Looks good.

Is there a decoder class too? I didn't find one... I figure `decode(encode(str))` should be the same as `str`

wadimw:
* Probably need some integration tests for backslashes against actual PO file content to ensure this is being used where needed
* Need to test roundtrip (unit and integration) to ensure symmetrical escape-unescape
* String ID also needs to be un/escaped
* Not sure about the comments - we need to check PO spec to verify if they also support some escaping
* PR should target branch `upstream-patched` (along with a label like e.g. #1049) until we agree on the maintenance responsibility model between us an Jean

wadimw:
Superceded by https://github.com/box/mojito/pull/1059

PR #1061, opened 2026-05-21
Add consistent pseudolocalization

wadimw:
This PR adds ability to produce pseudolocalized strings consistently instead of picking replacement characters at random.

In some cases, it's not desirable to get a completely different pseudolocalized string after each run of the `pseudo` command. To avoid it, this PR adds a CLI option `--substitute` which allows to select whether replacement characters are picked at random (`--substitute=RANDOM`), or consistently (`--substitute=CONSISTENT`).

Consistent pseudo means that for a given input string, pseudolocalized output will be invariant. Note that generally there will still be different diacritic characters inserted across the whole content (e.g. a `G` can still be one of `ĜĞĠĢ`) - the choice in consistent mode is based on the amount of previous occurrences of that character inside a given input string.

To preserve original behaviour, default value for this option is `RANDOM`.

### DIFF
--- a/webapp/src/test/java/com/box/l10n/mojito/pseudoloc/PseudoLocalizationTest.java
+++ b/webapp/src/test/java/com/box/l10n/mojito/pseudoloc/PseudoLocalizationTest.java
@@ -2,6 +2,7 @@ package com.box.l10n.mojito.pseudoloc;
 
 import static org.junit.Assert.*;
 
+import com.box.l10n.mojito.pseudoloc.PseudoLocalization.SubstituteType;
 import com.box.l10n.mojito.service.assetintegritychecker.integritychecker.MessageFormatIntegrityChecker;
 import com.box.l10n.mojito.service.assetintegritychecker.integritychecker.TextUnitIntegrityChecker;
 import com.box.l10n.mojito.service.assetintegritychecker.integritychecker.WhitespaceIntegrityChecker;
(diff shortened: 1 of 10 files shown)


webapp/src/test/java/com/box/l10n/mojito/pseudoloc/PseudoLocalizationTest.java
```diff
@@ -149,4 +150,50 @@ public void testConvertPluralMessageFormatStringToPseudoLoc3() {
...
+        "Placeholders should be preserved with consistent substitution", first.contains("{name}"));
+  }
+
+  @Test
+  public void testConsistentDoesNotConvertUnmappedChars() {
+    PseudoLocalization ps = new PseudoLocalization();
+    String result = ps.convertAsciiToDiacritics("qQV", SubstituteType.CONSISTENT);
+    assertEquals(
+        "Unmapped chars should remain unchanged with consistent substitution", "qQV", result);
+  }
```
ehoogerbeets, line 216: Now we need a few tests with SubstitutionType.RANDOM to verify its functionality still works as before.

ehoogerbeets (dismissed):
Approving so that after you add the requested unit tests, everything else is already good to go.

webapp/src/test/java/com/box/l10n/mojito/pseudoloc/PseudoLocalizationTest.java
wadimw: Added in ff4834b

ehoogerbeets:
The new tests look fine, but now there are seemingly unrelated unit test errors? Any ideas as to what is going on there?

ehoogerbeets (approved):

PR #1062, opened 2026-06-24
Add support for proxy servers with OpenAI

ehoogerbeets:
- modify OpenAIClient to handle proxy servers when talking to OpenAI
- add OpenAIProxyConfig to pick up the configuration for the proxy
    - proxy is allowed to be the same one as for talking to Box 3rd party API
- added same functionality for AI review
- added unit tests for the new functionality
- added documentation in the example configuration

### DIFF
--- a/common/src/main/java/com/box/l10n/mojito/openai/OpenAIProxyConfig.java
+++ b/common/src/main/java/com/box/l10n/mojito/openai/OpenAIProxyConfig.java
@@ -1,0 +1,12 @@
+package com.box.l10n.mojito.openai;
+
+public record OpenAIProxyConfig(String host, Integer port, String user, String password) {
+
+  public static OpenAIProxyConfig of(String host, Integer port, String user, String password) {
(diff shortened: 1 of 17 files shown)


common/src/main/java/com/box/l10n/mojito/openai/OpenAIProxyConfig.java
kkuliczkowski-box, line 1: Should we override `toString` method to redact the password (and ideally user)?
A Java record generates `toString(`)/`equals()`/`hashCode()` over all components, so password is printed in plaintext if this record is ever logged, included in an exception, or serialized
```
@Override
public String toString() {
  return "OpenAIProxyConfig[host=" + host + ", port=" + port
      + ", user=" + (user == null ? "null" : "***")
      + ", password=" + (password == null ? "null" : "***") + "]";
}
```

ehoogerbeets:
Review suggestion accepted. Please re-review

kkuliczkowski-box (approved):

PR #1065, opened 2026-07-22
Make TM import asynchronous

wadimw:
TM import would time out for larger files due to connection timeout. Changed the TM import to a pollable task instead (consistent with TM export and other potentially long-running tasks).

Also added a CLI config prop `l10n.resttemplate.readTimeoutSeconds` to allow temporarily overriding default connection timeout if needed urgently.

### DIFF
--- a/webapp/src/main/java/com/box/l10n/mojito/service/tm/TMImportService.java
+++ b/webapp/src/main/java/com/box/l10n/mojito/service/tm/TMImportService.java
@@ -8,6 +8,9 @@ import com.box.l10n.mojito.okapi.RawDocument;
 import com.box.l10n.mojito.service.asset.AssetRepository;
 import com.box.l10n.mojito.service.asset.AssetService;
 import com.box.l10n.mojito.service.locale.LocaleService;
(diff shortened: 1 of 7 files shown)


webapp/src/main/java/com/box/l10n/mojito/service/tm/TMImportService.java
```diff
@@ -43,18 +46,22 @@ public class TMImportService {
...
-  public void importXLIFF(Repository repository, String xliffContent, boolean updateTM) {
+  @Pollable(async = true, message = "Importing XLIFF")
+  public PollableFuture<Void> importXLIFF(
+      Repository repository, String xliffContent, boolean updateTM) {
 
     ImportExportedXliffStep importExportedXliffStep =
         new ImportExportedXliffStep(repository, xliffContent, updateTM);
-    importXLIFF(importExportedXliffStep, xliffContent);
+    doImportXLIFF(importExportedXliffStep, xliffContent);
+    return new PollableFutureTaskResult<>();
```
ehoogerbeets, line 64: Wait, what connects this returned value to the actual task?

PR #1068, opened 2026-07-25
Add --json flag to various CLI commands for use in AI translate scripts

ehoogerbeets:
## Summary

Adds a shared CLI `--json` convention so automation (especially AI-translate Jenkins jobs) can consume a stable machine-readable result instead of scraping ANSI console output.

- Introduces a common JSON envelope on stdout: `{ successful, command, data, error? }`, with human console progress suppressed when `--json` is set
- Wires `--json` into: `repo-create`, `repo-view`, `drop-export`, `drop-import`, `tm-export`, `tm-import`, and `repository-ai-translate`
- For `repository-ai-translate`, the AI translate service now detects remaining work at selection time (fetch `sourceTextMaxCount + 1` per locale) and stores `hasMore` / `hasMoreByLocale` on the pollable job output; the CLI surfaces those fields in the `--json` result so schedulers can loop until a repo is drained

## Test plan

- [x] `mvn spotless:check`
- [x] Unit: `JsonOutputTest`, AI translate config/auto-fix integrity tests
- [x] CLI integration: `RepoCreateCommandTest` (incl. `--json`), `RepoViewCommandTest`, `DropExportCommandTest`, `DropImportCommandTest`, `TMExportCommandTest`, `TMImportCommandTest`
- [ ] Manually run `repository-ai-translate --json -r <repo> -l hi-IN bn-IN` against a repo with more than `--source-text-max-count` strings and confirm `hasMore: true`, then re-run until `hasMore: false`
- [ ] Confirm failure paths still exit non-zero and emit `{ successful: false, error: "..." }` with `--json`

### DIFF
--- a/cli/src/main/java/com/box/l10n/mojito/cli/command/Command.java
+++ b/cli/src/main/java/com/box/l10n/mojito/cli/command/Command.java
@@ -5,9 +5,11 @@ import com.beust.jcommander.Parameters;
 import com.google.common.base.Preconditions;
 import java.util.Arrays;
 import java.util.List;
+import org.springframework.beans.factory.annotation.Autowired;
 
 /**
- * Base class for Commands, provides basic support for usage display.
+ * Base class for Commands, provides basic support for usage display and the shared {@code --json}
+ * machine-readable output convention.
  *
  * @author jaurambault
  */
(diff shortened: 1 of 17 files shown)


ehoogerbeets:
Abandoning. The "mojito-cli api" call does the same thing only better. We don't need this one any more.

PR #1069, opened 2026-07-28
Fix proxy bearer token

ehoogerbeets:


ehoogerbeets:
Abandoned in favor of #1070

PR #1070, opened 2026-07-30
Fix OpenAI webproxy auth with Apache HttpClient 5

ehoogerbeets:
## Summary
- Switch OpenAI HTTP clients from JDK `java.net.http.HttpClient` to Apache HttpClient 5 (already used elsewhere in Mojito) so proxy auth negotiation and the OpenAI `Authorization: Bearer` header work correctly (avoids JDK-8326949 / silent hangs through webproxy).
- Keep connect timeouts so unreachable proxies fail quickly instead of hanging, and restore/enable Basic tunneling for HTTPS CONNECT where needed.
- Add optional `proxy-preferred-auth-schemes` configuration so deployments can override HttpClient’s default auth scheme preference when a proxy advertises schemes it does not actually support (e.g. Digest before Basic).
- Add `OpenAIClientIntegrationTest` (opt-in via configured OpenAI properties) as a smoke test for end-to-end ChatGPT connectivity, including webproxy support.
- Document the new proxy auth scheme settings in the Spring Boot configuration docs.

## Context
Follow-up to [#1069](https://github.com/box/mojito/pull/1069) (`fixProxyBearerToken`), incorporating review feedback and the preferred-auth-schemes enhancement on branch `fixProxyBearerToken2`.

## Test plan
- [x] Unit tests pass for `OpenAIHttpClientFactory`, `OpenAIClient`, and AI translate/review config property wiring
- [x] With OpenAI + webproxy configured locally, run `OpenAIClientIntegrationTest` and confirm a successful ChatGPT round trip through the proxy
- [x] Verify default behavior (no `proxy-preferred-auth-schemes`) still works with HttpClient defaults
- [x] When set (e.g. `Basic,Digest`), confirm preferred schemes are applied on the request config and CONNECT succeeds against a Digest-advertising proxy that only works with Basic
- [x] Confirm AI translate / AI review services still construct clients with the updated proxy config

Made with [Cursor](https://cursor.com)


wadimw (approved):
Conditionally approving for now - as discussed, we need to clean up the redundant test scaffolding and overall condense the proxy-related changes into a more focused diff. (My draft: https://github.com/box/mojito/pull/1071)

ehoogerbeets:
Abandoning in favor of #1071

PR #1076, opened 2026-08-19
LXP-2065: Add Repo Types API for shared stack config

ehoogerbeets:
## Summary

This PR is a **persistence + REST foundation** for repo types. It does not assign a type to a repository, so grouping similar repositories is not user-visible yet.

- Introduces **Repo Types**: named shared configs (e.g. React, Android) that hold a type-layer AI prompt and default integrity checkers, so those settings are not duplicated per repository.
- Adds full backend CRUD under `/api/repo-types` (entities, Flyway `V69`, `RepoTypeService`, `RepoTypeWS`, and `RepoTypeClient`), with PATCH semantics where `null`/omitted fields leave values unchanged and a non-null `integrityCheckers` set replaces the full checker list.
- Documents the design and contracts in `docs/internal/Architecture.md`, and covers behavior with service + REST contract tests (TDD).

**In scope for this PR:** persist types and checkers; authenticated GET; mutating APIs for PM/ADMIN.

**Out of scope (follow-up):** assigning a type on a repository, CLI commands, UI, layered prompt assembly at runtime, and unioning type + repo checkers on push/import.

There is no CLI parity with this API. Operators cannot create, update, view, or delete types from the CLI (`repo-type-create` / `repo-type-update` / `repo-type-view` / `repo-type-delete` are follow-up). `repo-view` inspects a repository (id, repo-level checkers, locales) and will **not** show a repo type after this PR, because repositories cannot be assigned a type yet. When assignment lands, `repo-view` / `repo-create` / `repo-update` must grow a type field (LXP-2068). This PR does not close “everything the API can do, the CLI can do.”

## Test plan
- [ ] `mvn -pl webapp -am test -Dtest=RepoTypeServiceTest,RepoTypeWSTest -Dsurefire.failIfNoSpecifiedTests=false`
- [ ] `JAVA_HOME=<jdk-21> mvn spotless:check`
- [ ] Manually exercise `/api/repo-types` create → get → list/`?name=` → patch → delete (201/200/204, 404 missing, 409 duplicate name)
- [ ] Confirm Flyway `V69__Add_repo_type.sql` applies cleanly on a MySQL-backed environment


webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    logger.debug("Deleting unused integrity checkers for repo type id: {}", repoType.getId());
+    for (Map<String, RepoTypeIntegrityChecker> byType : existingToDelete.values()) {
+      for (RepoTypeIntegrityChecker toDelete : byType.values()) {
+        repoTypeIntegrityCheckerRepository.delete(toDelete);
+      }
+    }
+
+    repoTypeIntegrityCheckerRepository.saveAll(integrityCheckers);
```
zmoskala: The docs say a checker is unique per type on (assetExtension, integrityCheckerType), but the incoming list isn’t actually de-duplicated.

Set doesn’t help here: the entity has no equals/hashCode on that pair, so Jackson will keep two JSON objects with the same fields. saveAll just saves each item; the table has no unique constraint on that triple either.

So this body can create two identical rows:

`"integrityCheckers": [
  { "assetExtension": "json", "integrityCheckerType": "SIMPLE_PRINTF_LIKE" },
  { "assetExtension": "json", "integrityCheckerType": "SIMPLE_PRINTF_LIKE" }
]`

webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java
```diff
@@ -0,0 +1,277 @@
...
+
+    Set<RepoTypeIntegrityChecker> existingCheckers =
+        repoTypeIntegrityCheckerRepository.findByRepoType(repoType);
+    Map<String, Map<String, RepoTypeIntegrityChecker>> existingToDelete =
+        getIntegrityCheckerMap(existingCheckers);
+
+    for (RepoTypeIntegrityChecker integrityChecker : integrityCheckers) {
+      logger.debug(
+          "Setting repo type for integrity checker: {}", integrityChecker.getAssetExtension());
+      integrityChecker.setRepoType(repoType);
```
zmoskala: Ids on the incoming checkers are never cleared. When the (assetExtension, integrityCheckerType) pair already exists we overwrite the id on line 246, but when the pair is new we keep whatever id arrived in the request body - and saveAll on line 262 then merges into that row, which may belong to a different repo type. Realistic trigger: someone GETs type A, edits the JSON, PATCHes it onto type B without stripping ids; the result is a checker silently reparented.

Ids are server-owned here (identity is the extension/type pair), so I think we should drop them on input:

```suggestion
      integrityChecker.setRepoType(repoType);
      integrityChecker.setId(null);
```

The existing-pair branch below still restores the real id, so id reuse (covered by testUpdateIntegrityCheckersReusesIdForSamePair) keeps working. RepositoryService.updateAssetIntegrityCheckers has the same hole, so this is inherited rather than introduced - but it seems better not to copy it into a new API.

webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java
```diff
@@ -0,0 +1,277 @@
...
+    logger.debug("Check no repo type with name: {} exists", name);
+
+    if (repoTypeRepository.findByName(name) != null) {
+      throw new RepoTypeNameAlreadyUsedException(name + " is used by another repo type");
+    }
+
+    logger.debug("Create a RepoType with name: {}", name);
+
+    RepoType repoType = new RepoType();
+    repoType.setName(name);
```
zmoskala, line 86: name is mandatory (`@Basic(optional = false)` on the entity, `NOT NULL` in `V69__Add_repo_type.sql`) but nothing validates it before save.
- `POST /api/repo-types` with `{}` (`name` null) fails in the persistence layer, so the client gets **500** rather than 400.
- `{"name": ""}` (or whitespace) satisfies `NOT NULL` and **creates a type named `""`**.
- name/description longer than 255 chars also 500.
Architecture.md lists `name` as the one required field, so this looks like a gap rather than intent. Could we validate here and map it to 400?
```java
if (StringUtils.isBlank(name)) {
  throw new RepoTypeNameInvalidException("name is required");
}

File: webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java [hunk 4]
```diff
@@ -0,0 +1,277 @@
...
+      repoType.setName(name);
+    }
+    if (description != null) {
+      repoType.setDescription(description);
+    }
+    if (aiPrompt != null) {
+      repoType.setAiPrompt(aiPrompt);
+    }
+
+    if (name != null || description != null || aiPrompt != null) {
```
zmoskala, line 201: A PATCH that only sends integrityCheckers skips this save entirely, so repo_type.last_modified_date does not move even though the type's effective config changed. Since the entity extends AuditableEntity specifically to expose created/modified timestamps (Architecture.md:88), an operator diffing timestamps would conclude nothing changed.

Is that intentional? If not, touching the parent when checkers are replaced would fix it - either move the save below the updateIntegrityCheckers call, or add integrityCheckers != null to this condition and make sure the audit listener actually fires (a no-op save on an unchanged entity may not issue an UPDATE, so setLastModifiedDate(...) explicitly may be needed).

webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java
```diff
@@ -0,0 +1,277 @@
...
+  public RepoType createRepoType(
+      String name,
+      String description,
+      String aiPrompt,
+      Set<RepoTypeIntegrityChecker> integrityCheckers)
+      throws RepoTypeNameAlreadyUsedException {
+
+    logger.debug("Check no repo type with name: {} exists", name);
+
+    if (repoTypeRepository.findByName(name) != null) {
```
zmoskala, line 79: This check and the save on line 74 are not atomic. Two concurrent POSTs with the same name can both pass here; the loser then violates UK__REPO_TYPE__NAME and Hibernate throws DataIntegrityViolationException, which nothing catches - so the caller gets 500 instead of the 409 documented in Architecture.md:176.

RepoTypeNameAlreadyUsedException(Throwable) at RepoTypeNameAlreadyUsedException.java:21 is currently unused and looks like it was added for exactly this case. Could we use it?

```java
try {
  repoType = repoTypeRepository.saveAndFlush(repoType);
} catch (DataIntegrityViolationException e) {
  throw new RepoTypeNameAlreadyUsedException(e);
}
```

webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java
```diff
@@ -0,0 +1,277 @@
...
+
+    for (RepoTypeIntegrityChecker integrityChecker : integrityCheckers) {
+      logger.debug(
+          "Setting repo type for integrity checker: {}", integrityChecker.getAssetExtension());
+      integrityChecker.setRepoType(repoType);
+      Map<String, RepoTypeIntegrityChecker> existingForExtension =
+          existingToDelete.get(integrityChecker.getAssetExtension());
+      if (existingForExtension != null) {
+        RepoTypeIntegrityChecker existing =
+            existingForExtension.get(integrityChecker.getIntegrityCheckerType().name());
```
zmoskala: getIntegrityCheckerType() is dereferenced without a null check here (and again on 247 and 273). A body like {"integrityCheckers": [{"assetExtension": "properties"}]} passes Jackson - an unknown enum value is a 400, but an absent one is just null - and then NPEs inside the transaction, so the caller sees 500. A null assetExtension is quieter but also wrong: it becomes a null key in getIntegrityCheckerMap and a NOT NULL violation at flush.

Both fields are required per Architecture.md:140, so validating them at the top of updateIntegrityCheckers and throwing a 400-mapped exception would keep the contract honest. A RepoTypeWSTest case posting a checker without integrityCheckerType would lock it in.

webapp/src/main/java/com/box/l10n/mojito/rest/repotype/RepoTypeWS.java
```diff
@@ -0,0 +1,157 @@
...
+  /**
+   * Hard-deletes a repo type and its integrity checkers.
+   *
+   * <p>HTTP 204 on success (void return). Unknown id → HTTP 404.
+   *
+   * @param repoTypeId id of the type to delete
+   * @throws RepoTypeWithIdNotFoundException if missing → HTTP 404
+   */
+  @RequestMapping(value = "/api/repo-types/{repoTypeId}", method = RequestMethod.DELETE)
+  public void deleteRepoType(@PathVariable Long repoTypeId) throws RepoTypeWithIdNotFoundException {
```
zmoskala, line 178: The JavaDoc above and Architecture.md:177 both promise 204 No Content, but a void handler with no @ResponseStatus resolves to 200 OK, so the contract and the implementation disagree. A client written against the doc that treats only 204 as success would mis-handle a successful delete.

```suggestion
  @ResponseStatus(HttpStatus.NO_CONTENT)
  @RequestMapping(value = "/api/repo-types/{repoTypeId}", method = RequestMethod.DELETE)
  public void deleteRepoType(@PathVariable Long repoTypeId) throws RepoTypeWithIdNotFoundException {
```

(needs the org.springframework.web.bind.annotation.ResponseStatus import). Alternatively drop 204 from the doc and keep 200 like RepositoryWS does. Either way, RepoTypeWSTest:282 currently only checks that a later GET is 404 - asserting the delete status there would keep the two in sync.

webapp/src/main/java/com/box/l10n/mojito/entity/RepoType.java
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * Shared type-layer AI prompt. Defaults to empty string on create. Empty means “no type-layer
+   * instructions”; {@code null} should be normalized to empty when persisting.
+   *
+   * <p>Defaults to {@code null} so omitted JSON on PATCH deserializes as “leave unchanged”. Create
+   * normalizes {@code null} to empty string before persist.
+   */
+  @Lob
+  @Column(name = "ai_prompt", length = Integer.MAX_VALUE)
+  @JsonView(View.IdAndName.class)
```
zmoskala: View.IdAndName means "id and name only" everywhere else in this codebase — `Repository` keeps description on `View.RepositorySummary` and checkers on `View.Repository`. Here `name`, `description`, `aiPrompt` and `integrityCheckers` are all on `IdAndName`, so any future payload that nests a `RepoType` under an `IdAndName` view (e.g. a repository summary once the FK lands) will serialize the whole prompt.

Could we add a dedicated view and keep `IdAndName` to id + name?

```java
// View.java
// Include IdAndNameAndCreated + Modified if this API should return createdDate / lastModifiedDate
// (today RepoTypeWS uses IdAndName, so those columns never appear in JSON — see comment on RepoTypeWS).
public interface RepoType extends IdAndName, IdAndNameAndCreated, Modified {}
```

Then annotate `description` / `aiPrompt` / `integrityCheckers` with `View.RepoType.class` and use `@JsonView(View.RepoType.class)` on the `RepoTypeWS` methods. Nested summaries that still use `IdAndName` would get id+name only; this API would get the full type.

Non-blocking, but cheaper to do now than after the entity is referenced elsewhere.

webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java
```diff
@@ -0,0 +1,277 @@
...
+      throws RepoTypeWithIdNotFoundException, RepoTypeNameAlreadyUsedException {
+
+    RepoType repoType = getRepoTypeById(repoTypeId);
+
+    if (name != null) {
+      RepoType existing = repoTypeRepository.findByName(name);
+      if (existing != null && !repoType.getId().equals(existing.getId())) {
+        throw new RepoTypeNameAlreadyUsedException(name + " is used by another repo type");
+      }
+      repoType.setName(name);
```
zmoskala, line 191: Names are persisted exactly as sent here and on create (line 70). "React " and "React" are different values for findByName and for UK__REPO_TYPE__NAME, so operators can end up with near-duplicate types that look identical in a UI, and getRepoTypes("React") will miss the padded one.

A name.trim() before the uniqueness lookup in both createRepoType and updateRepoType would remove the ambiguity. Case sensitivity depends on the MySQL collation, so probably not worth normalizing, but trimming is cheap.

restclient/src/main/java/com/box/l10n/mojito/rest/entity/RepoType.java
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * @param aiPrompt type-layer prompt; empty string clears it
+   */
+  public void setAiPrompt(String aiPrompt) {
+    this.aiPrompt = aiPrompt;
+  }
+
+  /**
+   * @return integrity checkers; never {@code null} after construction
```
zmoskala: This is not true: the field is deliberately left null (see the field comment on line 24) so PATCH can omit it, and new RepoType() therefore returns null here. The risk is that whoever writes the follow-up CLI reads "never null", initialises new HashSet<>() to be safe, and ships a PATCH that serializes integrityCheckers: [] - which RepoTypeService:223 treats as "delete all checkers".

```suggestion
  /**
   * @return integrity checkers, or {@code null} when unset - {@code null} is omitted from PATCH
   *     bodies (leave unchanged), while an empty set clears all checkers
   */
```

zmoskala (commented):
RepoTypeWSTest / RepoTypeServiceTest do not cover: missing/blank name, over-length fields, checker missing type/extension, duplicate checker pairs, client-supplied checker ids, delete status code, or timestamps.

webapp/src/main/java/com/box/l10n/mojito/rest/repotype/RepoTypeWS.java
```diff
@@ -0,0 +1,157 @@
...
+  @Autowired RepoTypeService repoTypeService;
+
+  /**
+   * Returns a single repo type by id.
+   *
+   * @param repoTypeId path id
+   * @return the type including {@code integrityCheckers}
+   * @throws RepoTypeWithIdNotFoundException if missing → HTTP 404
+   */
+  @JsonView(View.IdAndName.class)
```
zmoskala: Same `@JsonView(View.IdAndName.class)` as the other three methods — so this API never returns `createdDate` / `lastModifiedDate`.

Those fields live on `AuditableEntity` as `IdAndNameAndCreated` and `Modified`. A parent view does not include child-only (or unrelated) views, so Jackson drops both. Architecture.md's example JSON also omits them.

If timestamps should be in the response, fold them into the dedicated `View.RepoType` suggested on `RepoType.java` (`extends IdAndName, IdAndNameAndCreated, Modified`) and switch these four methods to `@JsonView(View.RepoType.class)`. If they are DB-only, ignore this and we can clarify that on the `RepoType` JavaDoc instead.

Not asking for a second, different view change here.

restclient/src/main/java/com/box/l10n/mojito/rest/entity/RepoTypeIntegrityChecker.java
```diff
@@ -0,0 +1,79 @@
...
+public class RepoTypeIntegrityChecker {
+
+  private Long id;
+
+  /** Parent type; omitted when serializing nested under {@link RepoType}. */
+  @JsonBackReference("integrityCheckers")
+  private RepoType repoType;
+
+  private String assetExtension;
+  private IntegrityCheckerType integrityCheckerType;
```
zmoskala, line 18: This field is deserialized with restclient `IntegrityCheckerType`, which is missing `FLUENT` (the server enum has it).
Unknown enum values are not covered by `FAIL_ON_UNKNOWN_PROPERTIES = false` — Jackson fails the whole payload. So a type that was created with a `FLUENT` checker (easy via curl / a future UI) cannot be read back through `RepoTypeClient` at all.
Could we add `FLUENT` to the client enum in this PR? Architecture.md says type-level and repo-level checkers use the same names so they can be unioned later; a missing value breaks that on the client side.
A `RepoTypeWSTest` that creates a checker with `FLUENT` and GETs it through the client would lock this in. The current tests only use `MESSAGE_FORMAT` / `TRAILING_WHITESPACE`.

webapp/src/main/java/com/box/l10n/mojito/entity/RepoType.java
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * @param aiPrompt type-layer prompt; prefer empty string over {@code null} when clearing
+   */
+  public void setAiPrompt(String aiPrompt) {
+    this.aiPrompt = aiPrompt;
+  }
+
+  /**
+   * @return integrity checkers for this type; never {@code null} (may be empty)
```
zmoskala: Same mismatch as the restclient getter: this is not true for a deserialized request body.
The field is left `null` on purpose so omitted JSON on PATCH means “leave checkers unchanged” (see the field comment above). `new RepoType()` / `POST`/`PATCH` with no `integrityCheckers` key therefore returns `null` here. Callers that trust “never null” and pass `new HashSet<>()` will serialize `[]`, and `RepoTypeService.updateRepoType` treats that as “delete all checkers”.
After load/create the collection is an empty set, so the “never null” part is only true on the persist path — not on inbound JSON. Could we say that explicitly?
```java
  /**
   * @return integrity checkers, or {@code null} when unset. {@code null} on PATCH means leave
   *     unchanged; an empty set clears all checkers. After persist/load this is never {@code null}.
   */

File: webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java [hunk 14]
```diff
@@ -0,0 +1,277 @@
...
+      String name,
+      String description,
+      String aiPrompt,
+      Set<RepoTypeIntegrityChecker> integrityCheckers)
+      throws RepoTypeWithIdNotFoundException, RepoTypeNameAlreadyUsedException {
+
+    RepoType repoType = getRepoTypeById(repoTypeId);
+
+    if (name != null) {
+      RepoType existing = repoTypeRepository.findByName(name);
```
zmoskala, line 187: Same non-atomic uniqueness check as create, on the rename path.
Two concurrent PATCHes that set the same new `name` can both pass `findByName` here; the loser then violates `UK__REPO_TYPE__NAME`. Nothing catches `DataIntegrityViolationException`, so the caller gets 500 instead of the 409 in Architecture.md.
Wrapping only `createRepoType`'s save is not enough — the `save` a few lines below needs the same treatment (`saveAndFlush` inside the try, or the exception is thrown at commit after this method returns). The unused `RepoTypeNameAlreadyUsedException(Throwable)` constructor looks meant for this.
```java
try {
  repoType = repoTypeRepository.saveAndFlush(repoType);
} catch (DataIntegrityViolationException e) {
  throw new RepoTypeNameAlreadyUsedException(e);
}

File: webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java (hunk 9 above)
Comment (zmoskala) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java:191]: If we trim on create/update, `getRepoTypes` needs the same — otherwise `GET /api/repo-types?name=React%20` still misses `React`.
`isBlank` already treats `"   "` as "list all". A trailing space is not blank, so `findByName(name)` is exact and returns `[]`.
```java
RepoType repoType = repoTypeRepository.findByName(name.trim());
```
Worth a test: create `"React"`, query `?name=React ` (trailing space), expect the same row.

webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java
zmoskala: Related, once the field is present: `"json"`, `"json "`, and `".json"` are three different keys in `getIntegrityCheckerMap`, so upsert inserts duplicates instead of reusing the row. The JavaDoc on `assetExtension` says no leading dot.
If we're validating `integrityCheckerType` / `assetExtension` at the top of `updateIntegrityCheckers`, worth also `trim()` (and optionally stripping a leading `.`) so the pair we persist is the pair we match on.

webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java
zmoskala: Additionally, this would be consistent with CLI behavior - I tested with repo-type-create I am adding, and:
```
zmoskala@mbp-020010 mojito % mojito-local repo-type-create -n DupType
Create repo type: DupType

Repo type with name [DupType] already exists

zmoskala@mbp-020010 mojito % mojito-local repo-type-create -n DupType 
Create repo type: DupType

Repo type with name [DupType] already exists

zmoskala@mbp-020010 mojito % mojito-local repo-type-create -n DupType    
Create repo type: DupType

Repo type with name [DupType] already exists 
```

(2nd DupType was just "DupType", 2rd was "DupType "). In this case, the trailing space never reached the server. The CLI strips it before create. JCommander trims argument values, even when quoted. That is already documented in the repo: 
```/**
 * The current version we have for JCommander trims the argument values, even when quoted.
 * https://github.com/cbeust/jcommander/issues/417
 * https://github.com/cbeust/jcommander/commit/4aec38b4a0ea63a8dc6f41636fa81c2ebafddc18```

File: webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java (hunk 1 above)
Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java]: Addressed. Checkers are now an `@Embeddable` of only `(assetExtension, integrityCheckerType)` with `equals`/`hashCode` on that pair, stored as an `@ElementCollection` on `RepoType`. The join table primary key is `(repo_type_id, asset_extension, integrity_checker_type)`, so duplicate pairs cannot persist. `updateIntegrityCheckers` still collapses incoming duplicates (last occurrence wins) before replace.

File: webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java (hunk 2 above)
Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java]: Addressed by going further than clearing ids. Checker `id` and parent are no longer on `RepoTypeIntegrityChecker` at all — they were server housekeeping and created this merge/reparent hole. Clients get and send only `{ assetExtension, integrityCheckerType }`. Hibernate owns the join-table FK via the type's element collection, so a PATCH cannot target another type's row by id.

File: webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java (hunk 4 above)
Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java:201]: Addressed, not intentional. A checkers-only PATCH never dirty-marks `repo_type` (the collection lives in a join table), so `@LastModifiedDate` would not fire. `updateIntegrityCheckers` now sets `lastModifiedDate` before save so the parent timestamp moves when checkers are replaced. `createdDate` is unchanged. (Exposing these timestamps in the REST JSON view is a separate comment.)

File: webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java (hunk 3 above)
Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java:86]: Addressed. Create and update now validate before persist and throw `RepoTypeInvalidException` (HTTP 400) when `name` is null/blank/whitespace, or when `name` or `description` exceeds 255 characters. Empty name is no longer stored; over-length fields no longer 500 at flush. (Trimming `"React "` vs `"React"` is a later comment and is not in this change.)

File: webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java (hunk 14 above)
Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java:187]: Addressed together with the create-path comment. Rename goes through the same `saveAndFlush` helper, so a concurrent PATCH that hits `UK__REPO_TYPE__NAME` also becomes `RepoTypeNameAlreadyUsedException` (HTTP 409) instead of failing at commit.

File: webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java (hunk 5 above)
Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java:79]: Addressed. Create now uses `saveAndFlush` and maps `DataIntegrityViolationException` to `RepoTypeNameAlreadyUsedException` (the unused cause constructor) so a lost race on `UK__REPO_TYPE__NAME` is HTTP 409, not 500. The same helper is used on the rename save, per the follow-up on the PATCH path — wrapping only create would still 500 on concurrent renames.

Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java:79]: Correction on the earlier reply: we do **not** catch `DataIntegrityViolationException` inside the `@Transactional` service method. Catching it there (or wrapping it as `RepoTypeNameAlreadyUsedException` before the transaction ends) can leave the persistence context unusable, and a checked wrap can still try to commit.

Create and update use `saveAndFlush` so a lost race on `UK__REPO_TYPE__NAME` fails in-call. After the transaction rolls back, `RepoTypeWS` maps both `RepoTypeNameAlreadyUsedException` (sequential duplicate from `findByName`) and `DataIntegrityViolationException` (constraint failure after flush) to HTTP 409. Anything else uncaught is still Spring Boot’s default 500.

`RepoTypeNameAlreadyUsedException(Throwable)` is unused again; the WS catch is the mapping, not a service-layer wrap.

File: webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java (hunk 14 above)
Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java:187]: Correction on the earlier reply: rename uses the same `saveAndFlush` as create, and **the WS** maps `DataIntegrityViolationException` (and sequential `RepoTypeNameAlreadyUsedException`) to 409. We do not catch the constraint violation inside the transactional service — that would run while the session is already marked rollback-only.

A concurrent PATCH that loses `UK__REPO_TYPE__NAME` therefore 409s the same way as concurrent POST, rather than 500ing at commit.

File: webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java (hunk 6 above)
Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java]: Addressed. `updateIntegrityCheckers` now validates each checker before de-dupe: missing `integrityCheckerType` or blank `assetExtension` throws `RepoTypeInvalidException` (HTTP 400) instead of NPE / flush 500.

Unknown enum tokens (`NOT_A_CHECKER`) still fail in Jackson before the service (`HttpMessageNotReadableException` → 400). Tests cover missing fields (service + WS) and unknown names via raw JSON POST/PATCH.

Related follow-up on this line: `assetExtension` is trimmed and a single leading `.` is stripped so `json`, `json `, and `.json` are the same pair.

Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java]: Addressed together with the null-field comment. `normalizeChecker` trims `assetExtension` and strips one leading `.` before the de-dupe key and persist, so those three spellings collapse to `json`.

File: restclient/src/main/java/com/box/l10n/mojito/rest/entity/RepoTypeIntegrityChecker.java (hunk 12 above)
Comment (ehoogerbeets) [reply restclient/src/main/java/com/box/l10n/mojito/rest/entity/RepoTypeIntegrityChecker.java:18]: Addressed. Restclient `IntegrityCheckerType` now includes `FLUENT` (same order as the server enum). Create/GET of a type with a Fluent checker round-trips through `RepoTypeClient`.

File: webapp/src/main/java/com/box/l10n/mojito/rest/repotype/RepoTypeWS.java (hunk 7 above)
Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/rest/repotype/RepoTypeWS.java:178]: Addressed. DELETE now has `@ResponseStatus(HttpStatus.NO_CONTENT)` so success is 204, matching the JavaDoc and Architecture.md. `testDeleteRepoTypeRemovesType` asserts 204 (void + no annotation was 200). Missing id is still 404.

File: webapp/src/main/java/com/box/l10n/mojito/entity/RepoType.java (hunk 8 above)
Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/entity/RepoType.java]: Addressed. Added `View.RepoType` extending `IdAndName`, `IdAndNameAndCreated`, and `Modified`. `name` (and `id`) stay on `IdAndName`. Description, prompt, and checkers are on `View.RepoType` only, so a future nested summary under `IdAndName` will not serialize the prompt. GET/POST/PATCH use `@JsonView(View.RepoType.class)`.

File: webapp/src/main/java/com/box/l10n/mojito/rest/repotype/RepoTypeWS.java (hunk 11 above)
Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/rest/repotype/RepoTypeWS.java]: Addressed together with the dedicated-view comment. `View.RepoType` extends the created and modified views, so `/api/repo-types` now includes `createdDate` and `lastModifiedDate`. Restclient maps those fields; create asserts they are present. A prompt-only update test asserts `lastModifiedDate` moves and `createdDate` does not (checkers-only was already covered).

File: webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java (hunk 9 above)
Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java:191]: Addressed. Create and PATCH trim `name` before uniqueness and persist, so `"React "` is stored as `"React"`. A padded create of an existing name is 409. CLI/JCommander already trims `-n`, so this matches that behavior.

Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java:191]: Addressed. `getRepoTypes` still treats blank as list-all; a non-blank filter is trimmed before `findByName`. `GET ?name=React%20` finds `React`. Covered in service and WS tests.

Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java:191]: Addressed. Server-side trim now matches CLI trimming of `-n`, including the padded duplicate case you showed.

File: restclient/src/main/java/com/box/l10n/mojito/rest/entity/RepoType.java (hunk 10 above)
Comment (ehoogerbeets) [reply restclient/src/main/java/com/box/l10n/mojito/rest/entity/RepoType.java]: Addressed. The restclient getter JavaDoc now says the field can be `null` when unset: omit on PATCH to leave checkers unchanged; `[]` still clears all.

File: webapp/src/main/java/com/box/l10n/mojito/entity/RepoType.java (hunk 13 above)
Comment (ehoogerbeets) [reply webapp/src/main/java/com/box/l10n/mojito/entity/RepoType.java]: Addressed. Entity getter JavaDoc matches the PATCH contract: `null` means the field was never set (leave unchanged); after load/create it is a set, possibly empty.

**Comment by ehoogerbeets** [comment]
On the review note that `RepoTypeWSTest` / `RepoTypeServiceTest` were missing several cases — those are covered now:

- missing/blank name and over-length name/description → 400
- checker missing type or extension (and unknown enum names) → 400
- duplicate checker pairs collapse; extensions trim / strip a leading `.`
- checker ids: removed from the type (no client-supplied id / reparent)
- DELETE success → 204
- timestamps in the `View.RepoType` payload; checkers-only and prompt-only PATCH bump `lastModifiedDate`

Also added: unique-name 409 (including flush-time races mapped in the WS), name trim on create/update/GET, and `FLUENT` on the restclient enum.

File: webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java [hunk 15]
```diff
@@ -0,0 +1,333 @@
...
+  /**
+   * Collapses checkers that share {@code (assetExtension, integrityCheckerType)} to a single
+   * instance. Last occurrence wins so a later duplicate in the request body is what we persist.
+   * Validates and normalizes each checker first so the de-dupe key matches what is stored.
+   */
+  private Set<RepoTypeIntegrityChecker> uniqueByExtensionAndType(
+      Set<RepoTypeIntegrityChecker> integrityCheckers) {
+    Map<String, RepoTypeIntegrityChecker> unique = new LinkedHashMap<>();
+    for (RepoTypeIntegrityChecker checker : integrityCheckers) {
+      normalizeChecker(checker);
```
zmoskala, line 293: A body like `"integrityCheckers": [null]` still 500s: the loop calls `normalizeChecker(checker)` with no null check, so this NPEs inside the transaction instead of `RepoTypeInvalidException` (400). Same class of bug as a checker missing `integrityCheckerType` / `assetExtension`.

webapp/src/main/java/com/box/l10n/mojito/rest/repotype/RepoTypeWS.java
```diff
@@ -0,0 +1,164 @@
...
+          repoTypeService.createRepoType(
+              repoType.getName(),
+              repoType.getDescription(),
+              repoType.getAiPrompt(),
+              repoType.getIntegrityCheckers());
+      return new ResponseEntity<>(created, HttpStatus.CREATED);
+    } catch (RepoTypeNameAlreadyUsedException | DataIntegrityViolationException e) {
+      logger.debug("Cannot create the repo type", e);
+      return new ResponseEntity<>(
+          "RepoType with name [" + repoType.getName() + "] already exists", HttpStatus.CONFLICT);
```
zmoskala: Name is trimmed in the service before uniqueness, so `"React "` vs existing `"React"` is correctly 409 and nothing untrimmed is stored.

This body still uses `repoType.getName()` from the request, so the client sees `[React ]` (trailing space) instead of the name that actually conflicted. Could we trim here (or use the name from `RepoTypeNameAlreadyUsedException`) so the 409 text matches what was matched?

```java
String conflictName = repoType.getName() == null ? null : repoType.getName().trim();
return new ResponseEntity<>(
    "RepoType with name [" + conflictName + "] already exists", HttpStatus.CONFLICT);```

webapp/src/main/java/com/box/l10n/mojito/service/repotype/RepoTypeService.java
ehoogerbeets: Addressed. `normalizeChecker` now rejects a `null` list element with `RepoTypeInvalidException` (HTTP 400) instead of NPEing inside the transaction. Same class of invalid checker as a missing `integrityCheckerType` / `assetExtension` — we do not skip/drop nulls.

Covered in `RepoTypeServiceTest` and WS tests posting/PATCHing raw JSON `"integrityCheckers": [null]`.

webapp/src/main/java/com/box/l10n/mojito/rest/repotype/RepoTypeWS.java
ehoogerbeets: Addressed. POST and PATCH 409 bodies now go through a shared helper that trims the request name, so `"React "` vs existing `"React"` still 409s but the text is `[React]`, matching what uniqueness actually matched. `DataIntegrityViolationException` has no name on it, so we trim the request rather than reading the exception.

WS tests assert the padded-create and padded-rename 409 bodies quote the trimmed name.

nmkedziora (changes requested):
needs review

zmoskala (dismissed):

nmkedziora:
# Human & Agentic Review Notes

Thanks for `Architecture.md`. The in/out-of-scope table made this review possible.

I need the must-fix items below addressed in this PR (or explicitly split out with a follow-up). I will not +1 with them open. Please do not park them as “will fix later.”

Also please resolve in the PR description, before CLI/UI copy the words:

1. The AI Translations PRD says **project types**. This code and API say **repo type**. Same concept. Pick one term and use it everywhere. “Project” already means something else to Mojito operators; shipping both will confuse CLI flags, UI labels, and tickets. Document the choice in `Architecture.md` and in the PR body.
2. This branch is a **persistence + REST foundation**. It does not assign a type to a repository, so E2-F1 (“group similar repositories”) is not user-visible yet. Say that in the summary so nobody reviews this as a finished feature.

---

### Split unrelated work

The only production change that is not repo-type CRUD is `restclient` `IntegrityCheckerType` gaining `FLUENT`.

That is a real bugfix: the server enum already had `FLUENT`; the client did not. `repo-view` / `repo-update` can fail to deserialize a repository that uses a FLUENT checker. It is not a repo-type feature, and it changes a shared client contract used by existing CLI commands.

Please revert that enum change here and open a one-line PR: add `FLUENT` to the restclient enum so it matches the server. Add a test that the two enums have the same **name set** (do not compare ordinals; they already diverge).

`View.java`’s `RepoType` JSON view can stay; that is part of this API.

---

### Naming: grammar, uniqueness, closed set, rename tests

Name validation is under-specified for something people will type in a CLI later.

Today we trim, reject blank, cap at 255, and enforce exact-string uniqueness. There is no case folding, no pattern, no PascalCase rule, no separator rule for multi-word names. `React` and `react` can both exist (`utf8_bin`). That will bite the first time two people create both.

Please:

1. Document the allowed name grammar in `Architecture.md`. Suggestion: non-empty, trim, 1–255, unique; recommend PascalCase display names (`React`, `Android`, `WindowsResw`); do not invent a required separator.
2. Add a test that creating `React` does not block `react`, **or** (preferred) treat names as case-insensitive for uniqueness and implement that in `findByName` **and** the unique constraint. Pick one; do not leave it as an accident of collation.
3. Do **not** freeze names to a closed Java enum in this PR. Stacks get added (Django, Mintlify, …). A closed set means a code change + deploy to add `Flutter`. Keep an open string. A catalog of recommended types can be seed data later, not an enum.

Rename is not tested. `testUpdateRepoTypeTrimsName` PATCHes `"React  "`, which trims back to `"React"`. Deleting `repoType.setName(name)` in `updateRepoType` still passes the suite.

Add service + WS coverage for a real rename:

- create name `React`
- PATCH name to `React-ICU`
- GET by id returns `React-ICU`
- GET `?name=React` returns `[]`
- GET `?name=React-ICU` returns the same id

---

### Hard delete

Hard delete means there is no trash can, no deleted flag, no undo.

That is acceptable only while no repository points at a type. This branch does **not** implement a “type in use” guard — and cannot, because there is no `repository.repo_type_id` yet. `DELETE` succeeds if the type exists. The “refuse delete while assigned” rule is documented in `Architecture.md` / javadoc only.

Hard delete is fine on this PR: nothing can point at a type yet, so there is no “in use” check to write.

The next PR (the one that adds “this repository has type X”) must not merge without a delete guard. If a type is still assigned to any repository, DELETE must fail (or require unassigning first). Otherwise we can delete the type and leave repositories pointing at nothing.

I've put that on the assignment ticket https://jira.inside-box.net/browse/LXP-2068 a required item. A sentence in JavaDoc or Architecture.md is not enough — we will forget it.

---

### Auth

`RepoTypeWS` does not implement its own rules. It inherits `WebSecurityConfig`:

- `GET /api/**` → any authenticated user
- `POST` / `PATCH` / `DELETE /api/**` → PM or ADMIN

There is no USER-role test on this resource. Types will eventually hold the AI prompt applied to every repository of that type. Add one WS test: USER gets 403 on POST (and ideally PATCH/DELETE). GET as USER should still be 200.

---

### Must fix — error mapping (409 vs checkers)

Stop mapping every `DataIntegrityViolationException` to a duplicate-name 409.

`createRepoType` / `updateRepoType` catch that exception and return `409 RepoType with name [<request name>] already exists`. That is wrong when the violation is not `UK__REPO_TYPE__NAME` (e.g. the checker table’s composite primary key).

On a **checkers-only PATCH**, `name` is omitted in JSON, so it deserializes as `null`. We then quote the **request** field, not the stored name. The type in the database is still `React`. The body becomes `RepoType with name [null] already exists`. That does not mean the type’s name is null.

Required:

- Duplicate name → 409, message uses the trimmed name we actually tried to persist.
- Invalid / incomplete checkers → 400, message says it is a checker problem (null element, missing `assetExtension`, missing `integrityCheckerType`, unknown enum).
- Any other integrity failure → 500 (or a dedicated 400 if we have a real validation rule). Do not label it as a name conflict.

Handle this in the service (name unique constraint only) or inspect the constraint name. Do not use a multi-catch that assumes every DB failure is a duplicate name.

Document the body in `Architecture.md` and add WS tests for status + message class.

A repo type is: `id` (server), `name` (required on create; optional on PATCH), `description` (optional), `aiPrompt` (optional; create stores `""` if omitted), `integrityCheckers` (optional `{ assetExtension, integrityCheckerType }[]`), timestamps (server).

Create: only `name` is required. PATCH: every field optional; null/omitted = leave unchanged; `""` / `[]` is an explicit clear.

Failures to pin:

- 400 missing/blank name
- 400 name or description > 255
- 400 checker null / missing fields / unknown enum
- 409 duplicate name (create and rename)
- 404 PATCH/DELETE unknown id
- 401 unauthenticated
- 403 USER on POST/PATCH/DELETE

`aiPrompt` still has no max length. Add one, validate it like name/description, and test it. Once E2-F2 lands, this text is injected into every AI call for every repo of that type.

Integrity checker = one control of kind `(file extension, checker type)`. One type may have many checkers. Duplicate pairs collapse. Keep that in the contract tests.

---

### Must fix — delete vs concurrent save

Do not resurrect a deleted type.

`RepoTypeService` uses `findById(...).orElse(repoType)` on reload/update/checker replace. If the row is gone, we keep the caller’s instance and save it. A PATCH that races with DELETE can bring the type back.

If the id is gone, throw `RepoTypeWithIdNotFoundException` (404). Never `orElse(detached payload)` and save.

---

### Must fix — test schema vs production

### Must fix — test schema vs production

Prod/dev Mojito is MySQL and V69 does enforce unique checkers per type. That is not what the unit/integration tests run against.

Tests use in-memory HSQLDB with Flyway off, so Hibernate builds the schema from annotations. Those annotations do not declare the same unique key as V69 (`repo_type_id`, `asset_extension`, `integrity_checker_type`). The de-dupe tests only prove Java Set / LinkedHashMap behavior, not the MySQL constraint.

Please add the same uniqueness on `@CollectionTable` so the test database matches V69. Otherwise a regression in de-dupe can pass CI and only fail (or behave oddly) on MySQL.

If you drop the composite PK from V69 instead, say so and rely only on application de-dupe — but then do not document the PK as the source of truth.

---

### Must fix — `@Lob` on `aiPrompt`

Match how Mojito stores long text.

`RepoType.aiPrompt` uses `@Lob` + `@Column(length = Integer.MAX_VALUE)`. `@Lob` in this codebase is for binary blobs (`Image`, `MBlob`, `ApplicationCache`). Long strings use a bare `@Column(..., length = Integer.MAX_VALUE)` (`TMTextUnit.content`, `AssetContent`, `PollableTask`, `AiReviewProto`).

`@Lob` makes Hibernate treat this as a CLOB; `V69` creates `longtext`. That is another test-schema vs production mismatch.

Drop `@Lob`. Keep `@Column(name = "ai_prompt", length = Integer.MAX_VALUE)`. Add a documented max length and validate it in the service.

---

### Must fix — tests that stay green if you delete the production line

Several tests are green even if the behavior they name is broken. If a test can stay green after you delete the production line it is supposed to protect, it is not a test. Prove each item below by temporarily breaking the code, watching the test fail, then restoring.

1. **Rename** — see above. `testUpdateRepoTypeTrimsName` is not a rename test.
2. **HTTP 201 / 200** are never asserted. `RepoTypeClient.createRepoType` uses `postForObject`; `updateRepoType` returns the body only. `RepoTypeWSTest` already uses a raw template for DELETE (204). Do the same for POST and PATCH.
3. **`testDeleteRepoTypeRemovesCheckers`** does not look at checkers. The assertion is the same as `testDeleteRepoTypeRemovesType` (type row gone). Assert zero collection-table rows for the deleted id.
4. **No raw JSON contract test.** `@JsonView(View.RepoType)` and “checker JSON is only `assetExtension` + `integrityCheckerType`” are the API. Fetch `String`, parse, assert exact top-level keys and checker object shape. The restclient checker DTO has `@JsonIgnoreProperties(ignoreUnknown = true)`, so a leaked `repoTypeId` would be swallowed by every typed test we have.
5. **Update-path length validation** is untested (only blank name). PATCH over-length name and description; assert stored values are unchanged.
6. **Weak assertions:** `assertNotNull(expected.getMessage())` and `contains("name")` that also matches `"name is required"`. Assert the length-specific substring.
7. **Partial write:** `createRepoType` flushes the parent before validating checkers. Add: invalid checker → no type row; valid new name + invalid checker → name unchanged.

---

### CLI vs REST

There is no CLI parity for repo types. That is documented as follow-up in `Architecture.md`; it also needs to be in the PR description.

- REST today: GET list/by id, POST, PATCH, DELETE on `/api/repo-types`.
- CLI today: no `repo-type-create` / `update` / `view` / `delete`. Operators cannot do this from the CLI.

The AI Translations PRD does not require repo-type CLI in Epic 2. It does require CLI for E1-F4 (AI vs human locales) and E1-F5 (auto-translate on `repo-create` / `repo-update` / `repo-view`). Those are other slices. Do not imply this PR closed “everything the API can do, the CLI can do.”

`repo-view` can inspect a **repository** (id, repo-level integrity checkers, locales). It will **not** show a repo type after this PR, because repositories cannot be assigned a type yet. When assignment lands, `repo-view` (and `repo-create` / `repo-update`) must grow a type field. I called that out as a required follow-up on the https://jira.inside-box.net/browse/LXP-2068.

---

### Out of scope for this branch (do not add here)

The Repository AI Translation UI (`/ai-translate`) lets you pick a repo and has no locale picker. That is **Epic 1 (E1-F4)**, not Epic 2 and not this PR. The PRD: on-demand AI uses `repository-ai-translate` with an explicit locale list; default is human; AI is opt-in per locale. File that as a separate ticket (CLI + UI locale list). Do not fold it into `addRepoTypes`.

Screenshot 2026-08-24 at 15 36 41


Glossary (E2-F3) and derived locales (E2-F4 / Epic 3) are correctly absent.

---

**Must-fix to merge:** error mapping, 404 instead of resurrect-on-save, schema/JPA uniqueness alignment, drop `@Lob`, tests that actually pin rename / HTTP status / JSON shape / checker delete / length / rollback, split `FLUENT`, naming decision in the PR body. Everything else is follow-up but should be named, not implied.

ehoogerbeets:
Thanks Natalia — working through your notes in order. Replies so far:

### 1. “Project types” vs “repo type”

We are keeping **repo type** for this epic. That is the term in code (`RepoType`, `/api/repo-types`) and in [`Architecture.md`](https://github.com/box/mojito/blob/addRepoTypes/docs/internal/Architecture.md) (already written as repo type, not project type).

The older “project type” wording was corrected when we started implementing; the stories and this API now match. I am not adding a naming decision to Architecture.md — repo type is already the term throughout.

### 2. Persistence + REST foundation; grouping repos is not user-visible yet

Agreed on the facts, and this is already documented as a foundation slice rather than the finished “group similar repositories” feature.

The PR summary now opens with:

> This PR is a **persistence + REST foundation** for repo types. It does not assign a type to a repository, so grouping similar repositories is not user-visible yet.

It already had:

> **In scope for this PR:** persist types and checkers; authenticated GET; mutating APIs for PM/ADMIN.
>
> **Out of scope (follow-up):** assigning a type on a repository, CLI commands, UI, layered prompt assembly at runtime, and unioning type + repo checkers on push/import.

In Architecture.md, [scope of this server slice](https://github.com/box/mojito/blob/addRepoTypes/docs/internal/Architecture.md#L40-L59) lists persist + REST CRUD as in scope, and “Assign / clear a type on a repository” as follow-up (optional FK). Assignment is a later PR, not this one. No code change for this note.

### 3. Split unrelated work (`FLUENT` on restclient `IntegrityCheckerType`)

Agreed — that is a pre-existing client/server mismatch, not a repo-type feature. `FLUENT` is gone from this branch (restclient enum matches master again; the Fluent round-trip test is removed).

The fix plus name-set tests are in a separate PR: https://github.com/box/mojito/pull/1080

`View.java`’s `RepoType` JSON view stays here.

### 4. Naming: grammar, uniqueness, closed set, rename tests

Documented in Architecture.md (decision 8) and JavaDoc. Same idea as repository names:

- Required, trimmed, max 255, unique
- **Case-sensitive** (`React` and `react` are different types). We expect only a small set of types, so that is fine.
- Open string, not a Java enum — adding a stack does not need a code change. A recommended catalog can be seed data later.

Tests:

- Creating a name does not block the same letters in a different case (service + WS)
- Real rename: PATCH to a new name, GET by id returns the new name, GET `?name=` old is empty, GET `?name=` new returns the same id (service + WS)

(More replies below as we work through the rest of the list.)

ehoogerbeets:
### 5. Hard delete

Agreed — hard delete is correct on this PR because repositories cannot reference a type yet, so there is no “in use” check to write.

The delete guard belongs on the assignment story. It is already a required item on [LXP-2068](https://jira.inside-box.net/browse/LXP-2068) (your comment there). We are not adding more Architecture.md / JavaDoc for this; that is not the control.

No code change on this branch.

ehoogerbeets:
### 6. Auth

Agreed — `RepoTypeWS` inherits `WebSecurityConfig` (authenticated GET; POST/PATCH/DELETE → PM or ADMIN). There was no USER-role coverage, and types will hold the AI prompt applied to every repo of that type.

Landed in e6bfbd5b:

- `testUserRoleCanGetAndIsForbiddenToMutate` in `RepoTypeWSTest`
- USER GET by id → 200
- USER POST / PATCH / DELETE → 403, and the type is unchanged (no create, no description change, no delete)

The test logs in on a dedicated `CookieStoreRestTemplate`. The shared client interceptor treats 403 as a stale session and re-authenticates, so it cannot assert USER authorization failures. Username is kept short because `SPRING_SESSION_V2.PRINCIPAL_NAME` is `VARCHAR(100)`.

Architecture.md and `RepoTypeWS` JavaDoc now state USER GET 200 / USER mutate 403.

ehoogerbeets:
### 7. Error mapping (409 vs checkers)

Agreed — create/update mapped every `DataIntegrityViolationException` to a duplicate-name 409, and a checkers-only PATCH could quote `name [null]`.

Landed in d52de902:

- 409 only for `RepoTypeNameAlreadyUsedException` and constraint `UK__REPO_TYPE__NAME`
- 409 body uses the trimmed name that was actually persisted (`RepoTypeNameAlreadyUsedException.getName()`)
- A checkers-only PATCH with a null request name is not turned into `RepoType with name [null] already exists`
- Other integrity failures (e.g. checker composite PK) are rethrown → 500, not labeled as a name conflict
- Invalid checkers → 400 with the validation message (`integrity checker must not be null`, `integrityCheckerType is required`, `assetExtension is required`)

Unknown enum values stay Jackson 400 (not our message). Tests assert the full error strings, not substrings. Architecture.md **Error bodies** documents the 409 vs 400 split.

ehoogerbeets:
### 8. Delete vs concurrent save

Agreed — `findById(...).orElse(repoType)` plus `save` can re-insert a deleted type if the id is gone in this session. (Two HTTP requests on the same id are still serialized by the DB row lock; this is the leftover-instance fallback, not a substitute for a transaction.)

Landed in 24e7ccd7:

- Reload/update/checker replace use `getRepoTypeById` — missing id is `RepoTypeWithIdNotFoundException` (404)
- `updateIntegrityCheckers` does not save the caller's instance when the row is gone
- Sequential tests: delete, then `updateRepoType` / `updateIntegrityCheckers` with that id (or leftover instance) → 404, and the row stays gone

Architecture.md documents that missing type id on checker replace is 404, not a re-insert.

ehoogerbeets:
### 9. Test schema vs production

Agreed — CI is HSQL with Flyway off, so Hibernate must declare the same uniqueness V69 has on MySQL. We did not drop the composite PK.

Landed in 3cbb4280:

- `@CollectionTable` now has `UK__REPO_TYPE_INTEGRITY_CHECKER` on `(repo_type_id, asset_extension, integrity_checker_type)`
- Java de-dupe stays so a duplicate JSON body is still 200, not 500; the unique key is the backstop
- `testDuplicateCheckerPairRejectedByDatabaseConstraint`: three distinct pairs via `RepoTypeService`; a colliding insert on the join table is `DataIntegrityViolationException` (SQL on purpose — Spring persist would collapse the pair in memory). One test, no Hibernate-only skip — same uniqueness is V69's PK on MySQL

Architecture.md documents both the V69 PK and the JPA unique constraint.

ehoogerbeets:
### 10. `@Lob` on `aiPrompt`

Agreed on dropping `@Lob`. In this codebase that annotation is for binary blobs (`Image`, `MBlob`, `ApplicationCache`). Long strings use `@Column(..., length = Integer.MAX_VALUE)` (`TMTextUnit.content`, `AssetContent`, `AiReviewProto`). V69 already creates `longtext`; we did not change the migration.

We are **not** adding an application max length. A Mojito instance has on the order of tens of types. Operators keep prompts short themselves because this text is later injected into every AI call for every repo of that type — more text is more tokens and more cost. An arbitrary 64KiB / 1MiB cap would be a product limit we do not need. Practical ceilings are HTTP body size, heap, and MySQL `max_allowed_packet`, not a Mojito 400.

Landed in 14d63e37:

- Dropped `@Lob` on `RepoType.aiPrompt`; column stays `@Column(name = "ai_prompt", length = Integer.MAX_VALUE)`
- No service 400 for over-length `aiPrompt` (name/description stay 255)
- Architecture.md decision 2 records that
- Create (service + WS) and PATCH (service) persist a 10,000-character prompt — only to prove we do not reuse the 255 name cap and that HSQL still stores long text without `@Lob`

ehoogerbeets:
### 11. Tests that stay green if you delete the production line

Agreed. Several tests named behavior they did not pin. Rename (item 1) was already covered in the naming work. Weak length assertions (item 6) were already exact `must be at most 255 characters` on create. The rest landed in 21adb5ca (tests only; no production change — checker validation after `saveAndFlush` already rolled back):

- **HTTP 201 / 200** — raw `postForEntity` / `exchange` like DELETE’s 204. `RepoTypeClient.createRepoType` still uses `postForObject` and cannot see status.
- **`testDeleteRepoTypeRemovesCheckers`** — join-table `COUNT(*)` for that `repo_type_id` is 0, not only the type row gone.
- **Raw JSON** — GET as `String`, exact top-level keys (`id`, `name`, `createdDate`, `lastModifiedDate`, `description`, `aiPrompt`, `integrityCheckers`) and each checker is only `{ assetExtension, integrityCheckerType }`. Typed restclient tests cannot catch a leaked `repoTypeId` because the DTO ignores unknown properties.
- **PATCH length** — over-length name and description on update are 400; stored values unchanged (service + WS).
- **Partial write** — invalid checker on create leaves no type row; rename + invalid checker leaves the original name.

Full `RepoTypeServiceTest` + `RepoTypeWSTest` passed locally (96/96).

ehoogerbeets:
### 12. CLI vs REST

Agreed — and we are not adding CLI on this branch.

`Architecture.md` already lists `repo-type-create` / `update` / `view` / `delete` as follow-up. The PR summary called this a persistence + REST foundation with CLI out of scope; that was easy to read as a bullet instead of an operator fact.

Updated the PR description to say explicitly:

- No CLI parity: operators cannot do repo-type CRUD from the CLI
- `repo-view` will not show a type after this PR (nothing can assign a type yet)
- Assignment (LXP-2068) is where `repo-view` / `repo-create` / `repo-update` grow a type field
- This PR does not close “everything the API can do, the CLI can do”

E1-F4 / E1-F5 CLI work stays on those slices.

nmkedziora:
Thanks — I pulled and checked the branch. Notes 1–7 match the code.

The later commits also cover the rest of the must-fix list (no resurrect, JPA unique matching V69, drop @Lob, 201/200 + JSON shape + checker delete + rename/length/rollback tests). I am not holding the remaining nits.

Case-sensitive names (same as repositories) and no application cap on aiPrompt are fine given the rationale.

+1 from my side.

nmkedziora (approved):

zmoskala (approved):

PR #1083, opened 2026-08-27
Allow assigning and clearing a repo type on a repository

nmkedziora:
## Summary
- Adds optional `repository.repo_type_id` so a Mojito repo can be assigned a repo type (or left untyped / SQL `NULL`).
- `repo-create` / `repo-update` take `--repo-type <name>` (exact, case-sensitive after trim). Omitting the flag on create leaves the repo untyped; omitting it on update does **not** clear. `--clear-repo-type` unassigns and is mutually exclusive with `--repo-type`. `--repo-type NULL` looks up a type literally named `NULL`.
- `repo-view` prints the assigned type when present. Nested REST JSON is `id` + `name` only.
- `DELETE /api/repo-types/{id}` returns **409** while any repository still references the type; soft-deleting a repository clears the assignment so it cannot block type delete.

## Test plan
- [X]  ```./mvnw test -pl webapp -am -Dtest=RepositoryWSTest,RepoTypeWSTest -Dsurefire.failIfNoSpecifiedTests=false```
Screenshot 2026-09-08 at 17 07 25


zmoskala:
In general it looks good to me, but wanted to confirm one thing: `RepositoryRepository.findAll` method had no `@EntityGraph` before, and it backs `GET /api/repositories`. The PR attaches `Repository.legacy` to it, which loads a lot more than `repoType` (locales, locale stats, checkers, TM, user, screenshot run). Adding `repoType` to the graph used by `findById` / `findByName` makes sense, but I wanted to confirm on applying the full graph to the list query.
Was that intentional (list should load like get-by-id), or would an eager `repoType` / a smaller graph be enough so the list doesn’t pick up the rest?

zmoskala:
Also can you pls resolve conflicts?

cli/src/main/java/com/box/l10n/mojito/cli/command/param/Param.java
```diff
@@ -80,6 +80,14 @@ public class Param {
   public static final String REPO_TYPE_DESCRIPTION_DESCRIPTION =
       "Description of the repo type to create or update";
 
+  public static final String REPOSITORY_TYPE_LONG = "--repo-type";
+  public static final String REPOSITORY_TYPE_DESCRIPTION =
+      "Exact name of the repository type to assign";
+
+  public static final String CLEAR_REPOSITORY_TYPE_LONG = "--clear-repo-type";
```
ehoogerbeets, line 87: For consistency, we should clear the repo type using the same method as with other things in the CLI, which is to set it to empty string. This is the way it works with integrity checkers for repos and repo types already:

`mojito repo-update -n MyRepo -it ""`

which clears all the checkers. Similarly, we can do:

`mojito repo-update -n MyRepo --repo-type ""`

webapp/src/main/resources/db/migration/V70__Assign_repo_type_to_repository.sql
```diff
@@ -0,0 +1,3 @@
+alter table repository add column repo_type_id bigint;
```
ehoogerbeets, line 1: This should be in the V69 file. We have not upgraded from V68 yet.

cli/src/main/java/com/box/l10n/mojito/cli/command/param/Param.java
nmkedziora: Got it, updated in 07a2dcf53eed5a004bcb0c593c0d251c239d2409

webapp/src/main/resources/db/migration/V70__Assign_repo_type_to_repository.sql
nmkedziora: Moved in 87f0b2f41eb17ea282f4317668826e1b50c3a405

nmkedziora:
> In general it looks good to me, but wanted to confirm one thing: `RepositoryRepository.findAll` method had no `@EntityGraph` before, and it backs `GET /api/repositories`. The PR attaches `Repository.legacy` to it, which loads a lot more than `repoType` (locales, locale stats, checkers, TM, user, screenshot run). Adding `repoType` to the graph used by `findById` / `findByName` makes sense, but I wanted to confirm on applying the full graph to the list query. Was that intentional (list should load like get-by-id), or would an eager `repoType` / a smaller graph be enough so the list doesn’t pick up the rest?

@zmoskala changed as suggested, b5513dd0deb37058b8a567f78b77797219fc5320

zmoskala:
> > In general it looks good to me, but wanted to confirm one thing: `RepositoryRepository.findAll` method had no `@EntityGraph` before, and it backs `GET /api/repositories`. The PR attaches `Repository.legacy` to it, which loads a lot more than `repoType` (locales, locale stats, checkers, TM, user, screenshot run). Adding `repoType` to the graph used by `findById` / `findByName` makes sense, but I wanted to confirm on applying the full graph to the list query. Was that intentional (list should load like get-by-id), or would an eager `repoType` / a smaller graph be enough so the list doesn’t pick up the rest?
> 
> @zmoskala changed as suggested, [b5513dd](https://github.com/box/mojito/commit/b5513dd0deb37058b8a567f78b77797219fc5320)

Thanks! Looks good to me. I am leaving threads started by @ehoogerbeets for Edwin to resolve.

ehoogerbeets (approved):

zmoskala:
Fixed some tests to unblock us: `RepoCreateCommandTest` / `RepoUpdateCommandTest` expected `--repo-type` `" "` to be rejected with `repoType.name is required`. That is the REST rule (blank nested name → 400). It is not something the CLI can enforce: JCommander 1.48 trims flag values, even when quoted, so " " becomes "" before the command runs.

From there the commands do the documented empty-string behavior:

`repo-create --repo-type " "` creates an untyped repo (same as `""`)
`repo-update --repo-type " "` clears the assignment (same as `""` / `-it ""`)
The update test then NPEd on `getRepoType().getId()` because the type had been cleared.

I updated those CLI tests to match that, and added REST tests that POST/PATCH `" "` so the 400 + “don’t change the assignment” rule is still covered where the spaces actually reach the server.

@nmkedziora for us to be able to merge, CLA needs to be signed first.

PR #1086, opened 2026-09-01
Resolve flaky unit tests so that we can rely on them again in the CI

ehoogerbeets:
## Summary
- `GitBlameCommandTest` now builds a throwaway git repo in test setup, commits the fixtures with a fixed author and time, and asserts against that commit.
- Removes the CI skip guards and the partial spy that still called real `git blame` on the ambient checkout, which failed on shallow or partial clones.
- PO fixtures blame `file.js` at the temp repo root so usage paths no longer depend on this repository's layout.
- `QuartzJobsDeleteCommandTest` no longer asserts that the shared Quartz DYNAMIC group is empty. Other tests and the webapp keep scheduling jobs there, so a global count of zero was racy.
- That test now only tracks its own uniquely named `AJob` entries before and after `quartz-jobs-delete`.
- `BranchDeleteCommandTest.delete` no longer waits for background Quartz/`@Async` branch-statistics jobs. After `push` and after `import` it calls `BranchStatisticService.computeAndSaveBranchStatistics` so translated/untranslated flags are present before the assertions, instead of racing a contended scheduler (~50s timeout on CI).

## Test plan
- [ ] `mvn -pl cli -Dtest=GitBlameCommandTest test` passes locally
- [ ] The same test class passes with `GITHUB_ACTIONS=true` (previously skipped)
- [ ] CI on this PR runs `GitBlameCommandTest` without skip or shallow-clone failures
- [ ] `mvn -pl cli -Dtest=QuartzJobsDeleteCommandTest test` passes locally
- [ ] CI no longer fails `QuartzJobsDeleteCommandTest` with leftover `RepositoryStatisticsJob` (or other) dynamic jobs
- [ ] `mvn -pl cli -Dtest=BranchDeleteCommandTest#delete test` passes locally
- [ ] CI no longer fails `BranchDeleteCommandTest.delete` waiting for `b1, b2, b3, b4` untranslated flags


nmkedziora:
@ehoogerbeets what was the root cause for failing tests?

ehoogerbeets:
@nmkedziora Good question — all three were pre-existing flakes rather than anything new, but they had different root causes:

**`GitBlameCommandTest`** — The test asserted blame data from real commits in the Mojito repo's own history, and part of the code path still called real `git blame` rather than being stubbed. CI checkouts are shallow, so those old commits weren't reachable and the blame came back empty; it only passed on machines with full history. It now creates its own throwaway git repo with a fixed author and commit time, so the data it asserts on is data it created — and it no longer has to be skipped in CI.

**`QuartzJobsDeleteCommandTest`** — The test required the shared Quartz dynamic job group to be completely empty before and after its own run. The webapp and other tests are constantly scheduling jobs into that same group, so unrelated leftovers made the count non-zero at the moment it checked. It now watches only its own two uniquely named jobs and ignores everything else.

**`BranchDeleteCommandTest.delete`** — The test waited for branch translation statistics that are computed in the background by Quartz and async jobs, on a thread pool shared with every other CLI test. Its wait budget was a fixed ~50 seconds, which was enough on a laptop but not on a loaded CI agent, so it timed out even though nothing was broken. It now computes those statistics directly in the test, so the data is ready before the assertions instead of depending on scheduler load.

The common thread across all three: each one depended on shared or ambient state (the repo's git history, the global job scheduler, a contended thread pool) instead of state it controlled itself.

cli/src/test/java/com/box/l10n/mojito/cli/command/QuartzJobsDeleteCommandTest.java
```diff
@@ -56,14 +60,34 @@ public void testDeleteAllDynamicJobs() throws Exception {
...
+
+  /**
+   * Job keys are {@code canonicalClassName + "_" + uniqueId}. Match the same substring used by
+   * {@code quartz-jobs-view} assertions so leftover jobs from other tests are ignored.
+   */
+  private boolean hasTestJob(String uniqueId) {
+    try {
+      String needle = "AJob_" + uniqueId;
+      return quartzService.getDynamicJobs().stream().anyMatch(name -> name.contains(needle));
+    } catch (SchedulerException e) {
```
zmoskala, line 88: Nit: `false` here means both “job is gone” and “I couldn’t list jobs.” After delete we wait on `!hasTestJob(...)`, so a listing error looks like a successful delete. Safer to let the exception bubble and let `waitForCondition` retry.

zmoskala (approved):
Approving with comment - flagging the inverted meaning of false

cli/src/test/java/com/box/l10n/mojito/cli/command/QuartzJobsDeleteCommandTest.java
ehoogerbeets: Good catch — `false` on a listing error made `!hasTestJob(...)` look like a successful delete.

`hasTestJob` now wraps `SchedulerException` in a `RuntimeException` and rethrows it. `waitForCondition` already treats thrown exceptions as “not ready yet” and retries, so a listing failure waits instead of passing, and the test only succeeds when listing works and the jobs are actually gone.

Pushed in 2d5a310f.

PR #1087, opened 2026-09-21
Repo Type Integrity Checkers at Check Time

zmoskala:
## Summary

- Repository type integrity checkers now run at check time. `IntegrityCheckerFactory.getTextUnitCheckers` resolves a set union of type-owned and repository-owned checker types whose `assetExtension` matches the asset path's extension, so the same `(assetExtension, integrityCheckerType)` configured in both places is instantiated once, while different checker types on the same extension all run. Untyped repositories keep using only their own `-it` checkers.
- The type-owned half is loaded by an explicit JPQL query on `AssetIntegrityCheckerRepository` keyed by repository id plus asset extension. The check path is reached with detached assets, so resolving through `asset.getRepository().getRepoType()` would depend on a lazy association being initialized; the query avoids that and returns an empty set for an untyped repository.
- The assignment is read on every check rather than copied into repository config, so assigning a type, clearing it, or editing a type's checkers changes the next import with no repository-level change. This covers every path that goes through the factory: XLIFF import, localized-asset import, batch import, workbench checks, and pseudolocalization.
- `repo-view` stays repository-only for `-it` config: type-owned checkers print on a separate `Repository type checkers -->` line and are never merged into `Integrity checkers -->`. They come from a follow-up `RepoTypeClient.getRepoTypeById` because the nested `repoType` on the repository payload is still only `id` + `name`. `repo-create` / `repo-update` help now states that `-it` stores checkers on the repository only and that an assigned type also runs its own.
- No new REST endpoint and no schema change — the type-owned rows already existed on the repo-type API. Docs updated: the integrity-checkers and creating-repository guides describe the union, and the runtime superset moves out of "out of scope" in Architecture and off the ROADMAP list. CLI and UI for managing a type's checkers is still follow-up.

### Manual

_Placeholder — manual verification steps to be added._

### Automated

```bash
./mvnw -pl webapp -am -Dtest=AssetIntegrityCheckerServiceTest -Dsurefire.failIfNoSpecifiedTests=false spotless:check test
```

- `AssetIntegrityCheckerServiceTest` (`ServiceTestBase`, real DB): untyped repository with no checkers, type-only, repo-only, and the same pair on both; disjoint pairs in both directions (type checker rejects what the repo checker would allow, and the reverse); a typed repository whose type has an empty checker set still runs its own checkers; extension filtering and union asserted directly on the factory with a detached asset, checking that the lazy `repoType` is not initialized first; assigning a type, clearing it, and editing a type's checkers between two imports of the same text unit; and the workbench check, localized-asset import, and pseudolocalization paths driven by a type-only checker.
- `TextUnitBatchImporterServiceTest#testTypeOnlyIntegrityCheckerIsUsedOnBatchImport`: a type-owned `MESSAGE_FORMAT` excludes a broken placeholder on batch import, so the union applies to `asyncImportTextUnits` and not just the XLIFF path.
- `RepoViewCommandTest` (`CLITestBase`, real HTTP): `Repository type checkers` appears only when the assigned type has checkers, is absent for an untyped repository and for a type with none, and a repository `-it` checker added afterwards prints on `Integrity checkers` without the two lines collapsing into one.
- `RepoCreateCommandTest#testCreateHelpDocumentsTypeIntegrityUnion`: help documents both halves of the contract — `-it` stores repository checkers only, and an assigned type also runs its checkers — so dropping either sentence fails the test.


webapp/src/test/java/com/box/l10n/mojito/service/assetintegritychecker/AssetIntegrityCheckerServiceTest.java
```diff
@@ -52,36 +79,232 @@ public class AssetIntegrityCheckerServiceTest extends ServiceTestBase {
...
+  private static final String BROKEN_MESSAGE_FORMAT_TARGET =
+      "{numFiles, plural, one{Il y a un fichier} other{Il y a # fichiers}";
+  private static final String VALID_MESSAGE_FORMAT_TARGET =
+      "{numFiles, plural, one{Il y a un fichier} other{Il y a # fichiers}}";
+  private static final String MESSAGE_FORMAT_SOURCE_WITH_ELLIPSIS =
+      "{numFiles, plural, one{# There is one file…} other{There are # files…}}";
+  private static final String VALID_MESSAGE_FORMAT_TARGET_WITH_THREE_DOTS =
+      "{numFiles, plural, one{Il y a un fichier...} other{Il y a # fichiers...}}";
+  private static final String BROKEN_MESSAGE_FORMAT_TARGET_ALTERNATE =
+      "{numFiles, plural, one{Il y a deux fichiers} other{Il y a # fichiers}";
```
ehoogerbeets: Nitpick

Il y a deux fichiers => There are two files

This is in the singular plural choice.

ehoogerbeets (approved):
My only comment is a nitpick, so approving this. You can update the code or not. Your choice.

